PR #97, opened 2022-09-29
Resilient indexing

pandu-k:

- Feature

- If a document in a large add_document calls errors, then the entire call fails (rather than just the errorful doc)

- Add_documents continues, with errored documents being indicated in the response 

- Perhaps: if you have a logic predicated on an error response for an add_documents call, it would no longer work
- Instead you should inspect the response of the add documents call, and see if "errors" has been set to True. You can then inspect the returned items, to see which document/s triggered created an error and why

* **Other information**:
- Ran tox

### DIFF
--- a/src/marqo/tensor_search/tensor_search.py
+++ b/src/marqo/tensor_search/tensor_search.py
@@ -282,7 +282,7 @@ def add_documents(config: Config, index_name: str, docs: List[dict], auto_refres
         create_vector_index(config=config, index_name=index_name)
         index_info = backend.get_index_info(config=config, index_name=index_name)
 
-    if not docs:
+    if len(docs) == 0:
         raise errors.BadRequestError(message="Received empty add documents request")
 
     existing_fields = set(index_info.properties.keys())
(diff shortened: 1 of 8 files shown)


src/marqo/tensor_search/tensor_search.py
```diff
@@ -329,20 +352,29 @@ def add_documents(config: Config, index_name: str, docs: List[dict], auto_refres
                     # TODO put the logic for getting field parameters into a function and add per field options
                     image_method = index_info.index_settings[NsField.index_defaults][NsField.image_preprocessing][NsField.patch_method]
                     # the chunk_image contains the no-op logic as of now - method = None will be a no-op
-                    content_chunks, text_chunks = image_processor.chunk_image(field_content, 
-                                    device=selected_device,
-                                    method=image_method)            
+                    content_chunks, text_chunks = image_processor.chunk_image(
```
jn2clark: the error will also need to be caught here. if an image is chunked then there will be an attempt to retrieve it here prior to the vectorise call

jn2clark (commented):
I left another comment but the ```try``` for catching the error will also need to encompass the chunking as well

src/marqo/tensor_search/tensor_search.py
pandu-k: added the try except block around chunking and tested

jn2clark: Thank you!

jn2clark (approved):

PR #114, opened 2022-10-06
Generic model support

aryanagarwal9:

feature

Only models specified in the model_registry can be used

generic models, passed as a dictionary, are supported

No

* **Other information**:


src/marqo/s2_inference/s2_inference.py
```diff
@@ -47,6 +59,23 @@ def _create_model_cache_key(model_name: str, device: str) -> Tuple:
     model_cache_key = (model_name, device)
     return model_cache_key
 
+
+def _update_model_properties(model: Union[str, dict], model_name: str) -> None:
+    """updates MODEL_PROPERTIES if model details are specified by user
+    """
```
jn2clark: maybe do a check for the key names in the dict

jn2clark: "names", "dimensions", "type"
maybe type can be optional, sbert as default

src/marqo/s2_inference/s2_inference.py
```diff
@@ -209,14 +238,14 @@ def _load_model(model_name: str, device: str = get_default_device()) -> Any:
     """
 
     model_properties = get_model_properties(model_name)
-    
+
```
jn2clark: add some error handling here in case the model does not load properly

jn2clark: create one here https://github.com/marqo-ai/marqo/blob/mainline/src/marqo/s2_inference/errors.py

jn2clark: class ModelLoadError(S2InferenceError):
    pass

src/marqo/s2_inference/s2_inference.py
```diff
@@ -47,6 +59,23 @@ def _create_model_cache_key(model_name: str, device: str) -> Tuple:
...
+    if isinstance(model, dict):
+        model_properties = {model_name: model}
+        MODEL_PROPERTIES['models'].update(model_properties)
+
+
+def _update_available_models(model_cache_key: str, model_name: str, device: str, normalize_embeddings: bool) -> None:
+    """loads the model if it is not already loaded
+    """
+    if model_cache_key not in available_models:
+        available_models[model_cache_key] = _load_model(model_name, device=device)
```
jn2clark: this is where it will fail if the model does not exist or there is an error loading

jn2clark (changes requested):

src/marqo/s2_inference/s2_inference.py
jn2clark: "tokens" can be optional, default to 128
model_name.get("tokens", 128)

src/marqo/s2_inference/s2_inference.py
```diff
@@ -27,19 +27,33 @@ def vectorise(model_name: str, content: Union[str, List[str]], device: str = get
...
 
-def _create_model_cache_key(model_name: str, device: str) -> Tuple:
+
+def _get_model_name(model: Union[str, dict]) -> str:
+    """returns the name of the model
+    """
+    if isinstance(model, str):
+        return model
+    elif isinstance(model, dict):
+        return model['name']
```
jn2clark: i think we want the check for the valid dict to happen before this in case the "name" field is not present

src/marqo/s2_inference/s2_inference.py
```diff
@@ -53,6 +67,49 @@ def _create_model_cache_key(model_name: str, device: str) -> Tuple:
     model_cache_key = (model_name, device)
     return model_cache_key
 
+
+def _check_model_dict(model: dict) ->  bool:
```
jn2clark: can we move this earlier in the process and before the "name" key is accessed?

src/marqo/s2_inference/s2_inference.py
```diff
@@ -13,12 +13,12 @@
 available_models = dict()
 MODEL_PROPERTIES = load_model_properties()
 
-def vectorise(model_name: str, content: Union[str, List[str]], device: str = get_default_device(), 
+def vectorise(model: Union[str, dict], content: Union[str, List[str]], device: str = get_default_device(),
```
jn2clark: this is incorrectly named - either use the old name (`model_name` or also modify all the instances that call vectorise

pandu-k:
Because this affect's Marqo's interface, this branch should run this against the API tests, before merging.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -364,13 +364,23 @@ def add_documents(config: Config, index_name: str, docs: List[dict], auto_refres
                                  'code': image_err.code})
                         )
                         break
-                
+
                 normalize_embeddings = index_info.index_settings[NsField.index_defaults][NsField.normalize_embeddings]
                 infer_if_image = index_info.index_settings[NsField.index_defaults][NsField.treat_urls_and_pointers_as_images]
+
+                try:
+                    model_properties = s2_inference.get_model_properties_from_registry(model_name=index_info.model_name)
```
jn2clark: lets not do another call to s2_inference here, instead just check for the `model_properties` in the index defaults and pass through else pass through None

src/marqo/tensor_search/tensor_search.py
```diff
@@ -364,13 +364,23 @@ def add_documents(config: Config, index_name: str, docs: List[dict], auto_refres
...
+                except KeyError:
+                    index_defaults = index_info.get_index_settings()["index_defaults"]
+                    try:
+                        model_properties = index_defaults[NsField.model_properties]
```
jn2clark: `model_properties = dictionary.get(NsField.model_properties, None)`

src/marqo/tensor_search/tensor_search.py
```diff
@@ -117,6 +118,13 @@ def create_vector_index(
     return response
 
 
+def _check_model_name(index_settings):
+    model_name = index_settings[NsField.index_defaults].get(NsField.model)
+    model_properties = index_settings[NsField.index_defaults].get(NsField.model_properties)
```
jn2clark, line 141: maybe explicitly set the default in the `.get` call?

src/marqo/s2_inference/s2_inference.py
```diff
@@ -50,17 +53,65 @@ def _create_model_cache_key(model_name: str, device: str) -> Tuple:
     Returns:
```
pandu-k, line 53: Why are we encoding the key, rather than keeping it a string?

Also, can we make the key `"modelname_device_<PROPERTIES HASH OR HASH SUBSTRING>"` or something similar (would help with readability)

src/marqo/s2_inference/s2_inference.py
```diff
@@ -50,17 +53,65 @@ def _create_model_cache_key(model_name: str, device: str) -> Tuple:
...
+def _validate_model_properties(model_name: str, model_properties: dict) -> dict:
+    """validate model_properties, if not given then return model_registry properties
+    """
+    if model_properties is not None:
+        """checks model dict to see if all required keys are present
+        """
+        required_keys = ["name", "dimensions"]
+        for key in required_keys:
+            if key not in model_properties:
+                raise InvalidModelSettingsError(f'model {key} not in settings')
```
pandu-k: How about `InvalidModelPropertiesError` to keep this consistent?

PR #131, opened 2022-10-19
Content Routering in indexing and X clip

wanliAlex:

content_routering and add video support from x_clip model

we don't have content_routering and don't support video input

we now can support video and router the content

No

* **Other information**:
This is not the complete pull request. The x_clip and content_routering still need end-to-end test.


src/marqo/s2_inference/model_registry.py
```diff
@@ -541,13 +542,31 @@ def _get_random_properties() -> Dict:
     }
     return RANDOM_MODEL_PROPERTIES
 
+def _get_XCLIP_properties() -> Dict:
```
jn2clark, line 552: are these other models also available? https://github.com/microsoft/VideoX/tree/master/X-CLIP

src/marqo/s2_inference/processing/video.py
```diff
@@ -0,0 +1,28 @@
+import moviepy
+import PIL
+
+def convert_to_listofPIL(video):
```
jn2clark: lets also add the type hints

src/marqo/s2_inference/processing/video.py
```diff
@@ -0,0 +1,28 @@
...
+    frames = list(video.iter_frames())
+    return [PIL.Image.fromarray(i.astype("uint8"), 'RGB') for i in frames]
+
+def chunk_video(field_content : str, device = "cpu", method = "simple"):
+    clip = field_content
+    duration = clip.duration
+    content_chunks = []
+    reminder = duration % 10
+
+    if duration < 15:
```
jn2clark: how are these values (`15`) determined? we might want to have them as keyword args or set them in a config somewhere

src/marqo/s2_inference/processing/video.py
```diff
@@ -0,0 +1,28 @@
...
+        if reminder > 5.0:
+            num_of_chunks = duration // 10 + 1
+        else:
+            num_of_chunks = duration // 10
+        for i in range(int(num_of_chunks) - 1):
+            content_chunks.append(clip.subclip(i * 10, i * 10 + 10))
+        content_chunks.append(clip.subclip(10 * (num_of_chunks - 1), 10 * num_of_chunks + reminder))
+
+    trunk_duration = [i.duration for i in content_chunks]
+    assert duration == sum(trunk_duration), "The truncks are not correct"
```
jn2clark: lets raise a RuntimeError instead of the assert 
```
if condition:
    raise RuntimeError(error_message)
```

src/marqo/s2_inference/x_clip_utils.py
```diff
@@ -0,0 +1,282 @@
...
+        if self.model is None:
+            self.load()
+        # default to batch encoding
+        # Note that a single video also is a list
+        if isinstance(videos, list) and isinstance(videos[0], list):
+            video_input = format_and_load_XCLIP_videos(videos)
+        elif isinstance(videos, list) and isinstance(videos[0], ImageType):
+            video_input = [format_and_load_XCLIP_video(videos)]
+
+        #print("processing video")
```
jn2clark: we can remove these commented statements

src/marqo/tensor_search/utils.py
```diff
@@ -137,3 +148,91 @@ def check_device_is_available(device: str) -> bool:
...
+    if len(fname) == 0:
+        return None
+    return fname[0]
+
+def content_routering(field_content, infer_if_media = True):
+    if isinstance(field_content, str):
+        if infer_if_media:
+            if validators.url(field_content):
+                try:
+                    downloaded_file = wget.download(field_content)
```
jn2clark: can we control the name of the downloaded file here?

jn2clark: i think we might want to use temp dir for now

src/marqo/tensor_search/utils.py
```diff
@@ -137,3 +148,91 @@ def check_device_is_available(device: str) -> bool:
...
+    Get filename from content-disposition
+    """
+    if not cd:
+        return None
+    fname = re.findall('filename=(.+)', cd)
+    if len(fname) == 0:
+        return None
+    return fname[0]
+
+def content_routering(field_content, infer_if_media = True):
```
jn2clark: i think we will need to simplify this function - maybe split some parts out?

src/marqo/s2_inference/processing/video.py
pandu-k: There are 2 ways to set this duration parameter: 
1. We can create a new `IndexSettingsField.video_preprocessing` dictionary in configs.py, that can have a `video_chunk_length` parameter (or Jesse can come up with a better name for this param)

See here for reference: https://github.com/marqo-ai/marqo/blob/4a6dfd7a5170336aafd472092bc5c9828a0c3b7a/src/marqo/tensor_search/configs.py#L19

2. If you think the user can crash Marqo with an overly massive video chunk you define it with a default environment variable. 

- Default env vars defined here:  https://github.com/marqo-ai/marqo/blob/4a6dfd7a5170336aafd472092bc5c9828a0c3b7a/src/marqo/tensor_search/configs.py#L27
- To read the environment variable, use this function: https://github.com/marqo-ai/marqo/blob/4a6dfd7a5170336aafd472092bc5c9828a0c3b7a/src/marqo/tensor_search/utils.py#L164

src/marqo/s2_inference/model_registry.py
wanliAlex: done

src/marqo/s2_inference/x_clip_utils.py
wanliAlex: done

src/marqo/s2_inference/processing/video.py
wanliAlex: done

src/marqo/s2_inference/processing/video.py
wanliAlex: done

src/marqo/tensor_search/utils.py
wanliAlex: A temp directory is used to store the downloaded files.

As for the filename, mimetypes uses the file name to determine the type. Maybe we should just keep the original name?

src/marqo/s2_inference/processing/video.py
wanliAlex: The length of each chunk is 10s as the x_clip model is trained by videos with 10 seconds.
    
    
    The last chunk is very likely to be less than 10s, we solve the problem using the following rules:
    If the last chunk is longer than 5 seconds, we keep the chunk.
    If the last chunk is less than 5 seconds, we add this chunk to the previous one.

    For example, for a video with 34s, we will have 3 chunks with length [10s, 10s, 14s]
    As for a video with 38s, we will have 4 chunks with length [10s, 10s, 10s, 8s]

wanliAlex: I will add two configs, `video_chunk_length` and `video_chunk_merge_flag` to solve this problem.

src/marqo/tensor_search/enums.py
```diff
@@ -5,12 +5,26 @@
...
 
+class FileType:
+    youtube = "youtube"
+    tiktok = "tiktok"
+    url = "url"
+    local = "local_path"
+    straight_text = "straight_text"
+    ndarray = "ndarray"
+    PILImage = "PILImage"
+    ListOfPILImage = "ListOfPILImage"
```
pandu-k: Can we please keep these snake-cased?
PILImage -> pil_image
ListOfPILImage -> list_of_pil_images
Unless there's a special reason for this casing

src/marqo/tensor_search/utils.py
```diff
@@ -139,6 +150,89 @@ def check_device_is_available(device: str) -> bool:
...
+    Get filename from content-disposition
+    """
+    if not cd:
+        return None
+    fname = re.findall('filename=(.+)', cd)
+    if len(fname) == 0:
+        return None
+    return fname[0]
+
+def content_routering(field_content, download_path, infer_if_media = True):
```
pandu-k: This is function acts like an interface, can you please make the function signature more detailed? And please add a docstring.



For example: 
```
def content_routering(field_content: str, download_path: str, infer_if_media = True) -> Tuple[str, str, str]
"""
Args:
   field_content: <PARAMETER DESCRIPTION>
   download_path: <PARAMETER DESCRIPTION>
    infer_if_media: : <PARAMETER DESCRIPTION>

Returns 
    <DESCRIPTION OF RETURNED VALUES> 
```

src/marqo/tensor_search/enums.py
wanliAlex: done

src/marqo/tensor_search/utils.py
wanliAlex: solved. thanks for the comments! this makes the code much better.

PR #164, opened 2022-11-08
updated dev and normal readme

vicilliar:
- [ x] The commit message follows our guidelines
- [ x] Docs have been added / updated (for bug fixes / features)

Updated the Dev readme with more detailed instructions (git cloning, requirement installation) as well as AWS tips.
Updated the normal readme with better sample index (removed duplicate documents).

No breaking changes.

* **Other information**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,13 @@ import marqo
 
 mq = marqo.Client(url='http://localhost:8882')
 
+# Delete index if it exists (to avoid duplicate documents)
+try:
+    mq.index("my-first-index").delete()
+except:
+    print("my-first-index doesn't exist yet. creating it.")
+
+# Add documents to the index
 mq.index("my-first-index").add_documents([
     {
         "Title": "The Travels of Marco Polo",
--- a/README.md
+++ b/README.md
@@ -55,13 +55,6 @@ import marqo
 
 mq = marqo.Client(url='http://localhost:8882')
 
-# Delete index if it exists (to avoid duplicate documents)
-try:
-    mq.index("my-first-index").delete()
-except:
-    print("my-first-index doesn't exist yet. creating it.")
-
-# Add documents to the index
 mq.index("my-first-index").add_documents([
     {
         "Title": "The Travels of Marco Polo",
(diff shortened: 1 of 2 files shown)


src/marqo/README.md
```diff
@@ -135,6 +148,26 @@ $ sudo apt-get install -y nvidia-docker2
 ```
 Once this is installed, one of the previous Docker commands can be run (either step B., C., or D.).
 
+### Using Marqo on an AWS machine
```
Jeadie: *AWS EC2

Jeadie: Also, we may want to note that this is not recommended for production use cases

src/marqo/README.md
```diff
@@ -2,7 +2,19 @@
 Thank you for contributing to Marqo! Contributions from the open source community help make Marqo be the tensor engine
 you want. 
 
-### A. Run the Marqo application locally (outside of docker):
+Before starting, clone the github repo
```
Jeadie: This may not be applicable in all options. Maybe we should move this into Option A

README.md
```diff
@@ -55,6 +55,13 @@ import marqo
 
 mq = marqo.Client(url='http://localhost:8882')
 
+# Delete index if it exists (to avoid duplicate documents)
```
Jeadie: This is correct, but it makes the README more convoluted for first time users, who most likely, do not have an index created.

Jeadie (commented):
Some good changes in this, my only concern is the initial delete_index check. I think this makes the initial process longer than needed.

src/marqo/README.md
```diff
@@ -135,6 +149,28 @@ $ sudo apt-get install -y nvidia-docker2
...
+
+To install Docker (through terminal) go to the [Official Docker Website](https://docs.docker.com/engine/install/ubuntu/)
+
+2. Set up SSH Config (to stop timeouts)
+
+Edit the SSH config file with `nano ~/.ssh/config` then insert the line: `ServerAliveInterval 50`
+
+3. Run marqo-os
+```
+sudo docker rm -f marqo-os; sudo docker run -p 9200:9200 -p 9600:9600 -e "discovery.type=single-node" marqoai/marqo-os:0.0.2-arm
```
Jeadie, line 166: nit: you'd have to ensure you make an Arm Ec2 instance (which I think is the default).

Jeadie (approved):
This is a good PR.

PR #181, opened 2022-11-21
Get indexes

pandu-k:

Feature 

No way to retrieve all indexes

Get indexes endpoint reveals all available indexes. Also bumps the Marqo version to 0.0.8

No

Yes

* **Other information**:

### DIFF
--- a/src/marqo/tensor_search/backend.py
+++ b/src/marqo/tensor_search/backend.py
@@ -1,5 +1,6 @@
 """Communication with Marqo's persistence and search layer (OpenSearch)"""
 import json
+import typing
 from marqo.tensor_search.models.index_info import IndexInfo
 # client-specific modules - we may want to replace these:
 from marqo._httprequests import HttpRequests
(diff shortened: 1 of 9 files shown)


src/marqo/tensor_search/backend.py
```diff
@@ -149,5 +149,8 @@ def get_cluster_indices(config: Config):
     """Gets the name of all indices"""
     res = HttpRequests(config).get(path="_aliases")
     indices = set(res.keys())
-    relevant_indices = indices - constants.INDEX_NAMES_TO_IGNORE
+    indices_no_protected_prefixes = {index_name for index_name in indices
```
Jeadie, line 165: nit (i.e. not in this PR): this logic probably should be in a separate function.

Jeadie (approved):
Ideas from this PR. We should include in the PR template placeholders to point to related changes in `py-marqo` and `marqodocs`, that way, when someone looks at the PR they can a) make sure all the related changes will also be there b) use the docs and the client to understand what the change in `marqo` means

jn2clark:
> Ideas from this PR. We should include in the PR template placeholders to point to related changes in `py-marqo` and `marqodocs`, that way, when someone looks at the PR they can a) make sure all the related changes will also be there b) use the docs and the client to understand what the change in `marqo` means

i like this idea!

PR #193, opened 2022-11-28
Docker run examples use the latest tags

pandu-k:
- [X] Docs have been added / updated (for bug fixes / features)

Docs update

Ran DIND api test against the latest tag, to confirm it is OK.


vicilliar (approved):
Looks good! Just confirming: We don't have a marqo-os version tagged "latest" yet, so those will have to be manually updated when future versions come out?

pandu-k:
> Looks good! Just confirming: We don't have a marqo-os version tagged "latest" yet, so those will have to be manually updated when future versions come out?

Correct. We don't foresee that many updates to marqo-os. If we end up frequently updating this then we can switch to using a latest tag for that, too

Issue #189, opened 2022-11-23
[BUG] Improve error messaging on image download fail


Jeadie:
**Describe the bug**
With an image url that has non-standard issues, marqo fails because the content is not an image. This leads to an exception message in Marqo that is not informative for users. 

**To Reproduce**
1. Create image index and use image URL as defined below.
```python
mq.index("my-multimodal-index").add_documents([{
    "My Image": "https://upload.wikimedia.org/wikipedia/commons/9/92/2016-11-13_01_Kjeungskj%C3%A6r_Lighthouse%2C_S%C3%B8r-Tr%C3%B8ndelag%2C_Norway.jpg",
    "Description": "The hippopotamus, also called the common hippopotamus or river hippopotamus, is a large semiaquatic mammal native to sub-Saharan Africa",
    "_id": "hippo-facts"
}])
```
2. Observe error:
```
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/Users/jeadie/Github/marqo/src/marqo/s2_inference/s2_inference.py", line 37, in vectorise
    vectorised = available_models[model_cache_key].encode(content, normalize=normalize_embeddings, **kwargs)
  File "/Users/jeadie/Github/marqo/src/marqo/s2_inference/clip_utils.py", line 231, in encode
    return self.encode_image(inputs, normalize=normalize)
  File "/Users/jeadie/Github/marqo/src/marqo/s2_inference/clip_utils.py", line 200, in encode_image
    image_input = [format_and_load_CLIP_image(images)]
  File "/Users/jeadie/Github/marqo/src/marqo/s2_inference/clip_utils.py", line 81, in format_and_load_CLIP_image
    img = _load_image_from_path(image)
  File "/Users/jeadie/Github/marqo/src/marqo/s2_inference/clip_utils.py", line 60, in _load_image_from_path
    img = Image.open(requests.get(image, stream=True).raw)
  File "/Users/jeadie/Github/marqo/venv/lib/python3.8/site-packages/PIL/Image.py", line 3147, in open
    raise UnidentifiedImageError(
PIL.UnidentifiedImageError: cannot identify image file <_io.BytesIO object at 0x2ad868270>

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/jeadie/Github/marqo/venv/lib/python3.8/site-packages/uvicorn/protocols/http/httptools_impl.py", line 419, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
[...]

PR #198, opened 2022-11-29
Check HTTP status for remote image paths

Jeadie:
Introduce better error messages when downloading an image fails.

https://github.com/marqo-ai/marqo/issues/189

Check response object `resp.ok` before parsing the payload of the image response.

No

Yep

* **Other information**:
- I also updated docstrings and changed the name `_load_image_from_path` to `load_image_from_path` because it is not a private method (used out of file).

### DIFF
--- a/src/marqo/s2_inference/clip_utils.py
+++ b/src/marqo/s2_inference/clip_utils.py
@@ -6,7 +6,7 @@ import requests
 import numpy as np
 import clip
 import torch
-from PIL import Image
+from PIL import Image, UnidentifiedImageError
 import open_clip
 
 from marqo.s2_inference.types import *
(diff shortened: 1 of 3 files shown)


src/marqo/s2_inference/clip_utils.py
```diff
@@ -40,26 +40,29 @@ def format_and_load_CLIP_images(images: List[Union[str, ndarray, ImageType]]) ->
...
-    if os.path.isfile(image):
-        img = Image.open(image)
-    elif validators.url(image):
-        img = Image.open(requests.get(image, stream=True).raw)
+    if os.path.isfile(image_path):
+        img = Image.open(image_path)
+    elif validators.url(image_path):
+        resp = requests.get(image_path, stream=True)
+        if not resp.ok:
+            raise ValueError(f"image url {image_path} returned a {resp.status_code}. Reason {resp.reason}")
```
jn2clark: it might be worth changing to a better excpetion.
PIL already has one we could use
`from PIL import UnidentifiedImageError`

pandu-k (changes requested):
An error that is handled by Marqo web errors (such as the one Jesse suggested) would be caught and returned as an HTTP error

pandu-k (approved):

PR #232, opened 2022-12-14
Updated classifiers 

anukaal:
Updated classifiers for python 3.9


pandu-k:
Hi @anukaal . We don't use these classifiers, because the Marqo project is now distributed through Docker. 

However, our client is still distributed via PyPi. Feel free to raise this issue there: https://github.com/marqo-ai/py-marqo/issues

Issue #227, opened 2022-12-13
[ENHANCEMENT] trigger an exact match when the user uses search_method=LEXICAL and has "" in the query


tomhamer:
**Is your feature request related to a problem? Please describe.**
When searching text using Marqo, many users would like to allow the user an option like on google to use double quotes in their query and trigger an exact match. For example:

query: '"dog" with white hair' -> when the user triggers this with search_method=LEXICAL, they should get back only results that contain the exact word dog, and then execute BM25 as normal on the rest of the words in the query.

**Describe the solution you'd like**
We can enable this functionality through using the query_string parameter rather than the match parameter when doing lexical search here:

https://github.com/marqo-ai/marqo/blob/mainline/src/marqo/tensor_search/tensor_search.py#L869

**Describe alternatives you've considered**

**Additional context**

PR #243, opened 2022-12-18
add query strings to text

tomhamer:
Fix for https://github.com/marqo-ai/marqo/issues/227

When lexical search is used, search queries are treated as match queries. If the user specifies an exact match with a query like 'two "dog"' (which would require only docs with the exact phrase "dog" in it), the user still gets results with the other token "two" included. 

Now we move it to use query_string. This means that users are able to specify double quotes to enable an exact match.

No

Yes (there were some errors though I believe with the pipeline itself though - will follow up with the team).

* **Other information**:

- [Not yet] Docs have been added / updated (for bug fixes / features)
   - will append the required docs changes to this PR, posting here though for first pass.


pandu-k (changes requested):
Thanks for this - quite a simple solution!

Can a few tests be added? Especially around possible DSL usage - think if there can be any exploits.

vicilliar:
> 

Unit tests added to ensure query string functionality works properly. Other unit tests all passed. Haven't been able to observe any exploits yet.

pandu-k (approved):

pandu-k (commented):
Waiting on testing for urls and email addresses

vicilliar:
Notes on this solution (could be further improved):

1. Double quotes don't necessarily force a search term to be in the results. For example: 'two "dog"' will still return some results with "two" and some with "dog", following query string logic.
2. Query strings throw errors for invalid syntax. We may want to use simple query strings instead, which is less strict (ignores syntax errors)

pandu-k (changes requested):
Waiting for fix with URLs and email addresses

vicilliar:
Ready for merge.
Other enhancements turned into a new issue here: https://github.com/marqo-ai/marqo/issues/274
Documentation added. Commit here: https://github.com/marqo-ai/marqodocs/commit/0eccbd453dfab97a1ee43481a574daec1eb7a4b0

pandu-k (approved):

PR #245, opened 2022-12-20
[Onnx clip]Adding the clip_onnx to our avaible models for faster inference

wanliAlex:
features

no onnx model for clip

we have onnx model for clip model

no

not yet

no

not yet 

* **Other information**:
the model passes the performance test on a 50k image dataset
the detailed performance can be checked at [link](https://github.com/wanliAlex/fast-clip-research)
check this [site](https://huggingface.co/Marqo) for available models on huggingface 
check this [repo](https://github.com/marqo-ai/onnx-clip-converter) for onnx model generation

### DIFF
--- a/src/marqo/s2_inference/onnx_clip_utils.py
+++ b/src/marqo/s2_inference/onnx_clip_utils.py
@@ -69,25 +69,7 @@ def _load_image_from_path(image: str) -> ImageType:
     if os.path.isfile(image):
         img = Image.open(image)
     elif validators.url(image):
-        start = timer()
-        try:
-            f = requests.get(image, stream=True, timeout=0.01).raw
-        except requests.exceptions.Timeout:
-            logger.info(f"The http request to image = {image} timed out.")
-            pass
-        end = timer()
-        print(f"Http Request time = {round((end - start)*1000)}ms")
-
-        start = timer()
-        img = Image.open(f)
-        end = timer()
-        print(f"Open Image time = {round((end - start)*1000)}ms")
-        cpu_usage = (load15 / os.cpu_count()) * 100
-        print("The CPU usage is : ", cpu_usage)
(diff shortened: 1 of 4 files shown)


src/marqo/s2_inference/onnx_clip_utils.py
```diff
@@ -0,0 +1,286 @@
...
+import open_clip
+from huggingface_hub import hf_hub_download
+from marqo.s2_inference.types import *
+from marqo.s2_inference.logger import get_logger
+import onnxruntime as ort
+
+logger = get_logger(__name__)
+
+_HF_MODE_DOWNLOAD = {
+    '''
```
pandu-k: Typo? Should it be `_HF_MODEL_DOWNLOAD `?

pandu-k (changes requested):
The main piece of feedback is that onnx_clip_utils seems to be an outdated copy of clip_utils. Can you make ONNX_CLIP a subclass of CLIP? Having two copies of the same code isn't very [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) - unless you want to decouple the APIs for these two types of models.

We can already see issues with copying the code: clip_utils has already had an update with better error message, which hasn't propagated to your code: https://github.com/marqo-ai/marqo/blob/mainline/src/marqo/s2_inference/clip_utils.py#L62

pandu-k:
Thanks for your work on this PR @wanliAlex !

Just have a question:

Has this feature (not just the models in isolation) been tested at scale? We need to ensure that indexing and retrieval works as expected as part of the Marqo application.

wanliAlex:
> 

Yes, the end to end __indexing__ and __searching__ are tested on a 50k image dataset.

pandu-k (approved):
Thanks for this PR @wanliAlex !

PR #281, opened 2023-01-23
Concurrent image downloads 

pandu-k:
Feature

Images download serially in the add_documents call

Downloads the images beforehand, in a function that spins up 20 threads to download images

No

Yes. +Manual testing against NFT-style images.


src/marqo/s2_inference/clip_utils.py
```diff
@@ -341,10 +340,48 @@ def encode_image(self, images: Union[str, ImageType, List[Union[str, ImageType]]
         return self._convert_output(outputs)
 
 
-
-
-
-
+def get_multilingual_clip_properties() -> Dict:
```
vicilliar, line 492: is it good practice to have `get_multilingual_clip_properties` in `clip_utils` instead of `model_registry` like the others? would we eventually want to move, let's say `_get_onnx_clip_properties` to `onnx_clip_utils` as well?

src/marqo/s2_inference/model_registry.py
```diff
@@ -497,47 +497,7 @@ def _get_sbert_onnx_properties() -> Dict:
...
-                "name": "multilingual-clip/LABSE-Vit-L-14",
-                "visual_model": "openai/ViT-L/14",
-                "textual_model": 'M-CLIP/LABSE-Vit-L-14',
-                "dimensions": 768,
-                "type": "multilingual_clip",
-            }
-    }
-
-    return MULTILINGUAL_CLIP_PROPERTIES
+    return get_multilingual_clip_properties()
```
vicilliar: why do this instead of having functions call `get_multilingual_clip_properties` straight from `clip_utils` instead?

src/marqo/tensor_search/add_docs.py
```diff
@@ -0,0 +1,48 @@
...
+def threaded_download_images(allocated_docs: List[dict], image_repo: dict) -> None:
+    """A thread calls this function to download images for its allocated documents
+
+    This should be called only if treat URLs as images is True
+    """
+    for doc in allocated_docs:
+        for field in list(doc):
+            if isinstance(doc[field], str) and _is_image(doc[field]):
+                try:
+                    image_repo[doc[field]] = load_image_from_path(doc[field])
```
vicilliar: would this not cause overwrites? for example if Doc 1 and Doc 2 both have "Field A", they could both save their images under `image_repo["Field A"]`

src/marqo/s2_inference/clip_utils.py
pandu-k: It is nice that they were centralised in the model_registry. This move was necessary for Marqo to run properly to avoid a circular import error. 

In the future, there are a few ways to split it out. 
Option 1. We can put the model load mappings into a separate file: https://github.com/marqo-ai/marqo/blob/mainline/src/marqo/s2_inference/model_registry.py#L1565
Option 2. We can make the properties associated with model class

src/marqo/s2_inference/model_registry.py
pandu-k: actioned

src/marqo/s2_inference/clip_utils.py
pandu-k: But I think refactoring this should be its own PR

src/marqo/tensor_search/add_docs.py
pandu-k: `image_repo[doc[field]]` - the key of image_repo is the value associated with the `field`  in the doc dict. This would be a URL. So the image_repo would have this structure: 
```
{
    "https://s3.aws.com/my_image1.png": <PIL_image>, 
    "https://s3.aws.com/my_image2.png": <PIL_image> 
}
```

src/marqo/s2_inference/model_registry.py
pandu-k: good point

src/marqo/tensor_search/add_docs.py
vicilliar: ah okay, got it.

src/marqo/s2_inference/clip_utils.py
wanliAlex: thanks for the discussion. I had this circular import problem when I was writing the code.

src/marqo/tensor_search/add_docs.py
```diff
@@ -0,0 +1,48 @@
...
+    This should be called only if treat URLs as images is True
+
+    Returns:
+         An image repo: a dict <image pointer>:<image data>
+    """
+
+    docs_per_thread = math.ceil(len(docs)/thread_count)
+    copied = copy.deepcopy(docs)
+    image_repo = dict()
+    thread_allocated_docs = [copied[i: i + docs_per_thread] for i in range(len(copied))[::docs_per_thread]]
```
wanliAlex, line 52: do we need ?
```python
[copied[i: max(i+docs_per_thread, len(copied))] for i in range(len(copied))[::docs_per_thread]]
```

wanliAlex (commented):
Do we need to add some tests as test_image_concurrent_downloading speed to make sure we get better throughput?

src/marqo/s2_inference/clip_utils.py
```diff
@@ -421,7 +420,7 @@ def __init__(self, model_type: str = "multilingual-clip/ViT-L/14", device: str =
                             truncate: bool = True, **kwargs) -> None:
 
         self.model_name = model_type
-        self.model_info = model_registry._get_multilingual_clip_properties()[self.model_name]
+        self.model_info = get_multilingual_clip_properties()[self.model_name]
```
wanliAlex, line 425: we can use `self.model_info = kwarge.get("model_properties", dict())` and remove the added model_registry information later.

src/marqo/s2_inference/clip_utils.py
```diff
@@ -488,10 +487,48 @@ def encode_image(self, images: Union[str, ImageType, List[Union[str, ImageType]]
...
+        "multilingual-clip/LABSE-Vit-L-14":
+            {
+                "name": "multilingual-clip/LABSE-Vit-L-14",
+                "visual_model": "openai/ViT-L/14",
+                "textual_model": 'M-CLIP/LABSE-Vit-L-14',
+                "dimensions": 768,
+                "type": "multilingual_clip",
+            }
+    }
+    return MULTILINGUAL_CLIP_PROPERTIES
```
wanliAlex, line 534: this part can be removed

wanliAlex (changes requested):

src/marqo/s2_inference/clip_utils.py
pandu-k: Let's move this into a separate refactor PR, to decouple these

src/marqo/s2_inference/clip_utils.py
pandu-k: Let's move this into a separate refactor PR, to decouple these

src/marqo/tensor_search/add_docs.py
pandu-k: Python slicing is OK even if we slice past the last element
image

wanliAlex (approved):

Issue #317, opened 2023-02-13
[ENHANCEMENT] Multi-modal tensors combining


wanliAlex:
**Is your feature request related to a problem? Please describe.**
Currently, we don't have a way to combine the embedded tensors from multiple fields into one.
We would like an endpoint to allow users to combine multiple fields into one tensor.

**Describe the solution you'd like**
A detailed design document will be provided.

wanliAlex:
# Multi-modal Tensors Combining Design Document
## Introduction

Currently, for a document with the following format in tensor search at indexing,
```python
document = {"image_field_1" : "https://image-fieled-1.png", #v1
		   "image_field_2" : "https://image-fieled-2.png", #v2
		   "Text_field_1" : "text-field-1-random-random", #v3
		   "Text_field_2" : "text-field-2-random-random", #4
		   } 
```
we will need to inference each field once, and generate **4** tensors for this document. We denote the 4 tensors as $v_1, v_2, v_3, v_4$, respectively.

In search, 
```python
query = "i want to search something" #q1
```
we will vectorise the query into tensor as $q_1$ and compare the distance with the above 4 tensors to get the `_score`.
```math
score = \max \{ distance(v_i, q_1) | i \in \{ 1,2,3,4 \} \},
```
where $distance(\cdot,\cdot)$ is a *__linear__* distance measurement method (e.g., dot product).

In this design, we would like to find a way to combine several multi-modal fields into one. For example, for the
document mentioned above, we add a new field called `multi-modal-combination':
```python
document = {"image_field_1" : "https://image-fieled-1.png", #v1
		   "image_field_2" : "https://image-fieled-2.png", #v2
		   "text_field_1" : "text-field-1-random-random", #v3
		   "text_field_2" : "text-field-2-random-random", #v4
		   "multimodal_combination" : {
                                         "image_field_1": w1, #a float number,
					 "text_field_1": w3, #a float number,
					 "text_field_2": w4, #a float number} # v^*
		   } 
```

For this document, we will need to do inference 4 times to compute $v_1, v_2, v_3, v_4$ and compute the multi-modal-combination tensor $v^*$ as:
```math
v^* = (w_1 *v_1 + w_3*v_3 + w_4 *v_4) / (w_1 +w_3 +w_4)
```
and for the same search query, the score will then be computed as:
```math
score = \max \{distance(v_2, q_1), distance(v^*, q_1) \}.
```
[...]

PR #332, opened 2023-02-17
[feature] Multimodal tensor combination

wanliAlex:
support multimodal-tensor-combination

we don't have multimodal-tensor-combination

we can combine tensor from multimodal fields.

no

no

no

will have

* **Other information**:
this PR is related to the discussion here https://github.com/marqo-ai/marqo/issues/317


wanliAlex:
In the manual test below, 

```python
  def test_multimodal_tensor_combination_score(self):

      def get_score(document):
          try:
              tensor_search.delete_index(config=self.config, index_name=self.index_name_1)
          except IndexNotFoundError as e:
              pass

          tensor_search.create_vector_index(
              index_name=self.index_name_1, config=self.config, index_settings={
                  IndexSettingsField.index_defaults: {
                      IndexSettingsField.model: "ViT-B/32",
                      IndexSettingsField.treat_urls_and_pointers_as_images: True,
                      IndexSettingsField.normalize_embeddings: False
                  }
              })

          tensor_search.add_documents(config=self.config, index_name=self.index_name_1, docs=[document], auto_refresh=True)
          self.assertEqual(1, tensor_search.get_stats(config=self.config, index_name=self.index_name_1)["numberOfDocuments"])
          res = tensor_search.search(config=self.config, index_name=self.index_name_1,
                                     text="", result_count=1)
          print(res)
          print("_score:", res["hits"][0]["_score"])
          return res["hits"][0]["_score"]

      score_1 = get_score({
              "text_field": "A rider is riding a horse jumping over the barrier.",
              #"image_field": "https://raw.githubusercontent.com/marqo-ai/marqo/mainline/examples/ImageSearchGuide/data/image1.jpg",
          })

      print("----"*20)
      score_2 = get_score({
              #"text_field": "A rider is riding a horse jumping over the barrier.",
              "image_field": "https://raw.githubusercontent.com/marqo-ai/marqo/mainline/examples/ImageSearchGuide/data/image1.jpg",
          })
      print("----" * 20)
      score_3 = get_score({
          "combo_text_image": {
              "A rider is riding a horse jumping over the barrier.": {
                  "weight": 0.5,
              },
              "https://raw.githubusercontent.com/marqo-ai/marqo/mainline/examples/ImageSearchGuide/data/image1.jpg": {
                  "weight": 0.5,
              },
      }
      })

      assert (score_3 > min(score_1, score_2)) and (score_3 < max(score_1,score_2))
``` 

we have output as：
image


check the tensors marked by the red pen and the scores.

@pandu-k

pandu-k:
Running unit tests: https://github.com/marqo-ai/marqo/actions/runs/4228946165

pandu-k:
> 

Hi Li, yep that transformation makes sense to me

src/marqo/tensor_search/validation.py
```diff
@@ -343,3 +345,48 @@ def validate_settings_object(settings_object):
...
+        field_content: the field when it is a dict, especially used for multimodal tensor combination field
+        is_non_tensor_field: for multimodal tensor combination field, this should be True
+
+    Returns:
+        True or raise an error
+    '''
+    if len(field_content) < 2:
+        raise InvalidArgError(
+            f"The length of field_content `{field_content}` is smaller than 2. This is not a valid format of field content."
+            f"If you aim to use multimodal-tensor-combination, please check `https://docs.marqo.ai/0.0.14/` for more info.")
```
pandu-k: 1. this will be in release `0.0.15`
2. Users might not think that a JSON object has a length. It might make more sense to say "The multimodal-combination object `{field_content}` must have at least two fields."

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1657,3 +1663,118 @@ def get_cuda_info() -> dict:
...
+def vectorise_multimodal_combination_field(field: str, field_content:Dict[str,dict], copied:Dict, unsuccessful_docs,
+                                       total_vectorise_time, i, doc_id, selected_device,index_info):
+    '''
+    This function is used to vectorise multimodal combination field. The field content should
+    have the following structure:
+    field_conent = {"tensor_field_one" : {"weight":0.5, "parameter": "test-paramater-1"},
+                    "tensor_field_two" : {"weight": 0.5, parameter": "test-parameter-2"}},
+    Over all this is a simplified version of the vectorise pipeline in add_documents. Specifically,
+    1. we don't do any chunking here.
+    2. we don't use image repo for concurrent downloading.
```
pandu-k, line 1733: We need to make create an issue for this. Ideally, concurrent image downloading also downloads all image for multimodal combination field. The images will then be passed to this function

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1657,3 +1663,118 @@ def get_cuda_info() -> dict:
...
+    Args:
+        field_content: the field content that is a dictionary
+        copied: the copied document
```
pandu-k, line 1736: maybe just call this param `doc` - does `copied` have any special meaning or implication here?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1657,3 +1663,118 @@ def get_cuda_info() -> dict:
...
+        unsuccessful_docs: a list to store all the unsuccessful documents
+        total_vectorise_time: total vectorise time in the main body
+        i: an interator in the main body
```
pandu-k: Be more descriptive about this parameter. You might even want to rename it something like `doc_position`

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1657,3 +1663,118 @@ def get_cuda_info() -> dict:
...
+        doc_id: the document id
+        selected_device: device from main body
+        index_info: index_info from main body
+    Returns: return chunks: the appended chunks from main body
```
pandu-k: is this accurate? Isn't return chunks just the chunks (singleton-list of a dict with embedding and content) generated by this function?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1657,3 +1663,118 @@ def get_cuda_info() -> dict:
         raise errors.HardwareCompatabilityError(message=str(
             "ERROR: cuda is not supported in your machine!!"
         ))
+
+
+def vectorise_multimodal_combination_field(field: str, field_content:Dict[str,dict], copied:Dict, unsuccessful_docs,
+                                       total_vectorise_time, i, doc_id, selected_device,index_info):
```
pandu-k: Do we need `unsuccessful_docs` to be given to this function? Can't this function just generate a list of unsuccessful docs, which the calling function will append its own list of `unsuccessful_docs`? It would be better if functions return brand new objects, rather than modifying existing ones

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1657,3 +1663,118 @@ def get_cuda_info() -> dict:
...
+        copied: the copied document
+        unsuccessful_docs: a list to store all the unsuccessful documents
+        total_vectorise_time: total vectorise time in the main body
+        i: an interator in the main body
+        doc_id: the document id
+        selected_device: device from main body
+        index_info: index_info from main body
+    Returns: return chunks: the appended chunks from main body
+    document_is_valid:  if the document is a valid
+    unsuccessful_docs: appended unsucessful_docs
```
pandu-k, line 1746: Again we want to return new objects rather than updating an existing one

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1657,3 +1663,118 @@ def get_cuda_info() -> dict:
...
+            #     unsuccessful_docs.append(
+            #         (i, {'_id': doc_id, 'error': e.message,
+            #              'status': int(errors.InvalidArgError.status_code),
+            #              'code': errors.InvalidArgError.code})
+            #     )
+            #     return chunks_to_append, document_is_valid, unsuccessful_docs, total_vectorise_time
+
+
+            try:
+                if infer_if_image is False:
```
pandu-k: I don't get what this try except block is doing?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1657,3 +1663,118 @@ def get_cuda_info() -> dict:
...
+                    raise errors.InvalidArgError(
+                        f"The setting `treat_urls_and_pointers_as_images` is `{infer_if_image}` for a multimodal_tensor_combination field."
+                        f"This is not supported. Please change the setting `treat_urls_and_pointers_as_images` as `True` for multimodal_tensor_combination fields. "
+                    )
```
pandu-k: This shouldn't be an error - users should be able to have `treat_urls_and_pointers_as_images=False` and use this feature

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1657,3 +1663,118 @@ def get_cuda_info() -> dict:
...
+    unsuccessful_docs: appended unsucessful_docs
+    total_vectorise_time: new total vectorise time
+    '''
+    # field_conent = {"tensor_field_one" : {"weight":0.5, "parameter": "test-paramater-1"},
+    #                 "tensor_field_two" : {"weight": 0.5, parameter": "test-parameter-2"}},
+    document_is_valid = True
+    field_vector_weight = []
+    chunks_to_append = []
+    for sub_content, sub_content_para in field_content.items():
+        if isinstance(sub_content, (str, Image.Image)):
```
pandu-k: can we just do a check like this 
```
if not isinstance(sub_content, (str, Image.Image)):
     raise InvalidArgError(...)
````
this just saves as an indent

Comment (pandu-k) [reply src/marqo/tensor_search/tensor_search.py]: Or add a doc to the unsuccessful docs

File: src/marqo/tensor_search/tensor_search.py [hunk 11]
```diff
@@ -1657,3 +1663,118 @@ def get_cuda_info() -> dict:
...
+                         'code': errors.InvalidArgError.code})
+                )
+                return chunks_to_append, document_is_valid, unsuccessful_docs, total_vectorise_time
+            try:
+                start_time = timer()
+                vector_chunks = s2_inference.vectorise(
+                    model_name=index_info.model_name,
+                    model_properties=_get_model_properties(index_info), content=content_chunks,
+                    device=selected_device, normalize_embeddings=normalize_embeddings,
+                    infer=infer_if_image)
```
pandu-k: We can actually vectorise the chunks together. See here: https://github.com/marqo-ai/marqo/blob/mainline/src/marqo/tensor_search/tensor_search.py#L1246

pandu-k: This should save inference calls

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1657,3 +1663,118 @@ def get_cuda_info() -> dict:
...
+            except (s2_inference_errors.UnknownModelError,
+                    s2_inference_errors.InvalidModelPropertiesError,
+                    s2_inference_errors.ModelLoadError) as model_error:
+                raise errors.BadRequestError(
+                    message=f'Problem vectorising query. Reason: {str(model_error)}',
+                    link="https://marqo.pages.dev/latest/Models-Reference/dense_retrieval/"
+                )
+            except s2_inference_errors.S2InferenceError:
+                document_is_valid = False
+                image_err = errors.InvalidArgError(message=f'Could not process given image: {field_content}')
```
pandu-k: How do we know that an image error caused this

pandu-k: Can we make it more general, and pass through the message from the S2InferenceError?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1657,3 +1663,118 @@ def get_cuda_info() -> dict:
...
+        index_info: index_info from main body
+    Returns: return chunks: the appended chunks from main body
+    document_is_valid:  if the document is a valid
+    unsuccessful_docs: appended unsucessful_docs
+    total_vectorise_time: new total vectorise time
+    '''
+    # field_conent = {"tensor_field_one" : {"weight":0.5, "parameter": "test-paramater-1"},
+    #                 "tensor_field_two" : {"weight": 0.5, parameter": "test-parameter-2"}},
+    document_is_valid = True
+    field_vector_weight = []
```
pandu-k: A comment above this line, like `# field_vector_weight  is a list of <weight, vector> pairs` would be helpful

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1657,3 +1663,118 @@ def get_cuda_info() -> dict:
...
+                    (i, {'_id': doc_id, 'error': image_err.message, 'status': int(image_err.status_code),
+                         'code': image_err.code})
+                )
+                return chunks_to_append, document_is_valid, unsuccessful_docs, total_vectorise_time
+
+            field_vector_weight.append((sub_content_para["weight"], vector_chunks))
+
+    vector_chunk = np.squeeze(
+                np.sum([np.array(vector) * weight for weight, vector in field_vector_weight], axis=0) / np.sum(
+            [weight for weight, vector in field_vector_weight])).tolist()
```
pandu-k: is it possible for 0 division error? if we have weights = [-1, 1]

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1657,3 +1663,118 @@ def get_cuda_info() -> dict:
...
+
```
pandu-k, line 1859: Because we only return one chunk we could just return a chunk, and then append it in the calling function.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -51,6 +51,7 @@
 from marqo.tensor_search.index_meta_cache import get_cache, get_index_info
```
pandu-k, line 51: We need to extract the dict keys and add them to the filtering data. See below:
https://github.com/marqo-ai/marqo/blob/mainline/src/marqo/tensor_search/tensor_search.py#L480

tests/tensor_search/test_multimodal_tensor_combination.py
```diff
@@ -0,0 +1,289 @@
...
+            {
+                "combo_text_image": {
+                    "A rider is riding a horse jumping over the barrier.": 0.5,
+                    "https://raw.githubusercontent.com/marqo-ai/marqo/mainline/examples/ImageSearchGuide/data/image1.jpg": {
+                        "weight": 0,
+                    },
+                },
+                "_id": "123",
+            }, ], auto_refresh=True)
+        assert res["errors"] == True
```
pandu-k: Can you assert that you can't find it in the index?

tests/tensor_search/test_multimodal_tensor_combination.py
```diff
@@ -0,0 +1,289 @@
+import unittest.mock
```
pandu-k, line 1: 1. Can we get the tensor_facets (by get_document(show_vectors=True)) and assert they are correct? 
2. Can we get the original documents from Marqo-OS and do assertions on them? Is the meta data correct?

pandu-k (changes requested):

src/marqo/tensor_search/validation.py
wanliAlex: done

src/marqo/tensor_search/tensor_search.py
wanliAlex: let me try to use image repo in this method.

src/marqo/tensor_search/tensor_search.py
wanliAlex: this is really a good point! let me ask jesse about the normalization for the weight

src/marqo/tensor_search/tensor_search.py
wanliAlex: yes, i am returning one chunk now, like `chunks_to_append`.

src/marqo/tensor_search/tensor_search.py
wanliAlex: yes, forget this point after the merge, i will put them in a list.

wanliAlex: solved

src/marqo/tensor_search/tensor_search.py
wanliAlex: Comments from jesse: 1. If normalize is true then we should normalize after combining, 2. No, they can be negative 3. No, leave unchanged unless normalize is true, in which case just normalize after adding the vectors.  I will revise based on these.

src/marqo/tensor_search/tensor_search.py
wanliAlex: yes, we should return chunks_to_append from this function.

tests/tensor_search/test_multimodal_tensor_combination.py
wanliAlex: done

tests/tensor_search/test_multimodal_tensor_combination.py
wanliAlex: done

src/marqo/tensor_search/tensor_search.py
wanliAlex: image repo added

src/marqo/tensor_search/add_docs.py
```diff
@@ -48,6 +48,20 @@ def threaded_download_images(allocated_docs: List[dict], image_repo: dict,
                 except PIL.UnidentifiedImageError as e:
                     image_repo[doc[field]] = e
                     continue
+            # For multimodal tensor combination
+            elif isinstance(doc[field], dict):
+                for sub_field in list(doc[field]):
+                    if _is_image(sub_field):
+                        if sub_field in image_repo:
```
pandu-k, line 55: nit: combine these two if statements into one with an `and` - this prevents having to indent too deeply and makes it neater

src/marqo/tensor_search/tensor_search.py
```diff
@@ -482,9 +482,12 @@ def add_documents(config: Config, index_name: str, docs: List[dict], auto_refres
         for key, value in copied.items():
```
pandu-k, line 503: nit: the comment above this line will be outdated

src/marqo/tensor_search/tensor_search.py
```diff
@@ -627,6 +631,19 @@ def add_documents(config: Config, index_name: str, docs: List[dict], auto_refres
                     })
 
             # Add chunks_to_append along with doc metadata to total chunks
+
+            elif isinstance(field_content, dict):
+                chunks_to_append, document_is_valid, unsuccessful_doc_to_append, vectorise_time_to_add = \
+                    vectorise_multimodal_combination_field(field, field_content, copied, \
```
pandu-k: nit: unnecessary backslash

src/marqo/tensor_search/tensor_search.py
```diff
@@ -627,6 +631,19 @@ def add_documents(config: Config, index_name: str, docs: List[dict], auto_refres
...
+                                                           i, doc_id, selected_device, index_info, image_repo)
+
+                if document_is_valid is False:
+                    for unsuccessful_doc in unsuccessful_doc_to_append:
+                        unsuccessful_docs.append(unsuccessful_doc)
```
pandu-k: This shouldn't be a loop right?

tests/tensor_search/test_multimodal_tensor_combination.py
```diff
@@ -0,0 +1,362 @@
...
+
```
pandu-k, line 3: Can we also check if meta data is being added correctly?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -627,6 +631,19 @@ def add_documents(config: Config, index_name: str, docs: List[dict], auto_refres
                     })
 
             # Add chunks_to_append along with doc metadata to total chunks
+
+            elif isinstance(field_content, dict):
+                chunks_to_append, document_is_valid, unsuccessful_doc_to_append, vectorise_time_to_add = \
```
pandu-k: Can we have completely new names of objects returned from the function? Name clashes might have unexpected results. In this case, "document_is_valid" in particular, but all should have unique names that gives the reader information about what the objects is (e.g: `vectorise_time_to_add ` -> `combo_vectorise_time_to_add`)

pandu-k: Also chunks to append (plural chunks) isn't accurate right? `combo_chunk_to_append` is more accurate, right?

src/marqo/tensor_search/tensor_search.py
pandu-k: Should we do normalise == True if the index settings have normalise == True - Jesse said we should for a similar situation on the weighted query combo (even tho we also normalise later)

src/marqo/tensor_search/tensor_search.py
pandu-k: why don't we just return a single dict, rather than a list?

src/marqo/tensor_search/validation.py
```diff
@@ -343,3 +345,49 @@ def validate_settings_object(settings_object):
             f"Error validating index settings object. Reason: \n{str(e)}"
             f"\nRead about the index settings object here: https://docs.marqo.ai/0.0.13/API-Reference/indexes/#body"
         )
+
+
+def validate_dict(field_content: typing.Dict, is_non_tensor_field: bool):
```
pandu-k: Can we make tests for this?

pandu-k (changes requested):

src/marqo/tensor_search/add_docs.py
wanliAlex: i don't think this can be combined. can you check line 43?

src/marqo/tensor_search/tensor_search.py
wanliAlex: updated

src/marqo/tensor_search/tensor_search.py
wanliAlex: updated

src/marqo/tensor_search/tensor_search.py
wanliAlex: make it a tuple and append it in mainbody directly now.

src/marqo/tensor_search/tensor_search.py
wanliAlex: updated

tests/tensor_search/test_multimodal_tensor_combination.py
wanliAlex: done.

src/marqo/tensor_search/validation.py
wanliAlex: done.

tests/tensor_search/test_multimodal_tensor_combination.py
```diff
@@ -0,0 +1,604 @@
...
+            }, ], auto_refresh=True)
+        try:
+            tensor_search.get_document_by_id(config=self.config, index_name=self.index_name_1, document_id="123")
+            raise AssertionError
+        except DocumentNotFoundError:
+            pass
+
+    def test_validate_dict(self):
+        valid_dict = [{"test": {"weight": 0.3}, "test_test": {"weight": -0.3}},
+                      {"test": {"weight": 0.5}, "test_test": {"weight": 0.3}, "test_3": {"weight": 0.6}}]
```
pandu-k: - maybe a test for ints, too: `{"weight": 2}`
- A test for {"weight": -1, "weight": 1}

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1632,28 +1679,266 @@ def get_loaded_models() -> dict:
...
+    doc_copy = copy.deepcopy(doc)
+
+    # 4 list to store content_chunks, text_chunks to image and text.
+    # It can also be two list of tuples, but i don't think they can be dictioanry.
+    # I choose to use 4 lists as they are easy to handle.
+
+
+
+    text_content_chunks = []
+    text_text_chunks = []
```
pandu-k: Can we just keep a single variable, I can't see how `text_content_chunks` would be different from `text_text_chunks`. Also, `text_content` would be fine, as `chunks` is an overloaded term that we want to reduce unnecessary usage of.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1632,28 +1679,266 @@ def get_loaded_models() -> dict:
...
+    # field_conent = {"tensor_field_one" : {"weight":0.5, "parameter": "test-paramater-1"},
+    #                 "tensor_field_two" : {"weight": 0.5, parameter": "test-parameter-2"}},
+    combo_document_is_valid = True
+    field_vector_weight = []
+    combo_vectorise_time_to_add = 0
+    combo_chunk = {}
+    unsuccessful_doc_to_append = tuple()
+
+    # Copy the important mutable objects from main body for safety purpose
+    field_content_copy = copy.deepcopy(field_content)
```
pandu-k: nit: name this multi_modal_object copy

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1632,28 +1679,266 @@ def get_loaded_models() -> dict:
...
+                                 for _device_id in range(torch.cuda.device_count())]}
 
     else:
         raise errors.HardwareCompatabilityError(message=str(
             "ERROR: cuda is not supported in your machine!!"
         ))
+
+
+def vectorise_multimodal_combination_field(field: str, field_content: Dict[str, dict], doc: dict, doc_index: int,
+                                            doc_id:str, selected_device:str, index_info, image_repo):
```
pandu-k: nit: field_content is better called `multi_modal_object`

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1632,28 +1679,266 @@ def get_loaded_models() -> dict:
...
+    # It can also be two list of tuples, but i don't think they can be dictioanry.
+    # I choose to use 4 lists as they are easy to handle.
+
+
+
+    text_content_chunks = []
+    text_text_chunks = []
+
+    image_content_chunks = []
+    image_text_chunks = []
```
pandu-k: Can we be more explicit about what the lists are, here? For example: `PIL_images` and `image_pointer_texts`

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1632,28 +1679,266 @@ def get_loaded_models() -> dict:
...
+                    NsField.treat_urls_and_pointers_as_images]:
+                    if not isinstance(image_repo[sub_content], Exception):
+                        image_data = image_repo[sub_content]
+                    else:
+                        raise s2_inference_errors.S2InferenceError(
+                            f"Could not find image found at `{sub_content}`. \n"
+                            f"Reason: {str(image_repo[sub_content])}"
+                        )
+                else:
+                    image_data = sub_content
```
pandu-k: This is an odd structure. Can't we start with an _if_ statement like this:
```
if not index_info.index_settings[NsField.index_defaults][ NsField.treat_urls_and_pointers_as_images]:
    text_content = list(multi_modal_object.keys())
    PIL_images = []
    image_urls = []
else:
    # do the logic of dividing content amongst text_content, PIL_images and image_urls
```

pandu-k: It's odd having URLs being vectorised separately from the rest of the text if `index_info.index_settings[NsField.index_defaults][ NsField.treat_urls_and_pointers_as_images]` is `False`

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1632,28 +1679,266 @@ def get_loaded_models() -> dict:
...
+        field_content: the field content that is a dictionary
+        copied: the copied document
+        unsuccessful_docs: a list to store all the unsuccessful documents
+        total_vectorise_time: total vectorise time in the main body
+        doc_index: the index of the document. This is an interator variable `i` in the main body to iterator throught the docs
+        doc_id: the document id
+        selected_device: device from main body
+        index_info: index_info from main body
+    Returns:
+    combo_chunk_to_append: the combo_chunk to be appended to the main body
```
pandu-k: this description is outdated

tests/tensor_search/test_multimodal_tensor_combination.py
wanliAlex: added

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1632,28 +1679,266 @@ def get_loaded_models() -> dict:
...
+    normalize_embeddings = index_info.index_settings[NsField.index_defaults][
+        NsField.normalize_embeddings]
+    infer_if_image = index_info.index_settings[NsField.index_defaults][
+        NsField.treat_urls_and_pointers_as_images]
+
+
+    try:
+        start_time = timer()
+
+        text_vector_chunks = s2_inference.vectorise(
```
pandu-k: `text_vectors` might be enough - to reduce overloading the term `chunks`. Same for the images

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1632,28 +1679,266 @@ def get_loaded_models() -> dict:
...
+
+        image_vector_chunks = s2_inference.vectorise(
+            model_name=index_info.model_name,
+            model_properties=_get_model_properties(index_info), content=image_content_chunks,
+            device=selected_device, normalize_embeddings=normalize_embeddings,
+            infer=infer_if_image)
+
+        end_time = timer()
+        single_vectorise_call = end_time - start_time
+        combo_vectorise_time_to_add += single_vectorise_call
```
pandu-k: Should we do 2x single vectorise call timings, and log these to debug? Also if send an empty list to vectorise(), does it just get skipped?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -627,10 +635,30 @@ def add_documents(config: Config, index_name: str, docs: List[dict], auto_refres
...
 
             # Add chunks_to_append along with doc metadata to total chunks
+
+            elif isinstance(field_content, dict):
+
+                combo_chunk, combo_document_is_valid, unsuccessful_doc_to_append, combo_vectorise_time_to_add = \
+                    vectorise_multimodal_combination_field(field, field_content, copied,
+                                                           i, doc_id, selected_device, index_info, image_repo)
+
+                multimodal_field_for_filter[field] = list(field_content)
```
pandu-k: Can we just directly add it to the copied doc?

src/marqo/tensor_search/tensor_search.py
wanliAlex: done

src/marqo/tensor_search/tensor_search.py
wanliAlex: done

src/marqo/tensor_search/tensor_search.py
wanliAlex: yes, for text we only need on variable. 
I am using two here to be consistent with
https://github.com/marqo-ai/marqo/blob/5dcf4f7de7923fb794d4f69b789e7c24b28ae6b5/src/marqo/tensor_search/tensor_search.py#L540-L541

src/marqo/tensor_search/tensor_search.py
wanliAlex: solved

src/marqo/tensor_search/tensor_search.py
wanliAlex: no we can't because we are iterating through copied here.

src/marqo/tensor_search/tensor_search.py
wanliAlex: done

src/marqo/tensor_search/tensor_search.py
wanliAlex: done

src/marqo/tensor_search/tensor_search.py
wanliAlex: updated

src/marqo/tensor_search/tensor_search.py
wanliAlex: updated

src/marqo/tensor_search/tensor_search.py
pandu-k: Don't worry about it being consistent elsewhere, we need to think from the perspective that this is a standalone function, that should be understandable even if `_vector_text_search()` were to disappear.

wanliAlex: thanks. changed variables names.

src/marqo/tensor_search/tensor_search.py
wanliAlex: actually this is totally different from the single vectorise in mainbody. we are now measuring the time of 2 vectorise calls (text and image) and they are batched inference.

wanliAlex: i change name to `multimodal_inference_call`

src/marqo/tensor_search/backend.py
```diff
@@ -77,6 +78,13 @@ def add_customer_field_properties(config: Config, index_name: str,
     """
     engine = "lucene"
 
+    # check if there is multimodal fie;ds and convert the fields name to a list with the same
+    # format of customer_field_names
+    update_chunk_fields = copy.deepcopy(customer_field_names)
```
pandu-k: maybe call this `knn_field_names` - it better describes its purpose

src/marqo/tensor_search/backend.py
```diff
@@ -135,6 +152,13 @@ def add_customer_field_properties(config: Config, index_name: str,
         new_index_properties[validation.validate_field_name(new_prop)] = {
             "type": type_to_set
         }
+    if len(multimodal_combination_field) > 0:
+        for field_name, sub_field_names in multimodal_combination_field.items():
```
pandu-k: again we can be more descriptive (the children are actually not just field_names, they are actually more complex): 
`for multimodal_field, child_fields in ...`
`    for child_field_name, child_type in child_fields: `

src/marqo/tensor_search/backend.py
```diff
@@ -135,6 +152,13 @@ def add_customer_field_properties(config: Config, index_name: str,
...
+            for sub_field_name, type_to_set in sub_field_names:
+                new_index_properties[validation.validate_field_name(f"{field_name}.{sub_field_name}")] = {
+                    "type" : type_to_set,
+                }
```
pandu-k: Can we have the same structure as the OpenSearch mappings? 
```
new_index_properties[multimodal_field] = {
    "properties": {child_field_name: {"type": child_type}}
}
```
Index info has the responsibility for transforming this into dot. notation for use in queries

src/marqo/tensor_search/backend.py
```diff
@@ -135,6 +152,13 @@ def add_customer_field_properties(config: Config, index_name: str,
         new_index_properties[validation.validate_field_name(new_prop)] = {
             "type": type_to_set
         }
+    if len(multimodal_combination_field) > 0:
```
pandu-k: Note that we only add _new_ fields to the cache, we don't re-add existing ones. See line 148.
Let's apply the same pattern to multimodal fields: 
`new_multimodal_properties = multimodal_properties - existing_properties ` 

(set difference with all existing props should be enough)

src/marqo/tensor_search/enums.py
```diff
@@ -76,6 +76,7 @@ class OpenSearchDataType:
     float = "float"
     integer = "integer"
     to_be_defined = "to_be_defined"  # to be defined by OpenSearch
+    object = "nested"
```
pandu-k: we shouldn't be setting objects as nested

src/marqo/tensor_search/models/index_info.py
```diff
@@ -46,13 +46,29 @@ def get_true_text_properties(self) -> dict:
...
         are true text fields
         """
+        # get_text_properties will naturally return multimodal_fields
+        # Example: left-text_properties   right-true_text_properties
+        # {'Description': {'type': 'text'},                                                         {'Description': {'type': 'text'},
+        #  'Genre': {'type': 'text'},                                                                'Genre': {'type': 'text'},
+        #  'Title': {'type': 'text'},                                                                'Title': {'type': 'text'},
+        #  'my_combination_field': {'properties': {'lexical_field': {'type': 'text'}, ----->         'my_combination_field.lexical_field': {'type': 'text'},
+        #                                          'my_image': {'type': 'text'},                     'my_combination_field.my_image': {'type': 'text'},
+        #                                          'some_text': {'type': 'text'}}}}                   'my_combination_field.some_text': {'type': 'text'}}
```
pandu-k, line 55: Can we add this logic to `get_text_properties()`. Then, we can still get the flattened multimodal properties in this function from `get_text_properties()`, and then further filter to only keep `text` type fields.

src/marqo/tensor_search/constants.py
```diff
@@ -6,8 +6,10 @@
     'security-auditlog-'
 }
 
-ILLEGAL_CUSTOMER_FIELD_NAME_CHARS = {'.', '/', '\n'}
+ILLEGAL_CUSTOMER_FIELD_NAME_CHARS = { '/', '\n'}
```
pandu-k: Do we need this to ensure that searches work? We really do still need to prevent customers creating fields with `.` in their names. Perhaps split this into two sets: `ILLEGAL_ADD_DOCS_CUSTOMER_FIELD_NAME_CHARS` and `ILLEGAL_SEARCH_CUSTOMER_FIELD_NAME_CHARS`

If we end up splitting them, add comments to explain why they were split

src/marqo/tensor_search/tensor_search.py
```diff
@@ -349,12 +349,12 @@ def _infer_opensearch_data_type(
         # OpenSearch requires that all content of an array be the same type.
         # This function doesn't validate.
         to_check = sample_field_content[0]
+    elif isinstance(sample_field_content, dict):
+        to_check = list(sample_field_content.values())[0]
```
pandu-k: Actually won't the OpenSearch type be None or undefined? As it is an object?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -368,7 +368,7 @@ def _get_chunks_for_field(field_name: str, doc_id: str, doc):
 def add_documents(config: Config, index_name: str, docs: List[dict], auto_refresh: bool,
                   non_tensor_fields=None, device=None, update_mode: str = "replace",
                   image_download_thread_count: int = 20, image_download_headers: dict = None,
-                  use_existing_tensors: bool = False):
+                  use_existing_tensors: bool = False, mappings: dict = None):
```
pandu-k, line 378: Nit: it would be good to add a comment about what mappings is in the docstring

src/marqo/tensor_search/tensor_search.py
```diff
@@ -627,11 +638,29 @@ def add_documents(config: Config, index_name: str, docs: List[dict], auto_refres
                     })
 
             # Add chunks_to_append along with doc metadata to total chunks
+
+            elif isinstance(field_content, dict):
+                if mappings[field]["type"]=="multimodal_combination":
```
pandu-k, line 656: 1. Catch key error 
    then add to unsuccessful docs ("object field that doesn't appear in mappings ... see here for documentation")
3. Type not given/type != multimodal_combination:  
    then add to unsuccessful docs("only multimodal_combination type objects are supported ... see here for documentation")

src/marqo/tensor_search/validation.py
```diff
@@ -91,7 +91,7 @@ def validate_list(field_content: typing.List, is_non_tensor_field: bool):
     return True
 
 
-def validate_field_content(field_content: typing.Any, is_non_tensor_field: bool) -> typing.Any:
+def validate_field_content(field: typing.Any, field_content: typing.Any, is_non_tensor_field: bool, mappings: dict) -> typing.Any:
```
pandu-k: Add to the docstring what field is

src/marqo/tensor_search/tensor_search.py
```diff
@@ -417,12 +417,17 @@ def add_documents(config: Config, index_name: str, docs: List[dict], auto_refres
 
     existing_fields = set(index_info.properties.keys())
     new_fields = set()
+    # A dict to store the multimodal_fields and their sub_fields
+    # dict = {multi_modal_field_1 : set(),
+    #      =  multi_modal_field_2 : set()}
+    new_multimodal_combo_fields = dict()
```
pandu-k: Question: How does the logic differ from `new_fields`?

pandu-k (changes requested):

src/marqo/tensor_search/backend.py
wanliAlex: solved

src/marqo/tensor_search/backend.py
wanliAlex: solved

src/marqo/tensor_search/backend.py
wanliAlex: done

src/marqo/tensor_search/enums.py
wanliAlex: no need to set type for dict at all, solved

src/marqo/tensor_search/tensor_search.py
wanliAlex: an object should not be passed into this method to check at all.

src/marqo/tensor_search/tensor_search.py
wanliAlex: added

src/marqo/tensor_search/validation.py
wanliAlex: done

src/marqo/tensor_search/tensor_search.py
wanliAlex: all the fields in the `new_fields` will be updated to `knn_fields_names` , added for `filtering`, and added to `text_properties`. however, for `multimodal_combo_fields`, `parent_field` will be added to `knn_fields_names`. but it  should not appear as a stand-alone field in `filtering` or `text_properties`.

src/marqo/tensor_search/tensor_search.py
wanliAlex: this check is finished in `validate_dict()`

src/marqo/tensor_search/tensor_search.py
wanliAlex: done

src/marqo/tensor_search/tensor_search.py
wanliAlex: done

src/marqo/tensor_search/backend.py
wanliAlex: we take the extra cost for every `add_document` call. will fix it later.

src/marqo/tensor_search/models/index_info.py
wanliAlex: done

src/marqo/tensor_search/validation.py
```diff
@@ -343,3 +346,81 @@ def validate_settings_object(settings_object):
...
+    if field not in mappings:
+        raise InvalidArgError(
+            f"The field `{field}` contains a dictionary field content `{field_content}`."
+            f"However, this field `{field}` is not in the add_document parameter mappings `{mappings}`, which is not supported."
+            f"Please change the type of your field content."
+            f"If you aim to use dictionary filed content as a special field,"
+            f"please check `https://docs.marqo.ai/0.0.15/` for more info.")
+
+    if mappings[field]["type"] == "multimodal_combination":
+        validate_multimodal_combination(field_content, is_non_tensor_field, mappings[field])
```
pandu-k, line 381: Raise an error if it's type != multimodal_combination (as no other types are supported yet)

src/marqo/tensor_search/validation.py
```diff
@@ -343,3 +346,81 @@ def validate_settings_object(settings_object):
...
+
+    Args:
+        field_content: the field content
+        is_non_tensor_field: whether this is a non-tensor-field
+        field_mapping: the mapping to help validate this field content
+
+    Returns:
+
+    '''
+    if len(field_content) < 2:
```
pandu-k: Can we change this to `<1` ?  It's forseeable that you might want to have a tensor combination with only 1 field (if you want to manipulate the resultant vector, such as making it negative).

src/marqo/tensor_search/tensor_search.py
```diff
@@ -353,7 +353,8 @@ def _infer_opensearch_data_type(
         to_check = sample_field_content
 
     if isinstance(to_check, dict):
-        raise errors.InvalidArgError("Field content can't be an object.")
+        raise errors.InvalidArgError("Field content can't be an object. An object should not be passed into _infer_opensearch_data_type"
+                                     "to check.")
```
pandu-k, line 363: nit: this is more of a runtime error than a customer error (`errors.InvalidArgError()`)

src/marqo/tensor_search/tensor_search.py
```diff
@@ -418,11 +420,18 @@ def add_documents(config: Config, index_name: str, docs: List[dict], auto_refres
     existing_fields = set(index_info.properties.keys())
     new_fields = set()
 
+    # A dict to store the multimodal_fields and their (child_fields, opensearch_type)
+    # dict = {parent_field_1 : set((child_field_1, type), ),
+    #      =  parent_field_2 : set((child_field_1, type), )}
+    # Check backend to see the differences between multimodal_fields and new_fields
+    new_multimodal_combo_fields = dict()
```
pandu-k: nit: this probably better called `new_object_fields`

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1251,17 +1295,20 @@ def _vector_text_search(
...
+                                           [s2_inference.vectorise(
+                                               model_name=index_info.model_name,
+                                               model_properties=_get_model_properties(index_info),
+                                               content=batch, device=selected_device,
+                                               normalize_embeddings=index_info.index_settings['index_defaults'][
+                                                   'normalize_embeddings'],
+                                               image_download_headers=image_download_headers
+                                           )
+                                               for batch in to_be_vectorised]
+                                           )
```
pandu-k: nit: I prefer the previous formatting

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1632,28 +1686,423 @@ def get_loaded_models() -> dict:
...
+                    image_field_name.append(sub_field_name)
+
+                except s2_inference_errors.S2InferenceError as e:
+                    combo_document_is_valid = False
+                    unsuccessful_doc_to_append = \
+                        (doc_index, {'_id': doc_id, 'error': e.message,
+                                     'status': int(errors.InvalidArgError.status_code),
+                                     'code': errors.InvalidArgError.code})
+
+                    return combo_chunk, combo_document_is_valid, unsuccessful_doc_to_append, combo_vectorise_time_to_add
```
pandu-k: We are not returning `new_field_from_multimodal_combination` - we should return at least an empty set

pandu-k: so we don't get a downstream error

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1632,28 +1686,423 @@ def get_loaded_models() -> dict:
...
+        doc_index: the index of the document. This is an interator variable `i` in the main body to iterator throught the docs
+        doc_id: the document id
+        selected_device: device from main body
+        index_info: index_info from main body
+    Returns:
+    combo_chunk: the combo_chunk to be appended to the main body
+    combo_document_is_valid:  if the document is a valid
+    unsuccessful_docs: appended unsucessful_docs
+    combo_total_vectorise_time: the vectorise time spent in combo field
+    '''
```
pandu-k, line 1750: Mention that we are also returning a set: new_field_from_multimodal_combination

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1632,28 +1686,423 @@ def get_loaded_models() -> dict:
...
+            link="https://marqo.pages.dev/latest/Models-Reference/dense_retrieval/"
+        )
+    except s2_inference_errors.S2InferenceError:
+        combo_document_is_valid = False
+        image_err = errors.InvalidArgError(message=f'Could not process given image: {multimodal_object_copy}')
+        unsuccessful_doc_to_append = \
+            (doc_index, {'_id': doc_id, 'error': image_err.message, 'status': int(image_err.status_code),
+                 'code': image_err.code})
+
+        return combo_chunk, combo_document_is_valid, unsuccessful_doc_to_append, combo_vectorise_time_to_add
```
pandu-k: Do we need to return an empty set, too, that represents `new_field_from_multimodal_combination`?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1632,28 +1686,423 @@ def get_loaded_models() -> dict:
...
+    combo_chunk = {}
+    unsuccessful_doc_to_append = tuple()
+    new_field_from_multimodal_combination = set()
+
+    # Copy the important mutable objects from main body for safety purpose
+    multimodal_object_copy = copy.deepcopy(multimodal_object)
+    doc_copy = copy.deepcopy(doc)
+
+    # 4 lists to store the field name and field content to vectorise.
+    text_field_name = []
```
pandu-k: nit: make it a plural, since it is a list (`text_field_name ` -> `text_field_names`)

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1632,28 +1686,423 @@ def get_loaded_models() -> dict:
...
+    text_content_to_vectorise = []
+
+    image_field_name = []
```
pandu-k: nit: make it a plural, since it is a list (`image_field_name ` -> `image_field_names`)

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1632,28 +1686,423 @@ def get_loaded_models() -> dict:
...
+                 'code': image_err.code})
+
+        return combo_chunk, combo_document_is_valid, unsuccessful_doc_to_append, combo_vectorise_time_to_add
+
+    sub_field_name_list = text_field_name + image_field_name
+    vectors_list = text_vectors + image_vectors
+
+    try:
+        if not len(sub_field_name_list) == len(vectors_list):
+            raise s2_inference_errors.BatchInferenceSizeNotMatchError
```
pandu-k: `BatchInferenceSizeNotMatchError` should be a 5xx internal error. This is means that there is a problem/bug with the Marqo instance

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1632,28 +1686,423 @@ def get_loaded_models() -> dict:
...
+    except s2_inference_errors.BatchInferenceSizeNotMatchError:
+        combo_document_is_valid = False
+        doc_err = errors.InvalidArgError(message=f"Batch inference size does not match content for multimodal field {field}")
+        unsuccessful_doc_to_append = \
+            (doc_index, {'_id': doc_id, 'error': doc_err.message,
+                         'status': int(doc_err.status_code), 'code': doc_err.code})
+        return combo_chunk, combo_document_is_valid, unsuccessful_doc_to_append, combo_vectorise_time_to_add
```
pandu-k: return an empty set for `new_field_from_multimodal_combination `

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1632,28 +1686,423 @@ def get_loaded_models() -> dict:
...
+
+    combo_chunk = dict({
+        utils.generate_vector_name(field): vector_chunk,
+        # TODO think of a good highlight
+        TensorField.field_content: json.dumps(multimodal_object),
+        TensorField.field_name: field,
+    })
+    return combo_chunk, combo_document_is_valid, unsuccessful_doc_to_append, combo_vectorise_time_to_add, new_field_from_multimodal_combination
+
+# def vectorise_multimodal_combination_field(field: str, multimodal_object: Dict[str, dict], doc: dict, doc_index: int,
```
pandu-k: delete these comments

src/marqo/tensor_search/backend.py
```diff
@@ -113,6 +120,16 @@ def add_customer_field_properties(config: Config, index_name: str,
                 "type": enums.OpenSearchDataType.keyword,
                 "ignore_above": 32766  # this is the Marqo-OS bytes limit
             }
+    if len(multimodal_combination_field) > 0:
```
pandu-k: nit: even if the len == 0 we will simply do no iterations of the following for loop

src/marqo/tensor_search/backend.py
```diff
@@ -59,7 +60,7 @@ def get_index_info(config: Config, index_name: str) -> IndexInfo:
 
 def add_customer_field_properties(config: Config, index_name: str,
                                   customer_field_names: Iterable[Tuple[str, enums.OpenSearchDataType]],
-                                  model_properties: dict):
+                                  model_properties: dict, multimodal_combination_field: Dict[str, Iterable[Tuple[str, enums.OpenSearchDataType]]]):
```
pandu-k: pluralise `multimodal_combination_field` -> `multimodal_combination_fields` because there may be many of these fields

src/marqo/tensor_search/validation.py
```diff
@@ -343,3 +346,81 @@ def validate_settings_object(settings_object):
...
+        raise InvalidArgError(
+            f"The multimodal_combination_field `{field_content}` is a dictonary will less than 2 fields. "
+            f"This is not a valid format of field content."
+            f"If you aim to use multimodal_combination, it must contain at least 2 fields. "
```
pandu-k: comment above, regarding 2 fields

src/marqo/tensor_search/validation.py
```diff
@@ -343,3 +346,81 @@ def validate_settings_object(settings_object):
...
+            raise InvalidArgError(
+                f"Multimodal-combination field content `{key}:{value}` \n  "
+                f"of type `{type(key).__name__} : {type(value).__name__}` is not of valid content type (one of {constants.ALLOWED_MULTIMODAL_FIELD_TYPES})."
+            )
+
+        if not key in field_mapping["weights"]:
+            raise InvalidArgError(
+                f"Multimodal-combination field content `{key}:{value}` \n  "
+                f"is not in the multimodal_field mappings weights `{field_mapping['weights']}`. Each sub_field requires a weights."
+                f"Please add the `{key}` to the mappings.")
```
pandu-k: A link to usage would be good here

src/marqo/tensor_search/validation.py
```diff
@@ -343,3 +346,81 @@ def validate_settings_object(settings_object):
...
+
+
+    if is_non_tensor_field:
+        raise InvalidArgError(
+            f"Field content `{field_content}` \n  " 
+            f"of type `{type(field_content).__name__}` is the content for a multimodal_combination."
+            f"It CAN NOT be a `non_tensor_field`. Remove this field from `non_tensor_field` or"
```
pandu-k: nit: `CAN NOT` -> `CANNOT`

src/marqo/tensor_search/backend.py
wanliAlex: done

src/marqo/tensor_search/backend.py
wanliAlex: we can delete the if check then

src/marqo/tensor_search/backend.py
wanliAlex: problem solved

src/marqo/tensor_search/validation.py
wanliAlex: we add a new validate mappings to solve the problem.

src/marqo/tensor_search/validation.py
wanliAlex: done

src/marqo/tensor_search/tensor_search.py
wanliAlex: there isn't a proper runtime error for this. should i just use MarqoError

src/marqo/tensor_search/tensor_search.py
wanliAlex: done

src/marqo/tensor_search/tensor_search.py
wanliAlex: Revert back.

src/marqo/tensor_search/tensor_search.py
wanliAlex: done

src/marqo/tensor_search/tensor_search.py
wanliAlex: done.

src/marqo/tensor_search/tensor_search.py
wanliAlex: done.

src/marqo/tensor_search/tensor_search.py
wanliAlex: added

src/marqo/tensor_search/tensor_search.py
wanliAlex: added

src/marqo/tensor_search/tensor_search.py
wanliAlex: should we break the add_doc or just skip the doc?

src/marqo/tensor_search/tensor_search.py
wanliAlex: done

src/marqo/tensor_search/tensor_search.py
wanliAlex: done

src/marqo/tensor_search/validation.py
wanliAlex: solved

src/marqo/tensor_search/validation.py
wanliAlex: solved

src/marqo/tensor_search/backend.py
```diff
@@ -135,6 +152,19 @@ def add_customer_field_properties(config: Config, index_name: str,
...
+
+
+    for multimodal_field, child_fields in multimodal_combination_fields.items():
+        # update the new multimodal_field if it's not in it
+        if multimodal_field not in new_index_properties:
+            new_index_properties[validation.validate_field_name(multimodal_field)] = \
+                {"properties": {validation.validate_field_name(child_field_name): {"type":child_type}
+                 for child_field_name, child_type in child_fields}}
+        # update the new child fields if the multimodal_field already in it
+        elif multimodal_field in new_index_properties:
```
pandu-k: nit: `else` should suffice

src/marqo/tensor_search/backend.py
```diff
@@ -135,6 +152,19 @@ def add_customer_field_properties(config: Config, index_name: str,
...
+            for child_field_name, child_type in child_fields:
```
pandu-k, line 165: We are overriding the existing child's types. Maybe this is alright.

src/marqo/tensor_search/validation.py
```diff
@@ -343,3 +346,129 @@ def validate_settings_object(settings_object):
...
+    Returns:
+
+    '''
+    if len(field_content) < 1:
+        raise InvalidArgError(
+            f"The multimodal_combination_field `{field_content}` is an empty dictionary. "
+            f"This is not a valid format of field content."
+            f"If you aim to use multimodal_combination, it must contain at least 1 field. "
+            f"please check `https://docs.marqo.ai/0.0.15/` for more info.")
+
```
pandu-k, line 403: nit: only one-line gaps between code within functions

src/marqo/tensor_search/tensor_search.py
pandu-k: Didn't you make the mismatch error a subclass of InvalidArgError? Then we can just use that

src/marqo/tensor_search/tensor_search.py
```diff
@@ -502,16 +516,16 @@ def add_documents(config: Config, index_name: str, docs: List[dict], auto_refres
                 )
                 break
 
-            if field not in existing_fields:
+            if (field not in existing_fields) and (not isinstance(field_content, dict)):
                 new_fields_from_doc.add((field, _infer_opensearch_data_type(copied[field])))
 
             # Don't process text/image fields when explicitly told not to.
             if field in non_tensor_fields:
+                # TODO field content can not be non-tensor field
```
pandu-k: nit: this comment might throw off future devs

pandu-k:
Unit tests running: https://github.com/marqo-ai/marqo/actions/runs/4287400891

src/marqo/tensor_search/constants.py
```diff
@@ -6,8 +8,14 @@
...
 ILLEGAL_CUSTOMER_FIELD_NAME_CHARS = {'.', '/', '\n'}
 
-ALLOWED_CUSTOMER_FIELD_TYPES = [str, int, float, bool, list]
+ALLOWED_CUSTOMER_FIELD_TYPES = [str, int, float, bool, list, dict]
 
 NON_TENSORISABLE_FIELD_TYPES = [int, float, bool, list]
+
+ALLOWED_MULTIMODAL_FIELD_TYPES = [str]
+
+ALLOWED_MAPPINGS_TYPES = ["multimodal_combination"]
```
pandu-k: Can we use `MARQO_OBJECT_TYPES` rather than  `ALLOWED_MAPPINGS_TYPES`?

src/marqo/tensor_search/validation.py
pandu-k: Also this conflicts with `validate_mappings_object()`

src/marqo/tensor_search/validation.py
```diff
@@ -14,6 +14,7 @@
 from enum import Enum
 import jsonschema
 from marqo.tensor_search.models.settings_object import settings_schema
+from marqo.tensor_search.models.mappings_object import mappings_schema, multimodal_combination_schema
```
pandu-k, line 18: Make sure there is only one `mappings_validation` and only one `multimodal_combination_mappings_validation`

tests/tensor_search/test_multimodal_tensor_combination.py
```diff
@@ -0,0 +1,882 @@
...
+        score_3 = get_score({
+            "combo_text_image": {
+                "text_field" : "A rider is riding a horse jumping over the barrier.",
+                "image_field" : "https://raw.githubusercontent.com/marqo-ai/marqo/mainline/examples/ImageSearchGuide/data/image1.jpg",
+        },
+        })
+
+        assert (score_3 >= min(score_1, score_2)) and (score_3 <= max(score_1, score_2))
+
+    def test_multimodal_tensor_combiantion_tensor_value(self):
```
pandu-k: nit: typo in "combination"

tests/tensor_search/test_multimodal_tensor_combination.py
```diff
@@ -0,0 +1,882 @@
...
+            "combo_text_image": {
+            "text_field" : "A rider is riding a horse jumping over the barrier.",
+            "image_field" : "https://raw.githubusercontent.com/marqo-ai/marqo/mainline/examples/ImageSearchGuide/data/image1.jpg",
+                },
+        })
+
+        self.assertEqual(score_1, score_3)
+
+    def test_multimodal_tensor_combination_vectorise_call(self):
+        """check if the chunks are properly created in the add_documents"""
```
pandu-k, line 213: Can we also query OpenSearch, to see if the documents are created as expected?

tests/tensor_search/test_multimodal_tensor_combination.py
```diff
@@ -0,0 +1,882 @@
...
+                    }
+
+            # ensure we only call multimodal-combination twice
+            assert len(mock_multimodal_combination.call_args_list) == 2
+
+            return True
+
+        assert run()
+
+    def test_multimodal_field_content_dictionary_validation(self):
```
pandu-k, line 286: You can also test if well formatted dicts _can_ be inserted as expected (only the bad dicts should not be found).
1. Check this from OpenSearch
2. Check if the add_documents call returns an appropriate failure message (`add_docs_res = tensors_search.add_documents(...`)

src/marqo/tensor_search/backend.py
wanliAlex: done

src/marqo/tensor_search/constants.py
wanliAlex: done

src/marqo/tensor_search/validation.py
wanliAlex: done

src/marqo/tensor_search/validation.py
wanliAlex: this is for add _docs

src/marqo/tensor_search/tensor_search.py
wanliAlex: deleted

tests/tensor_search/test_multimodal_tensor_combination.py
wanliAlex: updated

tests/tensor_search/test_multimodal_tensor_combination.py
wanliAlex: done

tests/tensor_search/test_multimodal_tensor_combination.py
wanliAlex: done

pandu-k (approved):

PR #337, opened 2023-02-20
change enums.Device

wanliAlex:
We change enums.Device from inherited class to class

enums.Device is a inherited class to class
```python
class Device(str, Enum):
    cpu = "cpu"
    cuda = "cuda"
```

it is now a class with a custom `__str__` method.
```python
class Device(str, Enum):
    cpu = "cpu"
    cuda = "cuda"

    def __str__(self):
        # To pass the str(Device.cpu) == "cpu" check in clip
        return self.value
```

no
no

no

no

* **Other information**:
We change this due to an odd device check in clip models, which is
`str(Device.cpu) == "cpu"`, this will return `False` with inherited class.
Therefore, we change it to class to solve the problem.


pandu-k:
Does [Device validation](https://github.com/marqo-ai/marqo/blob/mainline/src/marqo/tensor_search/api.py#L134) work with this change?

pandu-k (approved):

PR #363, opened 2023-03-01
Bulk search

Jeadie:

* **Other information**:


src/marqo/tensor_search/configs.py
```diff
@@ -34,7 +34,7 @@ def default_env_vars() -> dict:
         EnvVars.MARQO_MAX_INDEX_FIELDS: None,
         EnvVars.MARQO_MAX_DOC_BYTES: 100000,
         EnvVars.MARQO_MAX_RETRIEVABLE_DOCS: 10000,
-        EnvVars.MARQO_MODELS_TO_PRELOAD: ['hf/all_datasets_v4_MiniLM-L6', "ViT-L/14"],
+        EnvVars.MARQO_MODELS_TO_PRELOAD: ['hf/all_datasets_v4_MiniLM-L6'],
```
pandu-k: We need to test if this has any detrimental effect on the cloud. For some reason, when I tested it last time, it actually increased latency. Maybe this was for another unrelated reason.

src/marqo/tensor_search/models/api_models.py
```diff
@@ -32,6 +32,28 @@ def validate_search_method(cls, value):
...
+class VectorisedJobs(BaseModel):
+    model_name: str
+    model_properties: Dict[str, Any]
+    content: List[Union[str, List[str]]]
+    device: str
+    normalize_embeddings: bool
+    image_download_headers: Optional[Dict]
+
+    def __hash__(self):
+        return hash(self.model_name) + hash(json.dumps(self.model_properties, sort_keys=True)) + hash(json.dumps(self.content, sort_keys=True)) + hash(self.device) + hash(self.normalize_embeddings) + hash(json.dumps(self.image_download_headers, sort_keys=True))
```
pandu-k: Put the attributes into a tuple, and then make a hash of the tuple? This is what I've seen

src/marqo/tensor_search/tensor_search.py
```diff
@@ -985,6 +989,108 @@ def refresh_index(config: Config, index_name: str):
     return HttpRequests(config).post(path=F"{index_name}/_refresh")
 
 
+@add_timing
+def bulk_search(query: BulkSearchQuery, device: str, marqo_config: config.Config, reranker: Union[str, Dict] = None, simplified_format=True, number_of_highlights: int = 3, return_doc_ids=False, verbose: bool = True):
```
pandu-k: What's the reasoning for having some parameters request level and others at the query level?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -985,6 +989,108 @@ def refresh_index(config: Config, index_name: str):
...
+        raise err
+
+    fill_index_cache(marqo_config, [q.index for q in query.queries])
+
+    ## TODO: Add lexical queries to bulk msearch
+    tensor_queries = list(filter(lambda q: q.searchMethod == SearchMethod.TENSOR, query.queries))
+    lexical_queries = list(filter(lambda q: q.searchMethod == SearchMethod.LEXICAL, query.queries))
+
+    if len(lexical_queries) + len(tensor_queries) == 0:
+        raise errors.InvalidArgError(f"Bulk search called with unknown search method(s)")
```
pandu-k: Is it possible that this occurs if bulk search is called with 0 queries? or is this filtered out earlier?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -985,6 +989,108 @@ def refresh_index(config: Config, index_name: str):
...
+        selected_device = marqo_config.indexing_device if device is None else device
+        search_results = _bulk_vector_text_search(marqo_config, query.queries, selected_device, simplified_format=simplified_format, number_of_highlights=number_of_highlights, return_doc_ids=return_doc_ids)
+
+    if len(lexical_queries) > 0:
+        # TODO: parallelise
+        search_results = [_lexical_search(
+            config=marqo_config, index_name=q.index, text=q.q, result_count=q.limit, offset=q.offset,
+            return_doc_ids=return_doc_ids, searchable_attributes=q.searchableAttributes, verbose=verbose,
+            filter_string=q.filter, attributes_to_retrieve=q.attributesToRetrieve
+        ) for q in lexical_queries]
```
pandu-k: We can't mix search methods - otherwise Lexical overwrites tensor

src/marqo/tensor_search/tensor_search.py
```diff
@@ -985,6 +989,108 @@ def refresh_index(config: Config, index_name: str):
...
+        s["limit"] = q.limit
+        s["offset"] = q.offset
+
+    if reranker is not None:
+        logger.info("reranking using {}".format(reranker))
+        for i, s in enumerate(search_results):
+            rerank(query.queries[i], s, reranker, device, 1 if simplified_format else number_of_highlights)
+            if not query.queries[0].highlights:
+                for hit in s["hits"]:
+                    del hit["_highlights"]
```
pandu-k: What is the reasoning for deleting these highlights?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -985,6 +989,108 @@ def refresh_index(config: Config, index_name: str):
...
+    for i, s in enumerate(search_results):
+        q = query.queries[i]
+        s["query"] = q.q
+        s["limit"] = q.limit
+        s["offset"] = q.offset
+
+    if reranker is not None:
+        logger.info("reranking using {}".format(reranker))
+        for i, s in enumerate(search_results):
+            rerank(query.queries[i], s, reranker, device, 1 if simplified_format else number_of_highlights)
```
pandu-k: simplified is always True I am pretty sure (simplified is False is unmaintained)

src/marqo/tensor_search/tensor_search.py
```diff
@@ -985,6 +989,108 @@ def refresh_index(config: Config, index_name: str):
...
+    if q.limit <= 0:
+        return errors.IllegalRequestedDocCount("search result limit must be greater than 0!")
+    if q.offset < 0:
+        return errors.IllegalRequestedDocCount("search result offset cannot be less than 0!")
+
+    # validate query
+    validation.validate_query(q=q.q, search_method=q.searchMethod)
+
+    # Validate result_count + offset <= int(max_docs_limit)
+    max_docs_limit = utils.read_env_vars_and_defaults(EnvVars.MARQO_MAX_RETRIEVABLE_DOCS)
```
pandu-k: Do we want to make this at the request level (open question)?

src/marqo/tensor_search/utils.py
```diff
@@ -253,4 +255,15 @@ def parse_lexical_query(text: str) -> Tuple[List[str], str]:
     required_terms = [term.replace('\\"', '"') for term in required_terms]
     optional_blob = optional_blob.replace('\\"', '"')
 
-    return (required_terms, optional_blob)
\ No newline at end of file
+    return (required_terms, optional_blob)
+
+
+def add_timing(f, key: str = "processingTimeMs"):
```
pandu-k, line 299: Can you add a docstring to this? This is specifically for endpoint functions function and it adds the `processingTimeMs` to the created dict

src/marqo/tensor_search/tensor_search.py
```diff
@@ -985,6 +989,114 @@ def refresh_index(config: Config, index_name: str):
...
+
+        REFRESH_INTERVAL_SECONDS = 2
+        # update cache in the background
+        cache_update_thread = threading.Thread(
+            target=index_meta_cache.refresh_index_info_on_interval,
+            args=(config, idx, REFRESH_INTERVAL_SECONDS))
+        cache_update_thread.start()
+
+
+def validate_query_input(q: BulkSearchQueryEntity) -> Optional[errors.MarqoError]:
```
pandu-k: move to validation.py

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1347,248 @@ def _lexical_search(
     return {'hits': res_list}
 
 
+def construct_vector_input_batches(query: Union[str, Dict], index_info) -> Union[List[str], List[List[str]]]:
```
pandu-k: docstring: this is search query's `q` parameter

pandu-k: docstring: "this does xyz..."

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1347,248 @@ def _lexical_search(
...
+            return [batch for batch in [text_queries, image_queries] if batch]
+        else:
+            return [[k for k, _ in ordered_queries], ]
+
+
+def get_vector_properties_to_search(searchable_attributes, index_info,  raise_on_searchable_attribs: bool):
+    if searchable_attributes is None:
+        return index_info.get_vector_properties().keys()
+    else:
+        if raise_on_searchable_attribs:
```
pandu-k: This code path is dead

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1347,248 @@ def _lexical_search(
...
+        )
+        else:
+            searchable_attributes_as_vectors = {utils.generate_vector_name(field_name=attribute)
+                                                for attribute in searchable_attributes}
+            # discard searchable attributes that aren't found in the cache:
+            return searchable_attributes_as_vectors.intersection(
+        index_info.get_vector_properties().keys())
+
+
+def construct_msearch_body_elements(vector_properties_to_search, offset, filter_string, index_info, result_count, vectorised_text, attributes_to_retrieve, index_name, contextualised_filter):
```
pandu-k: add types to params

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1347,248 @@ def _lexical_search(
...
+
+def bulk_msearch(config: Config, body: List[Dict]):
+    start_search_http_time = timer()
+    try:
+        response = HttpRequests(config).get(path=F"_msearch", body=utils.dicts_to_jsonl(body))
+        end_search_http_time = timer()
+        total_search_http_time = end_search_http_time - start_search_http_time
+        total_os_process_time = response["took"] * 0.001
+        num_responses = len(response["responses"])
+        logger.info(f"search (tensor) roundtrip: took {total_search_http_time:.3f}s to send {num_responses} search queries (roundtrip) to Marqo-os.")
```
pandu-k: logger.debug, now

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1347,248 @@ def _lexical_search(
...
+                raise errors.IllegalRequestedDocCount(
+            "Marqo-OS rejected the response due to too many requested results. "
+            "Try reducing the query's limit parameter") from e
+            elif 'parse_exception' in response["responses"][0]["error"]["root_cause"][0]["reason"]:
+                raise errors.InvalidArgError("Syntax error, could not parse filter string") from e
+            raise errors.BackendCommunicationError(f"Error communicating with Marqo-OS backend:\n{response}")
+        except (KeyError, IndexError) as e2:
+            raise e
+
+    logger.info(f"  search (tensor) Marqo-os processing time: took {total_os_process_time:.3f}s for Marqo-os to execute the search.")
```
pandu-k: logger.debug now

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1347,248 @@ def _lexical_search(
...
+def _bulk_vector_text_search(config: Config, queries: List[BulkSearchQueryEntity], selected_device, raise_on_searchable_attribs=False, simplified_format=True, number_of_highlights: int = 3, return_doc_ids=False,  result_count: int = 5):
+    if len(queries) == 0:
+        return []
+
+    start_preprocessing_time = timer()
+    jobs: List[VectorisedJobs] = []
+    job_to_query_idx: Dict[VectorisedJobs, int] = {} # Map job to index of query it came from
+
+    ## 1. Pre-process inputs ready for s2_inference.vectorise
+    for i in range(len(queries)):
```
pandu-k: you can use `enumerate()` to do the same thing

src/marqo/tensor_search/models/api_models.py
```diff
@@ -32,6 +32,56 @@ def validate_search_method(cls, value):
...
+    content: List[Union[str, List[str]]]
+    device: str
+    normalize_embeddings: bool
+    image_download_headers: Optional[Dict]
+
+    def __hash__(self):
+        return self.groupby_key() + hash(json.dumps(self.content, sort_keys=True))
+
+    def groupby_key(self) -> JHash:
+        return VectorisedJobs.get_groupby_key(self.model_name, self.model_properties, self.device, self.normalize_embeddings, self.image_download_headers)
```
pandu-k: we also need track/group by whether the content is_image

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1346,257 @@ def _lexical_search(
...
+                normalize_embeddings=v.normalize_embeddings,
+                image_download_headers=v.image_download_headers
+            ) for batch in v.content] # Todo: check if should be batching.
+        )
+
+    ## 3. Get vectors back from batch, process `ordered_queries` into each query.
+    qidx_to_vectors: Dict[Qidx, List[List[float]]] = defaultdict(list)
+    for qidx, ptr in qidx_to_job.items():
+        vectors = job_ptr_to_vectors[ptr.job_hash][ptr.start_idx: ptr.end_idx]
+        qidx_to_vectors[qidx].append(vectors)
```
pandu-k: so this is re-nesting the flattened content?
[i1, i2, i3, t1, t2, t3] -> [[i1, i2, i3], [t1, t2, t3]]

i = image, t = text

a comment explaining this might be nice

pandu-k: we have to ensure that we get the original mapping back, so that each query gets the correct weighting

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1346,257 @@ def _lexical_search(
...
+
+def assign_query_to_vector_job(q, jobs, content, index_info, device) -> VectorisedJobPointer:
+    vector_job = VectorisedJobs(
+        model_name=index_info.model_name,
+        model_properties=_get_model_properties(index_info),
+        content=content,
+        device=device,
+        normalize_embeddings=index_info.index_settings['index_defaults']['normalize_embeddings'],
+        image_download_headers=q.image_download_headers
+    )
```
pandu-k: docstring: what does this do?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1346,257 @@ def _lexical_search(
...
+    qidx_to_vectors: Dict[Qidx, List[List[float]]] = defaultdict(list)
+    for qidx, ptr in qidx_to_job.items():
+        vectors = job_ptr_to_vectors[ptr.job_hash][ptr.start_idx: ptr.end_idx]
+        qidx_to_vectors[qidx].append(vectors)
+
+        q = queries[qidx]
+        ordered_queries = list(q.q.items()) if isinstance(q.q, dict) else None
+        if ordered_queries:
+            # multiple queries. We have to weight and combine them:
+            index_info = get_index_info(config=config, index_name=q.index)
```
pandu-k: this isn't a blocking call to _mappings, right?

src/marqo/tensor_search/tensor_search.py
pandu-k: Add return type

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1346,257 @@ def _lexical_search(
...
+        else:
+            contextualised_filter = ''
+
+        vector_properties_to_search = get_vector_properties_to_search(qq.searchableAttributes, index_info, raise_on_searchable_attribs)
+        vectorised_text = qidx_to_vectors[qidx]
+        body = construct_msearch_body_elements(vector_properties_to_search, qq.offset, q.filter, index_info, qq.limit, vectorised_text, qq.attributesToRetrieve, qq.index, contextualised_filter)
+        query_to_body_parts[qidx] = body
+        query_to_body_count[qidx] = len(body)
+
+    aggregate_body = [line for body in query_to_body_parts.values() for line in body]
```
pandu-k: what does this do?

pandu-k: is this a reduce? If so I think it is more readable as an explicit `reduce()` call

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1346,257 @@ def _lexical_search(
...
+    ## 5. POST aggregate  to /_msearch
+    responses = bulk_msearch(config, aggregate_body)
+    start_postprocess_time = timer()
+
+    # 6. Get documents back to each query, perform "gather" operation
+    results = []
+    for qidx, count in query_to_body_count.items():
+        num_of_docs = count // 2
+        result = responses[:num_of_docs]
+        responses = responses[num_of_docs:] # remove docs from response for next query
```
pandu-k: mutating the responses in this loop makes me a bit uncomfortable

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1346,257 @@ def _lexical_search(
...
+            elif 'parse_exception' in response["responses"][0]["error"]["root_cause"][0]["reason"]:
+                raise errors.InvalidArgError("Syntax error, could not parse filter string") from e
+            raise errors.BackendCommunicationError(f"Error communicating with Marqo-OS backend:\n{response}")
+        except (KeyError, IndexError) as e2:
+            raise e
+
+    logger.info(f"  search (tensor) Marqo-os processing time: took {total_os_process_time:.3f}s for Marqo-os to execute the search.")
+    return responses
+
+def gather_documents_from_response(resp):
```
pandu-k: docstring: this is used to aggregate documents, for cases where the same document is retrieved for multiple field-queries

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1346,257 @@ def _lexical_search(
...
+        num_of_docs = count // 2
+        result = responses[:num_of_docs]
+        responses = responses[num_of_docs:] # remove docs from response for next query
+        query = queries[qidx]
+        gathered_docs = gather_documents_from_response(result)
+        if query.boost is not None:
+            gathered_docs = boost_score(gathered_docs, query.boost)
+        docs_chunks_sorted = sort_chunks(gathered_docs)
+
+        if simplified_format:
```
pandu-k: it's always simplified

tests/tensor_search/test_bulk_search.py
```diff
@@ -0,0 +1,1015 @@
+import math
```
pandu-k, line 2: test suggestions
- multiple searchable attributes
- wieghted queries
- boosting fields

pandu-k:
Maybe split each step into its own function? It would help constructing tests + reasoning about it

src/marqo/tensor_search/configs.py
Jeadie: This is my bad. I removed it solely to make local testing faster.

src/marqo/tensor_search/tensor_search.py
Jeadie: I kept anything that wasn't in the body, as a request level param. I didn't have a strong justification. I think we could change this depending on what we think customers will need/want.

src/marqo/tensor_search/tensor_search.py
Jeadie: This has been addressed.

src/marqo/tensor_search/tensor_search.py
Jeadie: This was previous behaviour? https://github.com/marqo-ai/marqo/blob/mainline/src/marqo/tensor_search/tensor_search.py#L1104

src/marqo/tensor_search/tensor_search.py
Jeadie: Will update. Will remove `simplified_format=True` from `def search(...`

Jeadie: This also makes `number_of_highlights: int`  redundant.

src/marqo/tensor_search/tensor_search.py
Jeadie: Potentially. At the very least we may want to cap the sum of doc requests in a bulk search.

src/marqo/tensor_search/tensor_search.py
Jeadie: This part, `if searchable_attributes is None:` ? Isn't `searchable_attributes: Union[None, List[str]]`

src/marqo/tensor_search/tensor_search.py
Jeadie: So we run `fill_index_cache` at the start of `bulk_search` for all the indices in the search requests. I think this should be sufficient to ensure `get_index_info` should be in cache.

src/marqo/tensor_search/tensor_search.py
Jeadie: This is a flattening of first order. e.g.: 
```
[[{}, {}, {}, {}], [{}, {}, {}], [{}, {}, {}, {}], [{}, {}]]
```
becomes
```
[{}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}]
```

src/marqo/tensor_search/tensor_search.py
Jeadie: It made me too, I made it a deep copy now.

tests/tensor_search/test_bulk_search.py
Jeadie: Yes, Testing is not done.

src/marqo/tensor_search/models/search.py
```diff
@@ -0,0 +1,42 @@
+import json
+from pydantic import BaseModel
+from typing import Any, Union, List, Dict, Optional, NewType
+
+Qidx = NewType('Qidx', int)
+JHash = NewType('JHash', int)
```
Pandu-Kerr: Maybe add comments/ expand these names so that they make sense to a dev with less context?

`QueryIndex = ...  # Indicates the position of a search query in a bulk search request`
`VectoriseJobHash = ....  # hash of a VectoriseJob`

src/marqo/tensor_search/utils.py
```diff
@@ -253,4 +257,19 @@ def parse_lexical_query(text: str) -> Tuple[List[str], str]:
...
+    """
```
Pandu-Kerr: nit: use Google docstring convention

src/marqo/tensor_search/models/search.py
```diff
@@ -0,0 +1,42 @@
...
+
+class VectorisedJobPointer(BaseModel):
+    job_hash: JHash
+    start_idx: int
+    end_idx: int
+
+class VectorisedJobs(BaseModel):
```
pandu-k, line 15: These are logically grouped together because they have the same (download, vectorise) args - maybe add a comment explaining this

src/marqo/tensor_search/models/search.py
```diff
@@ -0,0 +1,42 @@
...
+from pydantic import BaseModel
+from typing import Any, Union, List, Dict, Optional, NewType
+
+Qidx = NewType('Qidx', int)
+JHash = NewType('JHash', int)
+
+class VectorisedJobPointer(BaseModel):
+    job_hash: JHash
+    start_idx: int
+    end_idx: int
```
pandu-k, line 13: A docstring explaining this class' relationship with search queries:
`where job_hash and start/end index refers to the location of a search query's content`

src/marqo/tensor_search/validation.py
```diff
@@ -51,6 +52,33 @@ def validate_query(q: Union[dict, str], search_method: Union[str, SearchMethod])
...
+        raise IllegalRequestedDocCount(
+    f"The search result limit + offset must be less than or equal to the MARQO_MAX_RETRIEVABLE_DOCS limit of"
+    f"[{max_docs_limit}]. Marqo received search result limit of `{q.limit}` and offset of `{q.offset}`."
+    )
+
+    validate_boost(boost=q.boost, search_method=q.searchMethod)
+    if q.searchableAttributes is not None:
+        [validate_field_name(attribute) for attribute in q.searchableAttributes]
+    if q.attributesToRetrieve is not None:
+        if not isinstance(q.attributesToRetrieve, (List, typing.Tuple)):
```
pandu-k, line 79: nit: this type of sequence check may also be good for q.searchableAttributes

src/marqo/tensor_search/tensor_search.py
```diff
@@ -985,6 +990,84 @@ def refresh_index(config: Config, index_name: str):
     return HttpRequests(config).post(path=F"{index_name}/_refresh")
 
 
+@add_timing
+def bulk_search(query: BulkSearchQuery, marqo_config: config.Config, reranker: Union[str, Dict] = None, return_doc_ids: bool=False, verbose: bool = True, device=None, raise_on_searchable_attribs: bool=False):
+    errs = [validation.validate_bulk_query_input(q) for q in query.queries]
+    if any(errs):
```
pandu-k, line 1008: The pattern we use for add_documents is to return an error message in place of the problematic document, allowing the non-errored docs to propagate through

src/marqo/tensor_search/tensor_search.py
```diff
@@ -985,6 +990,84 @@ def refresh_index(config: Config, index_name: str):
...
+        start_rerank_time = timer()
+        rerank.rerank_search_results(search_result=result, query=query.q,
+                                     model_name=reranker, device=device,
+                                     searchable_attributes=query.searchableAttributes, num_highlights=num_highlights)
+        logger.debug(f"search ({query.searchMethod.lower()}) reranking using {reranker}: took {(timer() - start_rerank_time):.3f}s to rerank results.")
+    except Exception as e:
+        raise errors.BadRequestError(f"reranking failure due to {str(e)}")
+
+
+def fill_index_cache(config: Config, index_names: List[str]):
```
pandu-k: docstring here

pandu-k: Also, better described as `def update_indexes_in_background()`

src/marqo/tensor_search/tensor_search.py
pandu-k: As in the `raise_on_searchable_attribs` isn't used any more

src/marqo/tensor_search/tensor_search.py
```diff
@@ -985,6 +990,84 @@ def refresh_index(config: Config, index_name: str):
...
+def bulk_search(query: BulkSearchQuery, marqo_config: config.Config, reranker: Union[str, Dict] = None, return_doc_ids: bool=False, verbose: bool = True, device=None, raise_on_searchable_attribs: bool=False):
+    errs = [validation.validate_bulk_query_input(q) for q in query.queries]
+    if any(errs):
+        err = next(e for e in errs if e is not None)
+        raise err
+
+    if len(query.queries) == 0:
+        return {"result": []}
+
+    fill_index_cache(marqo_config, [q.index for q in query.queries])
```
pandu-k: we may need to validate that `.index` is in each "q" so we don't run into attrib. errors

src/marqo/tensor_search/tensor_search.py
```diff
@@ -985,6 +990,84 @@ def refresh_index(config: Config, index_name: str):
...
+
+    tensor_queries = dict(filter(lambda e: e[1].searchMethod == SearchMethod.TENSOR, enumerate(query.queries)))
```
pandu-k: add type signature

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1318,311 @@ def _lexical_search(
...
+        return []
+
+    start_preprocessing_time = timer()
+    selected_device = config.indexing_device if device is None else device
+
+    # 1. Pre-process inputs ready for s2_inference.vectorise
+    qidx_to_job, jobs = create_vector_jobs(queries, config, selected_device)
+
+    # 2. Vectorise in batches against all queries
+    job_ptr_to_vectors: Dict[JHash, List[float]] = vectorise_jobs(list(jobs.values()))
```
pandu-k: To ensure that we are vectorising in batches, we can mock vectorise (), and see if the number of calls is as expected
(if batch_size = 16, and number of docs = 32, and all args are the same, then  number of calls = 2)

src/marqo/tensor_search/tensor_search.py
```diff
@@ -985,6 +990,84 @@ def refresh_index(config: Config, index_name: str):
...
+        logger.info("reranking using {}".format(reranker))
+        for i, s in enumerate(search_results):
+            rerank(query.queries[i], s, reranker, device, 1)
+
+    return {
+        "result": search_results
+    }
+
+
+def rerank(query: BulkSearchQueryEntity, result: Dict[str, Any], reranker: Union[str, Dict], device: str, num_highlights: int):
```
pandu-k: reranking is defined here at the request level. Is there anything stopping users defining it at query level? 

Is there room in the API for it to be defined at that level?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -985,6 +990,84 @@ def refresh_index(config: Config, index_name: str):
     return HttpRequests(config).post(path=F"{index_name}/_refresh")
 
 
+@add_timing
+def bulk_search(query: BulkSearchQuery, marqo_config: config.Config, reranker: Union[str, Dict] = None, return_doc_ids: bool=False, verbose: bool = True, device=None, raise_on_searchable_attribs: bool=False):
```
pandu-k: This is a really nice entrypoint style pipeline function

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1318,311 @@ def _lexical_search(
...
+            searchable_attributes_as_vectors = {
+                utils.generate_vector_name(field_name=attribute) for attribute in searchable_attributes
+            }
+            # discard searchable attributes that aren't found in the cache:
+            return list(searchable_attributes_as_vectors.intersection(
+                index_info.get_vector_properties().keys()
+            ))
+
+
+def construct_msearch_body_elements(vector_properties_to_search: List[str], offset: int, filter_string: str, index_info: IndexInfo, result_count: int, vectorised_text: List[float], attributes_to_retrieve: List[str], index_name: str, contextualised_filter: str) -> List[Dict[str, Any]]:
```
pandu-k: docstring: `...constructs an msearch body for a single search query...`

pandu-k: `vectorised_text` is better as `query_vector`

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1318,311 @@ def _lexical_search(
...
+            search_query["_source"] = {"include": attributes_to_retrieve} if len(attributes_to_retrieve) > 0 else False
+
+        if filter_string is not None:
+            search_query["query"]["nested"]["query"]["knn"][f"{TensorField.chunks}.{vector_field}"][
+                "filter"] = {
+                    "query_string": {"query": f"{contextualised_filter}"}
+                }
+        body += [{"index": index_name}, search_query]
+    return body
+
```
pandu-k, line 1422: nit: 2 lines between functions (1 line between class methods)

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1318,311 @@ def _lexical_search(
...
+def bulk_msearch(config: Config, body: List[Dict]) -> List[Dict]:
```
pandu-k, line 1424: docstring: sends N search queries to Marqo-OS and receives one dict per field searched for each query...

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1318,311 @@ def _lexical_search(
...
+        where the same document is retrieved for multiple field-queries.
+    """
+    gathered_docs = dict()
+
+    for i, query_res in enumerate(resp):
+        for doc in query_res:
+            doc_chunks = doc["inner_hits"][TensorField.chunks]["hits"]["hits"]
+            if doc["_id"] in gathered_docs:
+                gathered_docs[doc["_id"]]["doc"] = doc
+                gathered_docs[doc["_id"]]["chunks"].extend(doc_chunks)
```
pandu-k, line 1465: This might actually be better as a set, at least as an intermediate structure

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1318,311 @@ def _lexical_search(
...
+            else:
+                gathered_docs[doc["_id"]] = {
+                    "_id": doc["_id"],
+                    "doc": doc,
+                    "chunks": doc_chunks
+                }
+
+        # Filter out docs with no inner hits:
+        for doc_id in list(gathered_docs.keys()):
```
pandu-k, line 1474: nit, we may want to log warnings if this happens (it shouldn't really happen)

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1318,311 @@ def _lexical_search(
...
+    return gathered_docs
+
+def assign_query_to_vector_job(q: BulkSearchQueryEntity, jobs: Dict[JHash, VectorisedJobs], content: List[Union[str, List[str]]], index_info: IndexInfo, device: str) -> VectorisedJobPointer:
+    """
+        For a individual query, assign its content (to be vectorised) to a vector job. If none exist with the correct
+        specifications, create a new job.
+
+        Mutates entries in, and adds values to `jobs` param.
+
+        Returns: A pointer to the location in a vector job that will have it's vectorised content.
```
pandu-k: typo, should be `... its vectorised content.`

pandu-k: Good docstring though

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1318,311 @@ def _lexical_search(
...
+                content=batch, device=v.device,
+                normalize_embeddings=v.normalize_embeddings,
+                image_download_headers=v.image_download_headers
+            ) for batch in v.content] # Todo: check if should be batching.
+        )
+    return result
+
+def get_query_vectors_from_jobs(queries: List[BulkSearchQueryEntity], qidx_to_job: Dict[Qidx, VectorisedJobs], job_to_vectors: Dict[JHash, VectorisedJobs], config: Config) -> Dict[Qidx, List[List[float]]]:
+    """
+    Retrieve the vectorised content associated to each query from the set of batch vectorise jobs.
```
pandu-k, line 1581: Describe the role of each param

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1235,6 +1318,311 @@ def _lexical_search(
...
+        vectors = job_to_vectors[ptr.job_hash][ptr.start_idx: ptr.end_idx]
+        # qidx_to_vectors[qidx].append(vectors)
+
+        q = queries[qidx]
+        ordered_queries = list(q.q.items()) if isinstance(q.q, dict) else None
+        if ordered_queries:
+            # multiple queries. We have to weight and combine them:
+            index_info = get_index_info(config=config, index_name=q.index)
+            weighted_vectors = [np.asarray(vec) * weight for vec, weight in zip(vectors, [w for _, w in ordered_queries])]
+            vectorised_text = np.mean(weighted_vectors, axis=0)
```
pandu-k: Let's rename `vectorised_text` to something more useful, `merged_vector`

src/marqo/tensor_search/tensor_search.py
pandu-k: nit: to me this is more readable `functools.reduce(lambda x, y: x + y, query_to_body_parts.values())`

pandu-k: Maybe add a comment above it saying this flattens it

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1376,51 +1763,51 @@ def _vector_text_search(
                             "include": ["__chunks.__field_content", "__chunks.__field_name"]
                         }
                     },
```
pandu-k: I think these changes just cook the indentation of the search, right? Can we keep the current indentation pattern, unless there's a beneficial reason to change it?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1376,51 +1763,51 @@ def _vector_text_search(
...
         }
 
         field_names = list(index_info.get_text_properties().keys())
         if attributes_to_retrieve is not None:
             search_query["_source"] = {"include": attributes_to_retrieve} if len(attributes_to_retrieve) > 0 else False
 
-        if filter_string is not None:
-            search_query["query"]["nested"]["query"]["knn"][f"{TensorField.chunks}.{vector_field}"][
+            if filter_string is not None:
+                search_query["query"]["nested"]["query"]["knn"][f"{TensorField.chunks}.{vector_field}"][
```
pandu-k: Why are these out-dented?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1376,51 +1763,51 @@ def _vector_text_search(
...
+                        readable_body[i]["query"]["nested"]["query"]["knn"][vec]["vector"] = \
+                            readable_body[i]["query"]["nested"]["query"]["knn"][vec]["vector"][:5]
+                    pprint.pprint(readable_body)
+            if verbose == 2:
+                pprint.pprint(body, compact=True)
+
+        if not body:
+            # empty body means that there are no vector fields associated with the index.
+            # This probably means the index is emtpy
+            return {"hits": []}
```
pandu-k: why are these out-dented?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1597,6 +1984,80 @@ def format_ordered_docs_simple(ordered_docs_w_chunks: List[dict]) -> dict:
...
+            cleaned = _clean_doc(dict(), doc_id=d['_id'])
+
+        cleaned["_highlights"] = {
+            d["chunks"][0]["_source"][TensorField.field_name]: d["chunks"][0]["_source"][
+                TensorField.field_content]
+        }
+        cleaned["_score"] = d["chunks"][0]["_score"]
+        simple_results.append(cleaned)
+    return {"hits": simple_results[:result_count]}
+
```
pandu-k, line 2151: nit: 2 lines between functions

src/marqo/tensor_search/models/search.py
Jeadie: I think expanding out the type everywhere would be verbose. I definitely think it'll need comments.

src/marqo/tensor_search/validation.py
Jeadie: We can. We don't do this yet in `def search`.

Jeadie: Also, since I added `BulkSearchQueryEntity` (and `SearchQuery`), pydantic naturally handles these type checking a construction.

src/marqo/tensor_search/tensor_search.py
Jeadie: Index cannot be None, because Pydantic object doesn't have it as Optional
```python
class BulkSearchQueryEntity(SearchQuery):
    index: str
``` 
(if i understand pydantic correctly)

src/marqo/tensor_search/tensor_search.py
Jeadie: Didn't want to touch the core logic here (at least in this one, already large, PR).

src/marqo/tensor_search/tensor_search.py
Jeadie: Again, as above.

src/marqo/tensor_search/tensor_search.py
Jeadie: Aren't they the same indentation?

src/marqo/tensor_search/tensor_search.py
Jeadie: Good find.

src/marqo/tensor_search/models/api_models.py
```diff
@@ -20,7 +19,7 @@ class SearchQuery(BaseModel):
     showHighlights: bool = True
     reRanker: str = None
     filter: str = None
-    attributesToRetrieve: List[str] = None
+    attributesToRetrieve: Union[None, List[str]] = None
```
pandu-k, line 22: how does this effect cURL users not including `attributesToRetrieve` in their request?

src/marqo/version.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "0.0.15"
+__version__ = "0.0.16-dev-0"
```
pandu-k: comment: this should be 0.0.16 before merging

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1032,17 +1033,21 @@ def bulk_search(query: BulkSearchQuery, marqo_config: config.Config, reranker: U
...
-        logger.info("reranking using {}".format(reranker))
-        for i, s in enumerate(search_results):
-            rerank(query.queries[i], s, reranker, device, 1)
+        ## TODO: filter out highlights within `_lexical_search` 
+        if not q.showHighlights:
+            for hit in s["hits"]:
+                del hit["_highlights"]
+
+        if q.reRanker is not None:
+            logger.info(f"reranking {i}th query using {q.reRanker}")
```
pandu-k: can we move this to the debug level?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1327,30 +1334,26 @@ def construct_vector_input_batches(query: Union[str, Dict], index_info) -> Union
             text_queries = [k for k, _ in ordered_queries if _is_image(k)]
             image_queries = [k for k, _ in ordered_queries if not _is_image(k)]
             return [batch for batch in [text_queries, image_queries] if batch]
+            return text_queries + image_queries
         else:
-            return [[k for k, _ in ordered_queries], ]
+            return [k for k, _ in ordered_queries]
```
pandu-k: comment: this will likely be overwritten by vector bulk search, but:
- Having a variable return type may be confusing 
- there are 2 return statements (lines 1336/1337) for some reason

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1508,21 +1512,32 @@ def vectorise_jobs(jobs: List[VectorisedJobs]) -> Dict[JHash, List[float]]:
...
-        result[v.groupby_key()] = functools.reduce(
-            lambda x, y: x + y,
-            [s2_inference.vectorise(
-                model_name=v.model_name, model_properties=v.model_properties,
-                content=batch, device=v.device,
-                normalize_embeddings=v.normalize_embeddings,
-                image_download_headers=v.image_download_headers
-            ) for batch in v.content] # Todo: check if should be batching.
-        )
+        # TODO: Handle exception for single job, and allow others to run.
```
pandu-k, line 1559: What's the plan for this?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1535,17 +1550,24 @@ def get_query_vectors_from_jobs(queries: List[BulkSearchQueryEntity], qidx_to_jo
...
 
-def _bulk_vector_text_search(config: Config, queries: List[BulkSearchQueryEntity], device=None, raise_on_searchable_attribs=False, result_count: int = 5) -> List[Dict]:
+def create_empty_query_response(queries: List[BulkSearchQueryEntity]) -> List[Dict]:
+    return list(
+        map(
+            lambda x: {"hits": []}, queries
+        )
+    )
+
+def _bulk_vector_text_search(config: Config, queries: List[BulkSearchQueryEntity], device=None) -> List[Dict]:
```
pandu-k, line 1669: what is this function?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1569,15 +1592,17 @@ def _bulk_vector_text_search(config: Config, queries: List[BulkSearchQueryEntity
...
         query_to_body_parts[qidx] = body
         query_to_body_count[qidx] = len(body)
 
-    aggregate_body = [line for body in query_to_body_parts.values() for line in body]
+    # Combine all msearch request bodies into one request body.
+    aggregate_body = functools.reduce(lambda x, y: x + y, query_to_body_parts.values())
     if not aggregate_body:
-        return []
+        # Must return empty response, per search query
+        return list(map(lambda x: {"hits": []}, queries))
```
pandu-k: you can call the create_empty_query_response() function you declared

src/marqo/tensor_search/tensor_search.py
pandu-k: It's looking like different indents to me: 
image

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1763,51 +1784,51 @@ def _vector_text_search(
...
-                    "exclude": ["__chunks.__vector_*"]
-                }
         }
 
         field_names = list(index_info.get_text_properties().keys())
         if attributes_to_retrieve is not None:
             search_query["_source"] = {"include": attributes_to_retrieve} if len(attributes_to_retrieve) > 0 else False
 
-            if filter_string is not None:
-                search_query["query"]["nested"]["query"]["knn"][f"{TensorField.chunks}.{vector_field}"][
```
pandu-k: not sure why it showing different indents here, they looks the same to me

tests/tensor_search/test_bulk_search.py
pandu-k: Also:
- multimodal search queries
- tests where one of the indexes isn’t in the index cache
- Many more tests with more than 1 query

src/marqo/tensor_search/models/api_models.py
Jeadie: Nothing (tested on basic curl). In essence the original typing was just wrong.

src/marqo/version.py
Jeadie: I don't remember changing this, but yes.

src/marqo/tensor_search/tensor_search.py
Jeadie: I thought I had removed line 1336, thinking we needed 1337 to work. I shall remove latter and check. Return type is confusing but a) it is the true return type and b) yes, we shall be overwritting this soon nonetheless.

src/marqo/tensor_search/tensor_search.py
Jeadie: Not in this PR. I don't think it's worth doing until your PR is added.

src/marqo/tensor_search/tensor_search.py
Jeadie: Will add docstring. This is the equivalent of `_vector_text_search(`

src/marqo/tensor_search/tensor_search.py
Jeadie: As stated, will add in future PR once bulk changes are in.

pandu-k:
unit tests: https://github.com/marqo-ai/marqo/actions/runs/4370424023

pandu-k:
unit tests: https://github.com/marqo-ai/marqo/actions/runs/4370667619

pandu-k:
unit tests: https://github.com/marqo-ai/marqo/actions/runs/4370667619

Jeadie:
Unit tests pass: https://github.com/marqo-ai/marqo/actions/runs/4410559397

pandu-k (approved):

PR #372, opened 2023-03-06
Automatic model ejection

wanliAlex:
Maqo will automatically eject models based on the memory usage

we can't eject a model automatically.

Maqo will automatically eject models based on the memory usage

noe

yes

no

will do


src/marqo/s2_inference/constants.py
```diff
@@ -0,0 +1,27 @@
...
+    "vit-g" : 5,
+    "vit-h" : 5,
+    "vit-bigg-14":6,
+}
+
+# Set this to be a special number so we can tell the model size is from default
+DEFAULT_MODEL_SIZE = 0.66
+
+MARQO_MAX_CPU_MODEL_MEMORY = read_env_vars_and_defaults('MARQO_MAX_CPU_MODEL_MEMORY')
+MARQO_MAX_CUDA_MODEL_MEMORY = read_env_vars_and_defaults('MARQO_MAX_CUDA_MODEL_MEMORY')
```
pandu-k: These are best defined here: https://github.com/marqo-ai/marqo/blob/mainline/src/marqo/tensor_search/configs.py#L28

And accessed like this: 
```
utils.read_env_vars_and_defaults(EnvVars.MARQO_MAX_CUDA_MODEL_MEMORY)
```

constants.py should be reserved for hardcoded vars (unlike env vars).

src/marqo/s2_inference/model_registry.py
```diff
@@ -276,7 +276,8 @@ def _get_open_clip_properties() -> Dict:
              'dimensions': 1024,
              'note': 'open_clip models',
              'type': 'open_clip',
-             'pretrained': 'laion2b_s12b_b42k'},
+             'pretrained': 'laion2b_s12b_b42k',
+             'model_size': 5,},
```
pandu-k: Interesting. Was this model a lot larger than its peers? What method should we use the figure this out? Manual testing for the most commonly used models?

src/marqo/s2_inference/processing/image.py
```diff
@@ -211,9 +211,11 @@ def _load_and_cache_model(self):
 
             self.model, self.preprocess = func(self.model_name, self.device)
 
-            available_models[model_cache_key] = self.model, self.preprocess
+            available_models[model_cache_key] = {"model": (self.model, self.preprocess),
+                                                 "time_stamp" : datetime.datetime.now()}
```
pandu-k: Is this better as `most_recently_used_time`? `time_stamp` doesn't explain exactly what this is for

src/marqo/s2_inference/reranking/model_utils.py
```diff
@@ -288,11 +290,12 @@ def load_hf_cross_encoder_model(model_name: str, device: str = 'cpu') -> Dict:
     model_cache_key = _create_model_cache_key(model_name, device)
 
     if model_cache_key in available_models:
-        model, tokenizer = available_models[model_cache_key] 
+        model, tokenizer = available_models[model_cache_key]["model"]
     else:
         logger.info(f"loading {model_name} on device {device} and adding to cache...")    
         model = AutoModelForSequenceClassification.from_pretrained(model_name).to(device)
         tokenizer = AutoTokenizer.from_pretrained(model_name)
+        available_models[model_cache_key] = {"model":(model, tokenizer), "time_stamp": datetime.datetime.now()}
```
pandu-k: Why is it `(model, tokenizer)` here? shouldn't the second val be processor?

src/marqo/s2_inference/reranking/model_utils.py
```diff
@@ -313,11 +316,11 @@ def load_owl_vit(model_name: str, device: str = 'cpu') -> Dict:
 
     if model_cache_key in available_models:
         logger.info(f"loading {model_cache_key} from cache...")
-        model, processor = available_models[model_cache_key] 
+        model, processor = available_models[model_cache_key]["model"]
     else:
         processor = OwlViTProcessor.from_pretrained(model_name)
         model = OwlViTForObjectDetection.from_pretrained(model_name).to(device)
-        available_models[model_cache_key] = model, processor
+        available_models[model_cache_key] = {"model": (model, processor), "time_stamp" : datetime.datetime.now()}
```
pandu-k: does processor == device?

src/marqo/s2_inference/s2_inference.py
```diff
@@ -81,17 +86,98 @@ def _update_available_models(model_cache_key: str, model_name: str, validated_mo
                              normalize_embeddings: bool) -> None:
     """loads the model if it is not already loaded
     """
+    model_size = validated_model_properties.get("model_size", constants.MODEL_TYPE_SIZE_MAPPING.get(validated_model_properties["type"], 1))
```
pandu-k: Throw warning for unknown model type, maybe?

src/marqo/s2_inference/s2_inference.py
```diff
@@ -81,17 +86,98 @@ def _update_available_models(model_cache_key: str, model_name: str, validated_mo
...
+        sorted_key_in_device = sorted(model_cache_key_in_device,
+                                      key=lambda x: available_models[x]["time_stamp"])
+        for key in sorted_key_in_device:
+            logger.info(f"Eject model = `{key.split('||')[0]}` with size = `{available_models[key].get('model_size', constants.DEFAULT_MODEL_SIZE)}` from device = `{device}` "
+                        f"to save space for model = `{model_name}`.")
+            del available_models[key]
+            if check_device_memory_status(device, model_size):
+                return True
+
+        if check_device_memory_status(device, model_size) is False:
```
pandu-k: Perhaps we can calculate this before ejecting all the models? So that the user doesn't have to wait around to reload their previous models

src/marqo/tensor_search/configs.py
```diff
@@ -39,6 +39,8 @@ def default_env_vars() -> dict:
         EnvVars.MARQO_MAX_CONCURRENT_SEARCH: 8,
         EnvVars.MARQO_THREAD_EXPIRY_TIME: 1800,     # 30 minutes
         EnvVars.MARQO_ENABLE_THROTTLING: "TRUE",
-        EnvVars.MARQO_LOG_LEVEL: "info"             # This env variable is set to "info" by default in run_marqo.sh, which overrides this value
+        EnvVars.MARQO_LOG_LEVEL: "info",           # This env variable is set to "info" by default in run_marqo.sh, which overrides this value
+        EnvVars.MARQO_MAX_CPU_MODEL_MEMORY: 4,
+        EnvVars.MARQO_MAX_CUDA_MODEL_MEMORY: 4,
```
pandu-k: Make sure the cloud team knows they have to manually set these values, so they don't get any surprises

tests/s2_inference/test_automatic_model_ejection.py
```diff
@@ -0,0 +1,133 @@
+import unittest
```
pandu-k: What sort of manual testing has been done?

src/marqo/s2_inference/processing/image.py
```diff
@@ -211,9 +211,11 @@ def _load_and_cache_model(self):
 
             self.model, self.preprocess = func(self.model_name, self.device)
 
-            available_models[model_cache_key] = self.model, self.preprocess
+            available_models[model_cache_key] = {"model": (self.model, self.preprocess),
+                                                 "most_recently_used_time" : datetime.datetime.now()}
```
pandu-k: nit, important keys are probably better declared by an enum:

```
class ModelCacheObjects:
    model = "model"
    most_recently_used_time = "most_recently_used_time"
```

This prevents hard to detect bugs (like from a typo)

src/marqo/s2_inference/s2_inference.py
```diff
@@ -79,19 +83,105 @@ def _create_model_cache_key(model_name: str, device: str, model_properties: dict
 def _update_available_models(model_cache_key: str, model_name: str, validated_model_properties: dict,
                              device: str,
                              normalize_embeddings: bool) -> None:
-    """loads the model if it is not already loaded
+    """loads the model if it is not already loaded.
+    Note this method assume the model_properties are validated.
     """
+    model_size = validated_model_properties.get("model_size", constants.MODEL_TYPE_SIZE_MAPPING.get(validated_model_properties["type"], 1))
     if model_cache_key not in available_models:
+        device_memory_manage(model_name, validated_model_properties, device)
```
pandu-k: `device_memory_manage` is a bit of a vague name. can it be named based on what it does?

src/marqo/s2_inference/s2_inference.py
```diff
@@ -79,19 +83,105 @@ def _create_model_cache_key(model_name: str, device: str, model_properties: dict
...
+    if "model_size" in model_properties:
+        return model_properties["model_size"]
+
+    name_info = (model_name + model_properties.get("name", "")).lower().replace("/", "-")
+    for name, size in constants.MODEL_NAME_SIZE_MAPPING.items():
+        if name in name_info:
+            return size
+
+    type = model_properties["type"]
+    return constants.MODEL_TYPE_SIZE_MAPPING.get(type, constants.DEFAULT_MODEL_SIZE)
```
pandu-k: Should we log a warning if we cant find the type in the MODEL_TYPE_SIZE_MAPPING?

src/marqo/s2_inference/s2_inference.py
```diff
@@ -79,19 +83,105 @@ def _create_model_cache_key(model_name: str, device: str, model_properties: dict
...
+
+
+def device_memory_manage(model_name: str, model_properties: dict, device: str) -> bool:
```
pandu-k: maybe `def load_model_to_device()...`

src/marqo/s2_inference/s2_inference.py
```diff
@@ -79,19 +83,105 @@ def _create_model_cache_key(model_name: str, device: str, model_properties: dict
...
+    Args:
+        model_name: The name of the model to load
+        model_properties: The model properties of the model
+        device: The target device to laod the model
+    Returns:
+        True we have enough space for the model
+        Raise an error and return False if we can't find enough space for the model.
+    '''
+    model_size = get_model_size(model_name, model_properties)
+    if check_device_memory_status(device, model_size):
```
pandu-k: This isn't threadsafe - remember this can be potentially be updated by multiple threads...

src/marqo/s2_inference/s2_inference.py
```diff
@@ -79,19 +83,105 @@ def _create_model_cache_key(model_name: str, device: str, model_properties: dict
...
+        return True
+    else:
+        model_cache_key_in_device = [key for key in list(available_models) if key.endswith(device)]
```
pandu-k: nit: `model_cache_key_for_device` makes more sense to me

src/marqo/s2_inference/s2_inference.py
```diff
@@ -79,19 +83,105 @@ def _create_model_cache_key(model_name: str, device: str, model_properties: dict
...
+        sorted_key_in_device = sorted(model_cache_key_in_device,
+                                      key=lambda x: available_models[x]["most_recently_used_time"])
```
pandu-k: Thread safety - what if another thread ejects the model?

src/marqo/s2_inference/s2_inference.py
pandu-k: Can we calculate this before ejecting the models, so we don't needlessly eject models?

src/marqo/s2_inference/s2_inference.py
```diff
@@ -79,19 +83,105 @@ def _create_model_cache_key(model_name: str, device: str, model_properties: dict
...
+            if check_device_memory_status(device, model_size):
+                return True
+
+        if check_device_memory_status(device, model_size) is False:
+            raise ModelCacheManageError(f"Marqo CANNOT find enough space to load model = `{model_name}` in device = `{device}`.\n"
+                                        f"Marqo tried to eject all the models on this device = `{device}` but still can't find enough space. \n"
+                                        f"Please use a smaller model or increase the memory threshold.")
+
+
+def check_device_memory_status(device:str, model_size: Union[float, int] = 1) -> bool:
```
pandu-k: be more precise with this language, explain exactly what the function does: `check_memory_threshold_for_model()`

Also, a default parameter for `model_size` doesn't make sense especially when you can get this from a `constants.DEFAULT_MODEL_SIZE` anyway.

src/marqo/s2_inference/s2_inference.py
```diff
@@ -79,19 +83,105 @@ def _create_model_cache_key(model_name: str, device: str, model_properties: dict
...
+    if check_device_memory_status(device, model_size):
+        return True
+    else:
+        model_cache_key_in_device = [key for key in list(available_models) if key.endswith(device)]
+        sorted_key_in_device = sorted(model_cache_key_in_device,
+                                      key=lambda x: available_models[x]["most_recently_used_time"])
+        for key in sorted_key_in_device:
+            logger.info(f"Eject model = `{key.split('||')[0]}` with size = `{available_models[key].get('model_size', constants.DEFAULT_MODEL_SIZE)}` from device = `{device}` "
+                        f"to save space for model = `{model_name}`.")
+            del available_models[key]
```
pandu-k, line 236: We probably need a try/except (in case the model is deleted via another thread)

src/marqo/errors.py
```diff
@@ -110,6 +110,9 @@ class IndexAlreadyExistsError(__InvalidRequestError):
     code = "index_already_exists"
     status_code = HTTPStatus.CONFLICT
 
+class ModelCacheManageError(__InvalidRequestError):
+    code = "model_cache_manage_error"
```
pandu-k: `model_cache_managent_error` reads nicer

src/marqo/s2_inference/avail_models.py
pandu-k, line 1: To be deleted (acknowledging that this is in progress)

src/marqo/s2_inference/constants.py
```diff
@@ -0,0 +1,24 @@
...
+    # model (GB) based on type.
+    "open_clip" : 1,
+    "clip" : 1,
+    "sbert" : 0.7,
+    "random" : 0.1,
+    "multilingual_clip" : 5,
+    "clip_onnx":1,
+    'sbert_onnx' : 0.7,
+}
+
```
pandu-k: Some documentation via comments would be good, for example `"MODEL_TYPE_SIZE_MAPPING"` represents defaults for models of their type, and that `MODEL_NAME_SIZE_MAPPING` is for the precise model names. 

Also documenting the order of precedence: for example: "matching a model `MODEL_NAME_SIZE_MAPPING` takes priority over matching MODEL_TYPE_SIZE_MAPPING etc"

src/marqo/s2_inference/processing/image.py
```diff
@@ -211,9 +211,11 @@ def _load_and_cache_model(self):
 
             self.model, self.preprocess = func(self.model_name, self.device)
 
-            available_models[model_cache_key] = self.model, self.preprocess
+            available_models[model_cache_key] = {AvailableModelsKey.model: (self.model, self.preprocess),
+                                                 AvailableModelsKey.most_recently_used_time : datetime.datetime.now()}
+
         else:
-            self.model, self.preprocess = available_models[model_cache_key]
+            self.model, self.preprocess = available_models[model_cache_key]["model"]
```
pandu-k: nit: use `AvailableModelsKey.model` here

src/marqo/s2_inference/reranking/model_utils.py
```diff
@@ -251,7 +253,7 @@ def load_sbert_cross_encoder_model(model_name: str, device: str = 'cpu', max_len
     model_cache_key = _create_model_cache_key(model_name, device)
 
     if model_cache_key in available_models:
-        model = available_models[model_cache_key] 
+        model = available_models[model_cache_key]["model"]
```
pandu-k: nit use `AvailableModelsKey.model` here

src/marqo/s2_inference/s2_inference.py
```diff
@@ -79,19 +93,38 @@ def _create_model_cache_key(model_name: str, device: str, model_properties: dict
...
+                                                     AvailableModelsKey.most_recently_used_time: most_recently_used_time,
+                                                     AvailableModelsKey.model_size: model_size}
+                logger.info(
+                    f'loaded {model_name} on device {device} with normalization={normalize_embeddings} at time={most_recently_used_time}.')
+            except:
+                raise ModelLoadError(
+                    f"Unable to load model={model_name} on device={device} with normalization={normalize_embeddings}. "
+                    f"If you are trying to load a custom model, "
+                    f"please check that model_properties={validated_model_properties} is correct "
+                    f"and the model has valid access permission. ")
```
pandu-k: clarity: `and the model has valid access permission.` -> `and Marqo has access to the weights file`

src/marqo/s2_inference/s2_inference.py
```diff
@@ -79,19 +93,38 @@ def _create_model_cache_key(model_name: str, device: str, model_properties: dict
...
-            raise ModelLoadError(
-                f"Unable to load model={model_name} on device={device} with normalization={normalize_embeddings}. "
-                f"If you are trying to load a custom model, "
-                f"please check that model_properties={validated_model_properties} is correct "
-                f"and the model has valid access permission. ")
+        if lock.locked():
+            raise ModelCacheManageError("Request rejected, as this request attempted to update the model cache, while"
+                                        "another request was updating the model cache at the same time.\n"
+                                        "Please wait for 10 seconds and send the request again.\n"
+                                        "If this problem persists, check `https://docs.marqo.ai/0.0.17/` for more info.")
```
pandu-k: The base error class requests users to make an issue (see [here](https://github.com/marqo-ai/marqo/blob/58541beb5e3326ababab02f15f7a712d7834bce6/src/marqo/errors.py#L67)).

Checking the docs isn't a silver bullet to solve their problems, but it could help them diagnose it quicker. So rather than advising: 
```
"If this problem persists, check `https://docs.marqo.ai/0.0.17/` for more info."
```
we could just add a reference to the docs to make it easier to find: 
```
"Marqo's documentation can be found here: `https://docs.marqo.ai/latest/`.
```

src/marqo/s2_inference/s2_inference.py
```diff
@@ -128,6 +161,126 @@ def _validate_model_properties(model_name: str, model_properties: dict) -> dict:
...
+                           available_models.items() if key.endswith("cpu")])
+        threshold = float(read_env_vars_and_defaults(EnvVars.MARQO_MAX_CPU_MODEL_MEMORY))
+    else:
+        raise ModelCacheManageError(
+            f"Unable to check the device cache for device=`{device}`. The model loading will proceed"
+            f"without device cache check. This might break down Marqo if too many models are loaded.")
+    if model_size > threshold:
+        raise ModelCacheManageError(
+            f"You are trying to load a model with size = `{model_size}` into device = `{device}`, which is larger than the device threshlod = `{threshold}`."
+            f"We CANNOT find enough space for the model. Please change the threshold by setting the environment variables.\n"
```
pandu-k: `We CANNOT` -> `Marqo cannot`

pandu-k: ` by setting the environment variables` ->  `by adjusting the environment variable`

src/marqo/s2_inference/s2_inference.py
```diff
@@ -128,6 +161,126 @@ def _validate_model_properties(model_name: str, model_properties: dict) -> dict:
...
+            f"You can check the detailed information at `https://docs.marqo.ai/0.0.17/Advanced-Usage/configuration/`.")
```
pandu-k: `You can check the detailed information at` -> `You can find more detailed information at`

src/marqo/s2_inference/s2_inference.py
```diff
@@ -128,6 +161,126 @@ def _validate_model_properties(model_name: str, model_properties: dict) -> dict:
...
+    '''
+    if "model_size" in model_properties:
+        return model_properties["model_size"]
+
+    name_info = (model_name + model_properties.get("name", "")).lower().replace("/", "-")
+    for name, size in constants.MODEL_NAME_SIZE_MAPPING.items():
+        if name in name_info:
+            return size
+
+    type = model_properties["type"]
```
pandu-k: are we guaranteed to have type in model_properties?

src/marqo/s2_inference/s2_inference.py
```diff
@@ -128,6 +161,126 @@ def _validate_model_properties(model_name: str, model_properties: dict) -> dict:
...
+    Args:
+        model_name (str): Actual model_name to be fetched from external library
+                        prefer passing it in the form of model_properties['name']
+        device (str, optional): _description_. Defaults to 'cpu'.
+
+    Returns:
+        Any: _description_
+    """
+    print(f"loading for: model_name={model_name} and properties={model_properties}")
+    from marqo.s2_inference.s2_inference import _get_model_loader, get_default_seq_length, get_default_device
```
pandu-k: What is the reasoning behind having these import statements here?

src/marqo/s2_inference/s2_inference.py
```diff
@@ -128,6 +161,126 @@ def _validate_model_properties(model_name: str, model_properties: dict) -> dict:
...
+    for name, size in constants.MODEL_NAME_SIZE_MAPPING.items():
+        if name in name_info:
+            return size
+
+    type = model_properties["type"]
+    return constants.MODEL_TYPE_SIZE_MAPPING.get(type, constants.DEFAULT_MODEL_SIZE)
+
+
+def _load_model(model_name: str, model_properties: dict, device: Optional[str] = None) -> Any:
+    """_summary_
```
pandu-k, line 304: is it worth explicitly making this only callable from update_available models?

tests/s2_inference/test_automatic_model_ejection.py
```diff
@@ -0,0 +1,214 @@
...
+        if  "Request rejected, as this request attempted to update the model cache" not in e.message:
+            q.put(AssertionError)
+        pass
+
+
+class TestAutomaticModelEject(unittest.TestCase):
+    def setUp(self) -> None:
+        clear_loaded_models()
+
+        self.inde_name = "test_index"
```
pandu-k: typo

tests/s2_inference/test_automatic_model_ejection.py
```diff
@@ -0,0 +1,214 @@
...
+        t.start()
+        for i in range(3):
+            t = threading.Thread(target=racing_vectorise_call, args=(test_model, test_content, q))
+            threads.append(t)
+            t.start()
+
+        for t in threads:
+            t.join()
+
+        assert q.empty()
```
pandu-k: my preference is to put model_load_rejection errors into the shared data structure (here a queue), and then assert that each error is the correct type. That way we can assert about the expected number of errors and that they re the correct error type (for example, perhaps q remains empty because the test doesn't work)

Mehul-Porwal (approved):

src/marqo/s2_inference/s2_inference.py
```diff
@@ -79,19 +93,39 @@ def _create_model_cache_key(model_name: str, device: str, model_properties: dict
...
+                most_recently_used_time = datetime.datetime.now()
+                available_models[model_cache_key] = {AvailableModelsKey.model: _load_model(model_name,
+                                                                                           validated_model_properties,
+                                                                                           device=device,
+                                                                                           calling_func = _update_available_models.__name__),
+                                                     AvailableModelsKey.most_recently_used_time: most_recently_used_time,
+                                                     AvailableModelsKey.model_size: model_size}
+                logger.info(
+                    f'loaded {model_name} on device {device} with normalization={normalize_embeddings} at time={most_recently_used_time}.')
+            except:
```
pandu-k: do we want to log the error that caused the ModelLoadError? Something like this:

logging.warning(f"Error loading model `{model_name}`, reason: {e} ")

src/marqo/s2_inference/s2_inference.py
```diff
@@ -128,6 +162,129 @@ def _validate_model_properties(model_name: str, model_properties: dict) -> dict:
...
+        if _check_memory_threshold_for_model(device, model_size, calling_func = _validate_model_into_device.__name__) is False:
+            raise ModelCacheManagementError(
+                f"Marqo CANNOT find enough space to load model = `{model_name}` in device = `{device}`.\n"
+                f"Marqo tried to eject all the models on this device = `{device}` but still can't find enough space. \n"
+                f"Please use a smaller model or increase the memory threshold.")
+
+
+def _check_memory_threshold_for_model(device: str, model_size: Union[float, int], calling_func: str =None) -> bool:
+    '''
+    Note: this function should only be called by `_update_available_models` for threading safeness.
```
pandu-k: nit: isn't it called by `_validate_model_into_device` which is called by `_update_available_models`

pandu-k (approved):

src/marqo/s2_inference/onnx_clip_utils.py
```diff
@@ -122,7 +122,7 @@ def encode_text(self, sentence, normalize=True):
 
     def encode_image(self, images, normalize=True):
         if isinstance(images, list):
-            image_input = format_and_load_CLIP_images(images)
+            image_input = format_and_load_CLIP_images(images, {})
```
pandu-k, line 125: nit: better to instantiate a dict like this: `dict()` as `{}` is ambiguous between an empty set vs empty dict

src/marqo/s2_inference/s2_inference.py
```diff
@@ -3,20 +3,32 @@
 """
 import functools
 import numpy as np
-from marqo.s2_inference.errors import VectoriseError, InvalidModelPropertiesError, ModelLoadError, UnknownModelError, ModelNotInCacheError
+from marqo.errors import ModelCacheManagementError
+from marqo.s2_inference.errors import VectoriseError, InvalidModelPropertiesError, ModelLoadError, UnknownModelError, \
```
pandu-k: nit: nicer to use brackets: 
```
from marqo.s2_inference.errors import (
    VectoriseError, InvalidModelPropertiesError, ModelLoadError, 
    UnknownModelError, ModelNotInCacheError
)
```

pandu-k: rather than backslashes

src/marqo/s2_inference/s2_inference.py
```diff
@@ -115,19 +127,41 @@ def _create_model_cache_key(model_name: str, device: str, model_properties: dict
...
+                             f"Error message is {str(e)}")
+                raise ModelLoadError(
+                    f"Unable to load model={model_name} on device={device} with normalization={normalize_embeddings}. "
+                    f"If you are trying to load a custom model, "
+                    f"please check that model_properties={validated_model_properties} is correct "
+                    f"and Marqo has access to the weights file.")
+    else:
+        most_recently_used_time = datetime.datetime.now()
+        logger.debug(f'renew {model_name} on device {device} with new time={most_recently_used_time}.')
+        available_models[model_cache_key][AvailableModelsKey.most_recently_used_time] = most_recently_used_time
```
pandu-k: what if the model has been de-loaded by another process in the meantime?  

We don't have to lock it or anything, but we need to ensure that users are getting a decent error (not a 5xx resulting from a KeyError).

For example:

(Some 4xx error code): Another request has deloaded the requested model `{model_name}` while Marqo was trying to use it (perhaps due to memory constraints). Please try your request again.

pandu-k: Because this update pattern can be found in multiple parts of the code, it could be best to create a function that can be called in multiple places

src/marqo/s2_inference/s2_inference.py
```diff
@@ -164,13 +198,142 @@ def _validate_model_properties(model_name: str, model_properties: dict) -> dict:
...
+        model_size: the size of the model to load
+    Returns:
+        True if we have enough space
+        False if we don't have enough space
+    '''
+    if calling_func not in ["unit_test", "_validate_model_into_device"]:
+        raise RuntimeError(f"The function `{_check_memory_threshold_for_model.__name__}` should only be called by "
+                           f"`unit_test` or `_validate_model_into_device` for threading safeness.")
+
+    if device.startswith("cuda"):
```
pandu-k, line 264: it becomes a bit ambiguous - is the CUDA threshold per device or grand total across all CUDA devices ? Either way we need to ensure we add this to the documentation

src/marqo/s2_inference/s2_inference.py
```diff
@@ -115,19 +127,41 @@ def _create_model_cache_key(model_name: str, device: str, model_properties: dict
 def _update_available_models(model_cache_key: str, model_name: str, validated_model_properties: dict,
                              device: str,
                              normalize_embeddings: bool) -> None:
-    """loads the model if it is not already loaded
+    """loads the model if it is not already loaded.
+    Note this method assume the model_properties are validated.
     """
+    model_size = get_model_size(model_name, validated_model_properties)
```
pandu-k: this doesn't need to be calculated if the model is already loaded right? maybe add under `if model_cache_key not in available_models:`

PR #373, opened 2023-03-07
Bulk vectorising for bulk search

pandu-k:

* **Other information**:


pandu-k:
bulk search tests passed 
unit tests: https://github.com/marqo-ai/marqo/actions/runs/4351112004

pandu-k:
to do: max batch size

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1325,18 +1324,30 @@ def _lexical_search(
     return {'hits': res_list}
 
 
-def construct_vector_input_batches(query: Union[str, Dict], index_info) -> Union[List[str], List[List[str]]]:
+def construct_vector_input_batches(query: Union[str, Dict], index_info) -> Tuple[List[str], List[str]]:
```
Jeadie, line 1324: Perhaps this could be better as `Dict[ModalityType, List[str]]`. It'd be easier to read, and if we add other modalities more simply

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1539,27 +1579,76 @@ def get_query_vectors_from_jobs(queries: List[BulkSearchQueryEntity], qidx_to_jo
...
+                    possible_jobs=qidx_to_job[qidx],
+                    jobs=jobs,
+                    job_to_vectors=job_to_vectors,
+                    treat_urls_as_images=index_info.index_settings[NsField.index_defaults][NsField.treat_urls_and_pointers_as_images],
+                    content=content),
+                 weight,
+                 content
+                ) for content, weight in ordered_queries
+            ]
+            # TODO how doe we ensure order?
```
Jeadie, line 1602: Yeah, I think a big question of going from `qidx_to_job: Dict[Qidx, VectorisedJobPointer]` to `Dict[Qidx, List[VectorisedJobPointer]]` is how we maintain order of content.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1702,8 +1800,8 @@ def _vector_text_search(
     else:  # is dict:
         ordered_queries = list(query.items())
         if index_info.index_settings[NsField.index_defaults][NsField.treat_urls_and_pointers_as_images]:
-            text_queries = [k for k, _ in ordered_queries if _is_image(k)]
```
Jeadie, line 1711: It seems this has always been a bug. What affect does this have, just a naming switch?

src/marqo/tensor_search/tensor_search.py
pandu-k: Yep that would make sense. But it might be a while before we add another modality, and we could cross that bridge when we get to it.

src/marqo/tensor_search/tensor_search.py
pandu-k: I can't think of how this affects maintaining order of content, as the important thing is that `VectorisedJobPointer` points to the correct locations. See how pointers are added [here](https://github.com/marqo-ai/marqo/pull/373/files#diff-bc6ddb28e3809766b86804afac01ac632de1e691377c5d3f43d0333b97ba79faR1497). 

What are the situations where you think this could fail? I can make some extra tests if you have ideas

src/marqo/tensor_search/tensor_search.py
pandu-k: I think this had no real effect on the execution, as the point of this part of the code was just to separate image and text queries from each other. Just a naming switch, as you said. But this caused an issue when it was refactored into bulk search (there were issues when checking for the jobs' content-type)

Jeadie (approved):

PR #390, opened 2023-03-14
Dupe IDs are handled when use_existing_tensors=True

vicilliar:
2 bug fixes

1. When `use_existing_tensors=True` and docs are added with duplicate IDs, a MarqoWebError with no status code is thrown.
2. When a doc with no chunks is replaced with `use_existing_tensors=True`, a KeyError occurs because it looks for `'_source`['__chunks']`

1. When `use_existing_tensors=True` and docs are added with duplicate IDs, docs are added normally. The last doc in the list with the same ID is the one that gets kept.
2. Chunkless docs simply return empty lists of chunks
3. MarqoWebError has a status code of 500 by default now.
4. Unit tests have been added to test duplicate IDs with and without `use_existing_tensors` and the chunkless docs bug

No

Yes

### DIFF
--- a/src/marqo/tensor_search/tensor_search.py
+++ b/src/marqo/tensor_search/tensor_search.py
@@ -369,6 +369,7 @@ def _infer_opensearch_data_type(
 
 def _get_chunks_for_field(field_name: str, doc_id: str, doc):
     # Find the chunks with a specific __field_name in a doc
+    # Note: for a chunkless doc (nothing was tensorised) --> doc["_source"]["__chunks"] == []
     return [chunk for chunk in doc["_source"]["__chunks"] if chunk["__field_name"] == field_name]
 
 
@@ -543,7 +544,6 @@ def add_documents(config: Config, index_name: str, docs: List[dict], auto_refres
             # Check if content of this field changed. If no, skip all chunking and vectorisation
             if ((update_mode == 'replace') and use_existing_tensors and existing_doc["found"]
                     and (field in existing_doc["_source"]) and (existing_doc["_source"][field] == field_content)):
-                # logger.info(f"Using existing vectors for doc {doc_id}, field {field}. Content remains unchanged.")
                 chunks_to_append = _get_chunks_for_field(field_name=field, doc_id=doc_id, doc=existing_doc)
 
             # Chunk and vectorise, since content changed.
@@ -653,9 +653,7 @@ def add_documents(config: Config, index_name: str, docs: List[dict], auto_refres
                         TensorField.field_content: text_chunk,
                         TensorField.field_name: field
                     })
-
-            # Add chunks_to_append along with doc metadata to total chunks
-
+            
             elif isinstance(field_content, dict):
                 if mappings[field]["type"]=="multimodal_combination":
                     combo_chunk, combo_document_is_valid, unsuccessful_doc_to_append, combo_vectorise_time_to_add,\
@@ -671,9 +669,11 @@ def add_documents(config: Config, index_name: str, docs: List[dict], auto_refres
                         if field not in new_obj_fields:
                             new_obj_fields[field] = set()
                         new_obj_fields[field] = new_obj_fields[field].union(new_fields_from_multimodal_combination)
+                        # TODO: we may want to use chunks_to_append here to make it uniform with use_existing_tensors and normal vectorisation
                         chunks.append({**combo_chunk, **chunk_values_for_filtering})
                         continue
-
+            
+            # Add chunks_to_append along with doc metadata to total chunks
             for chunk in chunks_to_append:
                 chunks.append({**chunk, **chunk_values_for_filtering})
 
@@ -895,6 +895,7 @@ def _get_documents_for_upsert(
             f"set by the environment variable `{EnvVars.MARQO_MAX_RETRIEVABLE_DOCS}`")
 
     # Chunk Docs (get field name, field content, vectors)
+
     chunk_docs = [
         {"_index": index_name, "_id": doc_id,
          "_source": {"include": [f"__chunks.__field_content", f"__chunks.__field_name", f"__chunks.__vector_*"]}}
@@ -915,9 +916,11 @@ def _get_documents_for_upsert(
 
     # Combine the 2 query results (loop through each doc id)
     combined_result = []
-    for doc_id in document_ids:
+
+    for doc_id in valid_doc_ids:
         # There should always be 2 results per doc.
         result_list = [doc for doc in res["docs"] if doc["_id"] == doc_id]
+
         if len(result_list) == 0:
             continue
         if len(result_list) not in (2, 0):
@@ -927,7 +930,7 @@ def _get_documents_for_upsert(
         for result in result_list:
             if result["found"]:
                 doc_in_results = True
-                if result["_source"]["__chunks"] == []:
+                if ("__chunks" in result["_source"]) and (result["_source"]["__chunks"] == []):
                     res_data = result
                 else:
                     res_chunks = result
@@ -936,12 +939,14 @@ def _get_documents_for_upsert(
                 dummy_res = result
                 break
 
-        # Put the chunks list in res_data, so it's complete
+        # Put the chunks list in res_data, so it contains all doc data
         if doc_in_results:
-            res_data["_source"]["__chunks"] = res_chunks["_source"]["__chunks"]
+            # Only add chunks if not a chunkless doc
+            if res_chunks["_source"]:
+                res_data["_source"]["__chunks"] = res_chunks["_source"]["__chunks"]
             combined_result.append(res_data)
         else:
-            # This result just says that the doc was not found
+            # This result just says that the doc was not found ("found": False)
             combined_result.append(dummy_res)
 
     res["docs"] = combined_result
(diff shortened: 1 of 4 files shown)


vicilliar:
Run passed: 268 https://github.com/marqo-ai/marqo/actions/runs/4418255521/jobs/7745155005#logs

vicilliar:
monitor: run 270 https://github.com/marqo-ai/marqo/actions/runs/4419170567

src/marqo/tensor_search/tensor_search.py
```diff
@@ -452,8 +452,12 @@ def add_documents(config: Config, index_name: str, docs: List[dict], auto_refres
...
 
     if update_mode == 'replace' and use_existing_tensors:
-        # Get existing documents
-        doc_ids = [doc["_id"] for doc in docs if "_id" in doc]
+        doc_ids = []
+
+        # Iterate through the list in reverse, only latest doc with dupe id gets added.
+        for i in range(len(docs)-1, -1, -1):
+            if ("_id" in docs[i]) and (docs[i]["_id"] not in doc_ids):
+                doc_ids.append(docs[i]["_id"])
```
pandu-k, line 465: what if there is a difference between the docs? This may be possible because it this operation isn't atomic. For example:
```
[{"_id": "1", "title":"Hello"}, 
{"_id": "1", "title":"World"}, 
{"_id": "1", "__chunks":[{"__title_vector": [0.1, 0.2]}]}
{"_id": "1", "__chunks":[{"__title_vector": [2.1, 0.3]}]}]
```

tests/tensor_search/test_add_documents_use_existing_tensors.py
```diff
@@ -85,6 +86,100 @@ def test_use_existing_tensors_non_existing(self):
             document_id="123", show_vectors=True)
         self.assertEqual(use_existing_tensors_doc, regular_doc)
 
+        tensor_search.delete_index(config=self.config, index_name=self.index_name_1)
```
pandu-k: if you delete the index, then you don't really overwrite the previous document

src/marqo/tensor_search/tensor_search.py
vicilliar: In this case, only the last item would be considered `{"_id": "1", "__chunks":[{"__title_vector": [2.1, 0.3]}]}`.

pandu-k (approved):

tests/tensor_search/test_add_documents_use_existing_tensors.py
vicilliar: done

vicilliar:
Reports on bug found by vitus "chunks" key error:
```([{"_id": "proper id", "field_1": 5678},
            {"_id": ["bad", "id"], "field_1": "zzz"},
            {"_id": "proper id 2", "field_1": 90}], 2)
```
error happens when adding this doc repeatedly

This only happens when `use_existing_tensors` is True after that document ALREADY EXISTS. then it can't find chunks?

```
# Combine the 2 query results (loop through each doc id)
        combined_result = []
        for doc_id in document_ids:
            # There should always be 2 results per doc.
            result_list = [doc for doc in res["docs"] if doc["_id"] == doc_id]
            if len(result_list) == 0:
                continue
            if len(result_list) not in (2, 0):
                raise errors.InternalError(f"Internal error fetching old documents. "
                                           f"There are {len(result_list)} results for doc id {doc_id}.")

            for result in result_list:
                if result["found"]:
                    doc_in_results = True
>                   if result["_source"]["__chunks"] == []:
E                   KeyError: '__chunks'

src/marqo/tensor_search/tensor_search.py:930: KeyError
```

vicilliar:
things we know:
1. Does not matter if original doc was placed there with update or replace.
2. It breaks when field content is int, but not str. It makes the source completely empty.


res_chunks
```{'_id': 'proper id',
  '_index': 'my-test-index-1',
  '_primary_term': 1,
  '_seq_no': 0,
  '_source': {},
  '_version': 1,
  'found': True},
```

res_data
```
 {'_id': 'proper id',
  '_index': 'my-test-index-1',
  '_primary_term': 1,
  '_seq_no': 0,
  '_source': {'__chunks': [], 'field_1': 5678},
  '_version': 1,
  'found': True}
```

question: why does res_chunks have no chunks? in what situation would _source be empty?

vicilliar:
Another situation:
"""
Error happens with
- available_product_codes: List and the one to remove is source_image_url
- despite treat_urls_as_pointers
- regardless of model used
"""

vicilliar:
Diagnosis (3/15/23)

The error
```
_get_documents_for_upsert
    if result["_source"]["__chunks"] == []:
KeyError: '__chunks'
```

Happens when the following conditions occur
1. A document is created with no tensor fields, therefore it has no `_source["__chunks"]`
2. Add documents is called using `use_existing_tensors` with the id of the chunkless doc.

A chunkless doc is one of the following:
1. Doc without a string/list/anything else that gets tensorized as a field. 
- int, bool do not get tensorized
- using `non_tensor_fields` on all fields to be tensorized will do this
- Interestingly, lists should not get tensorized, but they do not have the same problem as int and bool

Solution:
If a doc is chunkless, one of the 2 results from `_get_documents_for_upsert` that was supposed to have chunks will have empty source like this: `'_source': {}`
```
# chunkless result
[{'_id': 'proper id 2',
  '_index': 'my-test-index-1',
  '_primary_term': 1,
  '_seq_no': 0,
  '_source': {},
  '_version': 1,
  'found': True},
  
 {'_id': 'proper id 2',
  '_index': 'my-test-index-1',
  '_primary_term': 1,
  '_seq_no': 0,
  '_source': {'__chunks': [], 'field_2': 123},
  '_version': 1,
  'found': True}]
 ```
 
 If empty source is found, set that as `res_chunks`.

File: src/marqo/tensor_search/tensor_search.py [hunk 3]
```diff
@@ -931,7 +934,7 @@ def _get_documents_for_upsert(
         for result in result_list:
             if result["found"]:
                 doc_in_results = True
-                if result["_source"]["__chunks"] == []:
+                if ("__chunks" in result["_source"]) and (result["_source"]["__chunks"] == []):
                     res_data = result
                 else:
                     res_chunks = result
```
pandu-k, line 940: so `res_chunks` get set to result if `"__chunks"` is not in result?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -940,12 +943,14 @@ def _get_documents_for_upsert(
                 dummy_res = result
```
pandu-k, line 943: nit: may be more appropriate to call this something like `not_found_res`

src/marqo/tensor_search/tensor_search.py
```diff
@@ -934,12 +943,14 @@ def _get_documents_for_upsert(
...
                 break
 
-        # Put the chunks list in res_data, so it's complete
+        # Put the chunks list in res_data, so it contains all doc data
         if doc_in_results:
-            res_data["_source"]["__chunks"] = res_chunks["_source"]["__chunks"]
+            # Only add chunks if not a chunkless doc
+            if res_chunks["_source"]:
+                res_data["_source"]["__chunks"] = res_chunks["_source"]["__chunks"]
```
pandu-k, line 950: are there any assumptions being made here? For example, are we guaranteed that `res_chunks["_source"] != {}` implies `"__chunks" in res_chunks["_source"]`

pandu-k: The concern is a possible KeyError

src/marqo/tensor_search/tensor_search.py
pandu-k: Are we guaranteed that `"__chunks" not in result["_source"]` implies `result` is meant to be tensor request (as opposed to a body request)?

pandu-k:
Running unit tests: https://github.com/marqo-ai/marqo/actions/runs/4432692586

src/marqo/tensor_search/tensor_search.py
vicilliar: This is because the idea is for `res_chunks`, we only include chunk-related content like
```
"_source": {"include": [f"__chunks.__field_content", f"__chunks.__field_name", f"__chunks.__vector_*"]}
```

In our testing, we only have 1 scenario wherein `"__chunks"` is not in the result, and that is when we make a tensor request and no tensors exist. When we do a body request, the behavior is that `"__chunks"` is always there, it's just empty. This is the best way I found to differentiate between a tensor and body request, and it passes all the edge cases I can think of

src/marqo/tensor_search/tensor_search.py
vicilliar: Throughout my testing, yes, if `res_chunks["_source"]` is non-empty, it must have the `"__chunks"` key. this is because in its query, it only includes things inside the `"__chunks"` key.

if you want it to be assumptionless, I suppose
```
if "__chunks" in res_chunks["_source"]:
      res_data["_source"]["__chunks"] = res_chunks["_source"]["__chunks"]
```

would do it

pandu-k: That makes sense, thanks for the explanation

pandu-k (approved):

PR #407, opened 2023-03-23
Add documents concurrency fix

pandu-k:
Bug fix

During the add documents call, if another user deletes the index, the add_documents user gets a 500 due to a keyerror during the mappings update.

Now, the add_documents call attempts to find index mappings from Marqo-os if it isn't in the cache. If the index has been deleted in this time, the add_documents user will get "IndexNotFound" error. 

No

In progress

### DIFF
--- a/src/marqo/tensor_search/backend.py
+++ b/src/marqo/tensor_search/backend.py
@@ -13,7 +13,9 @@ from marqo import errors
 #
 from typing import Iterable, List, Union, Optional, Tuple, Dict
 from marqo.tensor_search.index_meta_cache import get_cache
+from marqo.tensor_search.index_meta_cache import get_index_info as get_cached_index_info
 import pprint
+
 
 def get_index_info(config: Config, index_name: str) -> IndexInfo:
     """Gets useful information about the index. Also updates the IndexInfo cache
@@ -108,7 +110,7 @@ def add_customer_field_properties(config: Config, index_name: str,
             }
         }
     }
-    existing_info = get_index_info(config=config, index_name=index_name)
+    existing_info = get_cached_index_info(config=config, index_name=index_name)
     new_index_properties = existing_info.properties.copy()
 
     # copy fields to the chunk for prefiltering. If it is text, convert it to a keyword type to save space
--- a/src/marqo/tensor_search/backend.py
+++ b/src/marqo/tensor_search/backend.py
@@ -108,7 +108,7 @@ def add_customer_field_properties(config: Config, index_name: str,
             }
         }
     }
-    existing_info = get_cache()[index_name]
+    existing_info = get_index_info(config=config, index_name=index_name)
     new_index_properties = existing_info.properties.copy()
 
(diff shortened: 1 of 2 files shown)


src/marqo/tensor_search/backend.py
```diff
@@ -108,7 +108,7 @@ def add_customer_field_properties(config: Config, index_name: str,
             }
         }
     }
-    existing_info = get_cache()[index_name]
+    existing_info = get_index_info(config=config, index_name=index_name)
```
wanliAlex: to what extent will this change slow the `add_documents` process?

tests/tensor_search/test_index_meta_cache.py
```diff
@@ -577,3 +579,78 @@ def threaded_while(thread_num, loop_record):
...
+            tensor_search.create_vector_index(config=self.config, index_name=self.index_name_1,
+                                              index_settings={"index_defaults": {"model": "random"}})
+            clear_cache_thread = threading.Thread(target=clear_cache)
+            clear_cache_thread.start()
+            tensor_search.add_documents(**{
+                "config": self.config, "index_name": self.index_name_1, "auto_refresh": True,
+                "docs": [
+                    {"Title": "Blah"}, {"Title": "blah2"}, {"Title": "Blah3"}, {"Title": "Blah4"},
+                ]
+            })
```
wanliAlex, line 615: do we need a `clear_cache_thread.join()` to make it safer?

tests/tensor_search/test_index_meta_cache.py
```diff
@@ -577,3 +579,78 @@ def threaded_while(thread_num, loop_record):
...
+            return True
+        assert run()
+
+    def test_add_documents_to_non_existent_index(self):
```
wanliAlex, line 619: in general, under what circumstance will an `index_cached` get deleted but the `index` still exists?

wanliAlex (changes requested):

src/marqo/tensor_search/backend.py
pandu-k: thanks - updated.

tests/tensor_search/test_index_meta_cache.py
pandu-k: As in to wait for the thread to complete before continuing on with the next tests? I don't know how this will make it safer

tests/tensor_search/test_index_meta_cache.py
pandu-k: This should be quite rare:
1. thread_a does add_documents
2. thread_a checks the cache. the index is there
3. thread_b deletes the index. the index is removed from cache
4. thread_c, on another Marqo instance, creates an index with the same name
5. thread_a checks cache at the end of add_documents. The index is not there, but does actually exist

This is simulated in `test_add_documents_to_unknown_index()`

wanliAlex (approved):

PR #420, opened 2023-04-03
Validate index setting parameters when creating index

Jeadie:
# Changes
 - Validate `"ann_parameters"` index settings at creation time.
 - Currently, these settings, even if invalid, are stored as the index settings. It is only when we use these settings (e.g. `add_documents`), that an error occurs.
 - This PR validates correct parameters settings when we `create_index`

### DIFF
--- a/src/marqo/tensor_search/backend.py
+++ b/src/marqo/tensor_search/backend.py
@@ -78,7 +78,6 @@ def add_customer_field_properties(config: Config, index_name: str,
     Returns:
         HTTP Response
     """
-    engine = "lucene"
     existing_info = get_cached_index_info(config=config, index_name=index_name)
 
     # check if there is multimodal fie;ds and convert the fields name to a list with the same
--- a/src/marqo/tensor_search/models/settings_object.py
+++ b/src/marqo/tensor_search/models/settings_object.py
@@ -101,12 +101,21 @@ settings_schema = {
                     "properties": {
                         NsFields.ann_method: {
                             "type": "string",
+                            "enum": ["hnsw"],
                             "examples": [
                                 "hnsw"
                             ]
                         },
+                        NsFields.ann_engine: {
+                            "type": "string",
+                            "enum": ["lucene"],
+                            "examples": [
+                                "lucene"
+                            ]
+                        },
                         NsFields.ann_metric: {
                             "type": "string",
+                            "enum": ["l1", "l2", "linf", "cosinesimil"],
                             "examples": [
                                 "cosinesimil"
                             ]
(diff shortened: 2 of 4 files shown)


src/marqo/tensor_search/backend.py
```diff
@@ -78,7 +78,6 @@ def add_customer_field_properties(config: Config, index_name: str,
     Returns:
         HTTP Response
     """
-    engine = "lucene"
```
Jeadie, line 81: Not needed

src/marqo/tensor_search/models/settings_object.py
```diff
@@ -101,12 +101,21 @@
                     "properties": {
                         NsFields.ann_method: {
                             "type": "string",
+                            "enum": ["hnsw"],
                             "examples": [
                                 "hnsw"
                             ]
                         },
+                        NsFields.ann_engine: {
```
Jeadie, line 109: We can now let this be with the remaining parameters, but since we limit it to Lucene, it is now safe.

Jeadie:
Unit tests: https://github.com/marqo-ai/marqo/actions/runs/4602278489

src/marqo/tensor_search/models/settings_object.py
```diff
@@ -117,12 +126,14 @@
                             "properties": {
                                 NsFields.hnsw_ef_construction: {
                                     "type": "integer",
+                                    "minimum": 1,
```
pandu-k, line 129: do we need to put maximums in here? otherwise wouldn't they get a 500 from OpenSearch, if they chose a ridiculous number?

src/marqo/tensor_search/models/settings_object.py
```diff
@@ -117,12 +126,14 @@
...
                                     "examples": [
                                         128
                                     ]
                                 },
                                 NsFields.hnsw_m: {
                                     "type": "integer",
+                                    "minimum": 1,
```
pandu-k, line 136: do we need to put maximums in here? otherwise wouldn't they get a 500 from OpenSearch, if they chose a ridiculous number?

tests/tensor_search/test_create_index.py
pandu-k, line 1: Do we have a test to ensure that allowed hnsw params can work OK?

pandu-k, line 1: Do we check the index settings from OpenSearch itself, to ensure that that the kNN index has the expected settings?

src/marqo/tensor_search/models/settings_object.py
Jeadie: ef_construction of 500 would actually be rather reasonable. In fact, the default is 512. We'd have to figure out what upper bounds we'd want to set. I don't think this would be self-evident.

tests/tensor_search/test_create_index.py
Jeadie: This, and the above, have been tested manually. I don't know if we want these as unit tests, not as performance or just manual testing

pandu-k (approved):

PR #421, opened 2023-04-04
tests ensuring Marqo-os knn settings == Marqo knn settings

pandu-k:


tests/tensor_search/test_create_index.py
```diff
@@ -139,6 +139,71 @@ def test_create_vector_index_custom_index_settings(self):
...
+                NsField.index_defaults: custom_settings})
+        tensor_search.add_documents(
+            config=self.config, index_name=self.index_name_1, docs=[{"Title": "wowow"}], auto_refresh=True)
+        mappings = requests.get(
+            url=self.endpoint + "/" + self.index_name_1 + "/_mapping",
+            verify=False
+        ).json()
+        params = mappings[self.index_name_1]['mappings']['properties']['__chunks']['properties']['__vector_Title']['method']
+        assert params['space_type'] == 'cosinesimil'
+        assert params['parameters'] == {'ef_construction': 128, 'm': 16}
```
vicilliar, line 170: do we want to assert that engine = 'lucene' as well, or should that be obvious?

vicilliar (approved):
Looks good, my only concern is we may want to test `params['space_type']` as well, if the user can set it

PR #424, opened 2023-04-04
Add maximum value validation for HNSW parameters

Jeadie:
# Changes
- Validate that HNSW parameters cannot be above set values.
  - For ef_construction, maximum can be set via environment variable `MARQO_EF_CONSTRUCTION_MAXIMUM_VALUE`.
  

 ## Testing
 - Unit tests
 - Local test
 - Upper limit testing

### Local Test
 ```
export MARQO_EF_CONSTRUCTION_MAXIMUM_VALUE="2"
```
Ensure the following fails
```
client.create_index(index_name, settings_dict={
    "index_defaults": {
        "ann_parameters": {
            "name": "hnsw",
            "space_type": "l2",
            "engine": "lucene",
            "parameters": {
              "ef_construction": 3,
              "m": 50 # acceptable value
            }
        },
    },
    "number_of_shards": 1
})
```

### Upper limit testing
 - Create index with `ef_construction`
 - Attempt to add 200 simple documents
 - MARQO_EF_CONSTRUCTION_MAXIMUM_VALUE =10,000,000 is a safe upper bound

### CODE REVIEW

File: src/marqo/tensor_search/enums.py [hunk 1]
```diff
@@ -101,7 +101,7 @@ class EnvVars:
     MARQO_ENABLE_THROTTLING = "MARQO_ENABLE_THROTTLING"
     MARQO_LOG_LEVEL = "MARQO_LOG_LEVEL"
     MARQO_ROOT_PATH = "MARQO_ROOT_PATH"
-
+    EF_CONSTRUCTION_MAXIMUM_VALUE = "MARQO_EF_CONSTRUCTION_MAXIMUM_VALUE"
```
pandu-k: Can we keep the name of the variable and the value identical (in line with the other env vars?). It would be good to keep the pattern unless there's a pressing reason to change it

pandu-k: Also we can use `MAX` rather than `MAXIMUM` for conciseness

pandu-k (changes requested):

src/marqo/tensor_search/enums.py
Jeadie: Oh sorry, I only saw the pattern on the RHS.

pandu-k (approved):

Issue #429, opened 2023-04-11
[ENHANCEMENT] Limit number of searchable tensor field


Jeadie:
## Feature
Limit the number of tensor fields that can be included in the searchable attributes, configurable by environment variables.

## Changes
- Environment variable `MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES` sets the maximum number of searchable attributes. 
  - `len(searchable_attributes) <= MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES`
- Restrict size of `searchable_attributes` for API endpoint `/indexes/{index_name}/search` when `searchMethod==TENSOR`
- Restrict size of `searchableAttributes` in each entry of body of `/indexes/bulk/search` for body entries with `searchMethod==TENSOR`

## Open Questions
- When `searchable_attributes` is unspecified, how do we select k from n attributes to search, when `n > MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES`

PR #430, opened 2023-04-12
Use env variable to configure how many attributes can be used for tensor search

Jeadie:


Jeadie:
Related issue: https://github.com/marqo-ai/marqo/issues/429

Jeadie:
Unit tests pass https://github.com/marqo-ai/marqo/actions/runs/4673313484

src/marqo/tensor_search/configs.py
```diff
@@ -55,5 +58,5 @@ def default_env_vars() -> dict:
         EnvVars.MARQO_ENABLE_THROTTLING: "TRUE",
         EnvVars.MARQO_LOG_LEVEL: "info",             # This env variable is set to "info" by default in run_marqo.sh, which overrides this value
         EnvVars.MARQO_EF_CONSTRUCTION_MAX_VALUE: 4096,
-    }
-
+        EnvVars.MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES: sys.maxsize
```
pandu-k: Can this be set to `None` by default? Because if it is None, then a query without `searchable_attributes` is legal.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1173,6 +1176,10 @@ def search(config: Config, index_name: str, text: Union[str, dict],
         args=(config, index_name, REFRESH_INTERVAL_SECONDS))
     cache_update_thread.start()
 
+    # Must do after cache refresh to get field names.
```
pandu-k: Doesn't this only check refresh the cache if the index is not found in the index cache (this is good because we don't want to slow down the search request with an index update request)?

src/marqo/tensor_search/validation.py
```diff
@@ -85,6 +90,21 @@ def validate_bulk_query_input(q: 'BulkSearchQueryEntity') -> Optional[MarqoError
...
+    
+    maximum_searchable_attributes = int(utils.read_env_vars_and_defaults(enums.EnvVars.MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES))
+    if searchable_attributes is not None:
+        num_searchable_attributes = len(searchable_attributes)
+    else:
+        num_searchable_attributes = len(field_names)
+
+    if num_searchable_attributes > maximum_searchable_attributes:
+        raise InvalidArgError(
+            f"Maximum searchable attributes for tensor search is {maximum_searchable_attributes}, received {num_searchable_attributes}."
```
pandu-k: thoughts on this idea? 

indicate what env var sets the maximum searchable attributes, so that users can know what configuration caused this

tests/tensor_search/test_bulk_search.py
```diff
@@ -59,6 +60,22 @@ def test_bulk_search_w_extra_parameters__raise_exception(self):
                 "parameter-not-expected": 1,
             }]))
 
+    @mock.patch.dict('os.environ', {**os.environ, **{'MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES': '0'}})
```
pandu-k: can we test with different cases of `MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES `? For example: 0, 1, 2, 10, None?

pandu-k: And cases where it fails/passes?

tests/tensor_search/test_search.py
pandu-k, line 1: can we test with different cases of MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES ? For example: 0, 1, 2, 10, None?

And cases where it fails/passes?

src/marqo/tensor_search/validation.py
```diff
@@ -85,6 +90,21 @@ def validate_bulk_query_input(q: 'BulkSearchQueryEntity') -> Optional[MarqoError
 
     return None
 
+def validate_searchable_attributes(field_names: List[str], searchable_attributes: Optional[List[str]], search_method: SearchMethod):
+    if search_method != SearchMethod.TENSOR:
+        return
+    
+    maximum_searchable_attributes = int(utils.read_env_vars_and_defaults(enums.EnvVars.MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES))
```
pandu-k: Raise error the max searchable attributes is specified, but none are given as part of the query

pandu-k (changes requested):

Jeadie:
Running unit tests: https://github.com/marqo-ai/marqo/actions/runs/4694802272

Jeadie:
bad merge fix in above unit tests, rerunning, https://github.com/marqo-ai/marqo/actions/runs/4694888269

Jeadie:
unit tests: https://github.com/marqo-ai/marqo/actions/runs/4696970755

pandu-k:
API tests: https://github.com/marqo-ai/marqo/actions/runs/4716830282

tests/tensor_search/test_search.py
```diff
@@ -54,6 +56,44 @@ def test_each_doc_returned_once(self):
...
+        with self.assertRaises(InvalidArgError):
+            tensor_search.add_documents(
+                config=self.config, index_name=self.index_name_1, docs=[
+                    {"abc": "Exact match hehehe", "other field": "baaadd", "_id": "5678"},
+                    {"abc": "random text", "other field": "Close match hehehe", "_id": "1234"},
+                ], auto_refresh=True)
+            tensor_search.search(
+                config=self.config, index_name=self.index_name_1, text="Exact match hehehe", return_doc_ids=True
+            )
+
```
pandu-k, line 109: How about a test with a really large `MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES`, and a low number of document fields. Then assert it should fail if it searchable attributes isn't set. 

This ensures we're testing the codepath where it should fail if searchable attributes isn't set

tests/tensor_search/test_search.py
pandu-k, line 1: Can we add tests for the valid cases? Where MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES is larger than searchable attributes?

tests/tensor_search/test_validation.py
```diff
@@ -301,6 +304,64 @@ def test_boost_validation_None_ok(self):
...
+            search_method=enums.SearchMethod.TENSOR
+        )
+
+    def test_searchable_attributes_is_none_max_value_set_raise_invalid_arg_error(self):
+        try:
+            validation.validate_searchable_attributes(
+                self.field_names,
+                searchable_attributes=None,
+                search_method=enums.SearchMethod.TENSOR
+            )
```
pandu-k, line 330: insert raising an assertion error here - this ensure that the test fails if the expected error is not raised

src/marqo/tensor_search/validation.py
```diff
@@ -85,6 +90,29 @@ def validate_bulk_query_input(q: 'BulkSearchQueryEntity') -> Optional[MarqoError
 
     return None
 
+def validate_searchable_attributes(field_names: List[str], searchable_attributes: Optional[List[str]], search_method: SearchMethod):
```
pandu-k: Adding a docstring would be good

pandu-k: devs would need context/explanation for this, because there's logic which means it's not a super-general function (e.g.: this function assumes MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES is set)

pandu-k: making the name more specific would also help (it's not a general searchable_attributes validation function)

tests/tensor_search/test_validation.py
pandu-k: also, consider `environ` here too, to make this test more deterministic

src/marqo/tensor_search/tensor_search.py
Jeadie: Don't need to do this anymore. Previously when we were going to use field_names when no searchable_attributes provided, we needed field names from the cache. I can remove this.

tests/tensor_search/test_search.py
Jeadie: removing `field_names` will solve this problem

tests/tensor_search/test_validation.py
Jeadie: Smart.

tests/tensor_search/test_bulk_search.py
Jeadie: We can, but I don't see the point.

Jeadie:
Just a timeout problem. Rerunning: https://github.com/marqo-ai/marqo/actions/runs/4718068319

tests/tensor_search/test_search.py
```diff
@@ -54,6 +56,77 @@ def test_each_doc_returned_once(self):
...
+                {"abc": "Exact match hehehe", "other field": "baaadd", "_id": "5678"},
+                {"abc": "random text", "other field": "Close match hehehe", "_id": "1234"},
+            ], auto_refresh=True)
+        tensor_search.search(
+            config=self.config, index_name=self.index_name_1, text="Exact match hehehe",
+            searchable_attributes=["other field"], return_doc_ids=True
+        )
+
+    @mock.patch.dict('os.environ', {**os.environ, **{'MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES': '0'}})
+    def test_search_with_excessive_searchable_attributes(self):
```
pandu-k: exact same name as other unit test above ^^ , please make it unique so it runs properly

tests/tensor_search/test_search.py
```diff
@@ -54,6 +56,77 @@ def test_each_doc_returned_once(self):
...
+                    {"abc": "Exact match hehehe", "other field": "baaadd", "_id": "5678"},
+                    {"abc": "random text", "other field": "Close match hehehe", "_id": "1234"},
+                ], auto_refresh=True)
+            tensor_search.search(
+                config=self.config, index_name=self.index_name_1, text="Exact match hehehe",
+                searchable_attributes=["other field"], return_doc_ids=True
+            )
+    
+    def test_search_with_searchable_attributes_max_attributes_is_none(self):
+        # Cannot set {'MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES': None} in mock.patch.dict
```
pandu-k: i managed to mock environ here: https://github.com/marqo-ai/marqo/blob/mainline/tests/tensor_search/test_add_documents.py#L1099

tests/tensor_search/test_search.py
pandu-k: what do you mean?

Jeadie: We were talking about different issues. Test `test_search_with_allowable_num_searchable_attributes` Should resolve this.

tests/tensor_search/test_search.py
Jeadie: Interesting. Mine produced a unittest error. perhaps either because
1. We're using it as a method decorator (outside even a unittest function)
2. `mock.patch.dict` vs `mock.patch`

tests/tensor_search/test_search.py
Jeadie: These are practically equivalent. Will remove.

tests/tensor_search/test_search.py
Jeadie: It is `mock.patch.dict`

tests/tensor_search/test_search.py
pandu-k: I think we still need this test. We want to ensure an error is raised if `MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES` is set but `searchable_attributes` aren't included in `search()`. 

We need to ensure that errors are raised due to `searchable_attributes` not being included in `search()` rather than the `len(searchable_attributes) > MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES`

In this particular test it is ambiguous why the error is raised, because the total number of implicit searchable attributes (3) is larger than `MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES` (2)

tests/tensor_search/test_bulk_search.py
pandu-k: it's a one-off cost (that can be minimised by copy-pasting) for stronger guarantees about how this function works indefinitely into the future

pandu-k (commented):
added single comments

tests/tensor_search/test_search.py
Jeadie: We have those tests in `TestValidateSearchableAttributes` but we can also just have them here.

Jeadie: `test_search_with_no_searchable_attributes_but_max_searchable_attributes_env_set` is that test? Or are you just wanting `MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES > len(fields)`?

pandu-k: Yep. Let's test both cases:
`MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES > len(fields)`
and 
`MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES < len(fields)`

pandu-k (approved):

pandu-k:
unit tests: https://github.com/marqo-ai/marqo/actions/runs/4737999867

PR #436, opened 2023-04-14
Deletion optimisation

pandu-k:
Optimisation

The delete documents by ID endpoint uses Marqo-os' delete by query. This can result in 5xxs being returned for deletion calls, after many calls. 

The delete documents by ID endpoint now gives Marqo-os' endpoint bulk delete instructions

No

In progress: https://github.com/marqo-ai/marqo/actions/runs/4727168886

* **Other information**:
To dos:
- Large scale test


Jeadie:
This is more than just changing how, in Opensearch, we perform delete operations. It's also an architectural change.

src/marqo/tensor_search/delete_docs.py
```diff
@@ -0,0 +1,96 @@
...
+        raise RuntimeError(f"Config set to use unknown backend `{config.backend}`. "
+                           f"See tensor_search.enums.SearchDB for allowed backends")
+
+    return format_delete_docs_response(del_response)
+
+
+# -- Marqo-OS-specific deletion implementation: --
+
+
+def delete_documents_marqo_os(config: Config, deletion_instruction: MqDeleteDocsRequest) -> MqDeleteDocsResponse:
```
codebrain, line 51: Be careful with unbounded size on `deletion_instruction.document_ids`. There are limits to the size of HTTP requests. This is usually not a problem, but if you are sending millions of operations this can add up quickly.

Some other feedback.

- Consider chunking requests into, say, 10,000 operations and then sending these requests in parallel
- You should:
  - potentially attempt retries, assuming request is retryable based on response HTTP code (e.g. `429`)
  - manage back-offs
  - send multiple requests in parallel for higher throughput
- See [here](https://www.elastic.co/blog/indexing-documents-with-the-nest-elasticsearch-net-client) lookup "BulkAllObservable<T> helper" heading for inspiration
- Inspect the bulk response(s) to determine they have been successful
- Consider [disabling/changing the refresh interval](https://www.elastic.co/guide/en/elasticsearch/reference/current/tune-for-indexing-speed.html#_unset_or_increase_the_refresh_interval) around your bulk request (ensure it is correctly set back!)

codebrain (commented):
No need for me to approve, but I have left some feedback based on issues I would consider addressing.

pandu-k:
unit tests: https://github.com/marqo-ai/marqo/actions/runs/4739613494

pandu-k:
unit tests (post merging mainline): https://github.com/marqo-ai/marqo/actions/runs/4739643546

src/marqo/tensor_search/api.py
```diff
@@ -236,7 +236,7 @@ def delete_index(index_name: str, marqo_config: config.Config = Depends(generate
 @app.post("/indexes/{index_name}/documents/delete-batch")
 def delete_docs(index_name: str, documentIds: List[str], refresh: bool = True,
                       marqo_config: config.Config = Depends(generate_config)):
-    return tensor_search.delete_documents(
```
Jeadie: Why the import change? (IIRC tensor_search.<OPERATION> is standard).

src/marqo/tensor_search/delete_docs.py
```diff
@@ -0,0 +1,77 @@
...
+This module handles the delete documents endpoint
+"""
+import datetime
+import json
+from marqo._httprequests import HttpRequests
+from marqo.config import Config
+from marqo.tensor_search import validation, utils, enums
+from marqo.tensor_search.models.delete_docs_objects import MqDeleteDocsResponse, MqDeleteDocsRequest
+
+# -- Marqo delete endpoint interface: --
```
Jeadie, line 11: Is this referring to anything in particular?

src/marqo/tensor_search/models/delete_docs_objects.py
```diff
@@ -0,0 +1,24 @@
...
+"""
+
+import datetime
+from typing import NamedTuple, Literal, List
+
+
+class MqDeleteDocsResponse(NamedTuple):
+    """An object that holds the data we send back to users"""
+    index_name: str
+    status_string: Literal["succeeded"]
```
Jeadie, line 13: This is because unsuccessful responses result in python exceptions, correct?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -2597,6 +2562,18 @@ def _create_score_modifiers_tensor_search_query(result_count, offset, vector_fie
     }
     return search_query
 
+
+def delete_documents(config: Config, index_name: str, doc_ids: List[str], auto_refresh):
+    """Delete documents from the Marqo index with the given doc_ids """
+    return delete_docs.delete_documents(
```
Jeadie, line 2568: When we go about an entire reformatting of tensor_search/, it may be worth having `format_delete_docs_response` in this function instead. Therefore operation specific files deal in Pydantic objects, and the tensor_search.py operations are responsibly for mapping args -> RequestObject, then ResponseObject -> Dict

Jeadie: Or have the decoding within the pydantic model (i.e. controlled via the `.json` method).

src/marqo/tensor_search/utils.py
```diff
@@ -175,20 +176,52 @@ def read_env_vars_and_defaults(var: str) -> Optional[str]:
     """Attempts to read an environment variable.
     If none is found, it will attempt to retrieve it from
     configs.default_env_vars(). If still unsuccessful, None is returned.
+    If it's an empty string, None is returned.
     """
-    try:
-        var = os.environ[var]
-        if var is not None and len(var) == 0:
+
+    def none_if_empty(value: Optional[str]) -> Optional[str]:
```
Jeadie, line 182: nit: function in a function is often an anti-pattern.

src/marqo/tensor_search/utils.py
```diff
@@ -175,20 +176,52 @@ def read_env_vars_and_defaults(var: str) -> Optional[str]:
     """Attempts to read an environment variable.
     If none is found, it will attempt to retrieve it from
     configs.default_env_vars(). If still unsuccessful, None is returned.
+    If it's an empty string, None is returned.
```
Jeadie, line 179: There is a risk here we can no longer distinguish between 
1. a value was not provided
2. An 'empty' value was provided.

I don't believe right now this is an issue, but it is often required.

src/marqo/tensor_search/validation.py
```diff
@@ -598,3 +599,39 @@ def validate_score_modifiers_object(score_modifiers: List[dict]):
             f"Please revise your score_modifiers based on the provided error."
             f"\n Check `https://docs.marqo.ai/0.0.17/API-Reference/search/#score-modifiers` for more info."
         )
+
+
+def validate_delete_docs_request(delete_request: MqDeleteDocsRequest, max_delete_docs_count: int):
```
Jeadie, line 604: I only now realise that MqDeleteDocsRequest is a NamedTuple, not a Pydantic BaseModel. I think we'd find that if we use a BaseModel, most of this is automatically validated at __init__ time.

Jeadie: And I think you can do both custom, and extended validation for something like `if (len(delete_request.document_ids) > max_delete_docs_count) and max_delete_docs_count is not None:`

tests/tensor_search/test_utils.py
```diff
@@ -165,6 +172,55 @@ def run():
...
+            @mock.patch("marqo.tensor_search.configs.default_env_vars", mock_default_env_vars)
+            @mock.patch("os.environ", mock_real_environ)
+            def run():
+                result = utils.read_env_vars_and_defaults_ints(var=key)
+                assert result == expected, f"Expected {expected}, got {result}"
+                return True
+
+            assert run()
+
+    def test_read_env_vars_and_defaults_ints_invalid_values(self):
```
Jeadie, line 204: (A comment for future improvements). There is a common pattern of using a mocked object across a range of parameters. It's something like a `@pytest.mark.parameterized` with a `@mock.patch("foo", return_value="bar)`. It could be worth figuring out how to do so. I think it'd make alot of our tests, a) more readable, b) easier to see for which values, what tests fail.

src/marqo/tensor_search/models/delete_docs_objects.py
pandu-k: So far. This is more of a placeholder that has been become a part of the API. There may be a future state where there could be other statuses

src/marqo/tensor_search/utils.py
pandu-k: Agreed

src/marqo/tensor_search/utils.py
pandu-k: Yep I agree this is an issue, but this is the functionality implemented originally - worth keeping in mind as future work.

src/marqo/tensor_search/api.py
pandu-k: done

pandu-k: (as in made it consistent)

pandu-k:
merged mainline back in. 
unit tests: https://github.com/marqo-ai/marqo/actions/runs/4749577191
api tests: https://github.com/marqo-ai/marqo/actions/runs/4749578141

Jeadie (approved):

Issue #417, opened 2023-04-02
[BUG] Number of shards recalculated for index settings


pandu-k:
**Describe the bug**
Number of shards are defined at index creation time and aren't updated

**Suggested fix**
Get the updated shard count when get index settings is requested

PR #437, opened 2023-04-14
Call `_cat/shards/{index_name}?format=json` to get number_of_shards

Jeadie:
Issue: #417

### DIFF
--- a/src/marqo/tensor_search/backend.py
+++ b/src/marqo/tensor_search/backend.py
@@ -16,15 +16,7 @@ from marqo.tensor_search.index_meta_cache import get_cache
 from marqo.tensor_search.index_meta_cache import get_index_info as get_cached_index_info
 import pprint
 
-def get_settings(index_name: str, marqo_config: Config):
-    """Get the settings for a specific index."""
-    shards = get_num_shards(config=marqo_config, index_name=index_name)
 
-    index_info = get_index_info(config=marqo_config, index_name=index_name)
-    index_info.index_settings["number_of_shards"] = shards
-
-    return index_info.index_settings
-    
 def get_num_shards(config: Config, index_name: str) -> int:
     """Returns the number of shards assigned to an index from Opensearch"""
     resp  = HttpRequests(config).get(path=F"_cat/shards/{index_name}?format=json")
--- a/src/marqo/tensor_search/backend.py
+++ b/src/marqo/tensor_search/backend.py
@@ -1,20 +1,16 @@
 """Communication with Marqo's persistence and search layer (OpenSearch)"""
 import copy
 import json
-import typing
-from marqo.tensor_search.models.index_info import IndexInfo
-# client-specific modules - we may want to replace these:
+from typing import Iterable, Tuple, Dict
+
+from marqo import errors
 from marqo._httprequests import HttpRequests
 from marqo.config import Config
-from marqo.errors import MarqoError
-from marqo.tensor_search import validation, constants, enums
-from marqo.tensor_search import utils
-from marqo import errors
-#
-from typing import Iterable, List, Union, Optional, Tuple, Dict
+from marqo.tensor_search import validation, constants, enums, utils
+from marqo.tensor_search.enums import IndexSettingsField
 from marqo.tensor_search.index_meta_cache import get_cache
 from marqo.tensor_search.index_meta_cache import get_index_info as get_cached_index_info
-import pprint
+from marqo.tensor_search.models.index_info import IndexInfo
 
 
 def get_index_info(config: Config, index_name: str) -> IndexInfo:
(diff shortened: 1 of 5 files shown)


src/marqo/tensor_search/backend.py
```diff
@@ -1,20 +1,16 @@
 """Communication with Marqo's persistence and search layer (OpenSearch)"""
 import copy
 import json
-import typing
```
Jeadie: This is just import cleanup

src/marqo/tensor_search/backend.py
```diff
@@ -31,28 +27,25 @@ def get_index_info(config: Config, index_name: str) -> IndexInfo:
         NonTensorIndexError, if the index's mapping doesn't conform to a Tensor Search index
 
     """
-    res = HttpRequests(config).get(path=F"{index_name}/_mapping")
```
Jeadie: This function was extremely verbose. Fundamentally I changed `[]` notation for implicit `.get("key", {})` for reduced boolean logic. i.e.

So instead of:
```
if not (index_name in res and "mappings" in res[index_name] and "_meta" in res[index_name]["mappings"])
```

It is easier to read
```
not (mapping.get(index_name, {}).get("mappings", {}).get("_meta")):
```

src/marqo/tensor_search/backend.py
```diff
@@ -31,28 +27,25 @@ def get_index_info(config: Config, index_name: str) -> IndexInfo:
...
+    shards  = HttpRequests(config).get(path=F"_cat/shards/{index_name}?format=json")
```
Jeadie: https://opensearch.org/docs/1.1/opensearch/rest-api/cat/cat-shards/

Jeadie: Response format
```json
[
    {
        "index": "sample-index1",
        "shard": "1",
        "prirep": "p",
        "state": "STARTED",
        "docs": "0",
        "store": "208b",
        "ip": "172.17.0.2",
        "node": "6891ffa8309d"
    },
    {
        "index": "sample-index1",
        "shard": "1",
        "prirep": "r",
        "state": "UNASSIGNED",
        "docs": null,
        "store": null,
        "ip": null,
        "node": null
    },
    {
        "index": "sample-index1",
        "shard": "0",
        "prirep": "p",
        "state": "STARTED",
        "docs": "0",
        "store": "208b",
        "ip": "172.17.0.2",
        "node": "6891ffa8309d"
    },
    {
        "index": "sample-index1",
        "shard": "0",
        "prirep": "r",
        "state": "UNASSIGNED",
        "docs": null,
        "store": null,
        "ip": null,
        "node": null
    }
]
```

src/marqo/tensor_search/backend.py
```diff
@@ -31,28 +27,25 @@ def get_index_info(config: Config, index_name: str) -> IndexInfo:
...
-    if "index_settings" in res[index_name]["mappings"]["_meta"]:
-        index_settings = res[index_name]["mappings"]["_meta"]["index_settings"]
-    else:
+    index_settings = mapping[index_name]["mappings"]["_meta"].get("index_settings")
+    if not index_settings:
         raise errors.NonTensorIndexError(
-            "get_index_info: couldn't identify index_settings "
-            F"in index mappings! Mapping: {res}")
+            f"get_index_info: couldn't identify index_settings in index mappings! Mapping: {mapping}")
+    index_settings[IndexSettingsField.number_of_shards] = num_shards
```
Jeadie: This is the only override on `index_settings`

tests/tensor_search/test_backend.py
```diff
@@ -46,6 +46,18 @@ def test_get_index_info_no_index(self):
         except IndexNotFoundError as s:
             pass
 
+    @mock.patch('marqo._httprequests.HttpRequests.get', side_effect=[
```
Jeadie, line 51: Essentially we mock the `GET _cat/shards/index1` endpoint to return 10 shards, instead of the 5 shards (`NsFields.number_of_shards: 5,`) returned by the default index settings.

Jeadie:
Unit tests: https://github.com/marqo-ai/marqo/actions/runs/4716409254

src/marqo/tensor_search/backend.py
pandu-k: do these have the same truth tables?

pandu-k: would be awesome if you could empirically verify

pandu-k (changes requested):
We need to ensure that only the index_settings call makes the extra call to `cat/shards`

Jeadie:
https://github.com/marqo-ai/marqo/actions/runs/4718077064

src/marqo/tensor_search/backend.py
```diff
@@ -16,6 +16,20 @@
 from marqo.tensor_search.index_meta_cache import get_index_info as get_cached_index_info
 import pprint
 
+def get_settings(index_name: str, marqo_config: Config):
```
pandu-k: This function makes more sense in the tensor_search.py module

pandu-k: Or in a completely separate index_settings_endpoint.py module

tests/tensor_search/test_backend.py
pandu-k, line 1: how about tests for weird responses from the cat shards endpoint? Does it result in a parsing error?

Jeadie: On a non-200 we will get a raise_for_status exception via the HTTPRequests. Are you concerned about network-malformed payloads? of version updates?

pandu-k: just any non expected response. Are we guaranteed that this API is 100% stable and will never mutate (perhaps in version updates, or in situations we haven't considered)

pandu-k: (or they have a bug)

Jeadie: We fix opensearch versions

pandu-k (approved):

pandu-k:
unit tests: https://github.com/marqo-ai/marqo/actions/runs/4738025142

PR #370, opened 2023-03-03
Revised add punkt tokenizer download into Dockerfile

vicilliar:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**
Change to improve timing on start-up.

**What is the current behavior? (You can also link to an open issue here)**
On startup, a series of initialisations are run [link](https://github.com/marqo-ai/marqo/blob/mainline/src/marqo/tensor_search/on_start_script.py#L14), including loading nltk tokenizer support [link](https://github.com/marqo-ai/marqo/blob/mainline/src/marqo/tensor_search/on_start_script.py#L107)

**What is the new behavior (if this is a feature change)?**
Downloading and storing the tokenizers/punkt data into the Docker image. On startup, outputs log message 
```
nltk data: tokenizers/punkt already downloaded
```
to confirm that tokenizers have already been downloaded during buildtime.

**Does this PR introduce a breaking change? (What changes might users need to make in their application due to this PR?)**
No

**Have unit tests been run against this PR? (Has there also been any additional testing?)**
No

**Related Python client changes (link commit/PR here)**
No

**Related documentation changes (link commit/PR here)**
No

**Other information:**
No

* **Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes/features)
- [x] Docs have been added / updated (for bug fixes / features)

PR #438, opened 2023-04-14
Fixed bug in test_use_existing_tensors_all_data_types

vicilliar:
Bug fix

Every 3-4 runs of unit tests in github actions, an assertion error occurs with `test_use_existing_tensors_all_data_types`

Errors no longer occur in the test (see runs unit_test_CI #365 to unit_test_CI #370)
- Test uses `assertAlmostEqual` instead of `assertEqual`
- Fixed error of multiple documents adding values of different datatypes to `Title` field

No

Yes

### DIFF
--- a/tests/tensor_search/test_add_documents_use_existing_tensors.py
+++ b/tests/tensor_search/test_add_documents_use_existing_tensors.py
@@ -53,10 +53,12 @@ class TestAddDocumentsUseExistingTensors(MarqoTestCase):
         r2 = tensor_search.add_documents(
             config=self.config, index_name=self.index_name_1, docs=[d1, d1],
             auto_refresh=True, use_existing_tensors=True)
+
         for item in r1['items']:
             assert item['result'] == 'created'
         for item in r2['items']:
             assert item['result'] == 'created'
+        
 
     def test_use_existing_tensors_non_existing(self):
         """check parity between a doc created with and without use_existing_tensors, then overwritten,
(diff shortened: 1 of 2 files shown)


tests/tensor_search/test_add_documents_use_existing_tensors.py
```diff
@@ -534,4 +539,4 @@ def test_use_existing_tensors_all_data_types(self):
                 config=self.config, index_name=self.index_name_1,
                 document_ids=[doc["_id" ]for doc in doc_arg], show_vectors=True)
             
-            self.assertEqual(d1, d2)
\ No newline at end of file
+            self.assertAlmostEqual(d1, d2)
```
wanliAlex: i think you are comparing the vectors here. do you want to use np.allclose and set a threshold?

wanliAlex (changes requested):

tests/tensor_search/test_add_documents_use_existing_tensors.py
vicilliar: Looking at other tests, I decided to change back to `assertEqual`, as I don't think that was the cause of the error anyway, it was the different data types being put in the same field.

We have had no problems directly comparing doc vector embeddings in other tests, as seen here:
test_non_tensor_field_list: https://github.com/marqo-ai/marqo/blob/mainline/tests/tensor_search/test_add_documents.py#L1127
test_use_existing_tensors_dupe_ids: https://github.com/marqo-ai/marqo/blob/mainline/tests/tensor_search/test_add_documents_use_existing_tensors.py#L143

vicilliar:
running 10 unit test pipelines to ensure this test does not fail in any

vicilliar:
all 10 runs passed
image

tests/tensor_search/test_add_documents_use_existing_tensors.py
wanliAlex: thanks for the in-depth description. approved

wanliAlex (approved):

PR #465, opened 2023-05-05
Replicas limit

wanliAlex:
we add a new environmental variable, "MARQO_MAX_NUMBER_OF_REPLICAS" to control the maximum number of replicas that can be set when creating an index.

no limit for number of replicas

we have this limit now

no

running

no

will do

* **Other information**:
no


src/marqo/tensor_search/models/settings_object.py
```diff
@@ -192,6 +192,7 @@
         NsFields.number_of_replicas: {
             "type": "integer",
             "minimum": 0,
+            "maximum": int(read_env_vars_and_defaults(EnvVars.MARQO_MAX_NUMBER_OF_REPLICAS)),
```
pandu-k: you can use this method to parse env vars into ints: https://github.com/marqo-ai/marqo/blob/mainline/src/marqo/tensor_search/utils.py#L202

It does things like give a useful error message if the value can't be parsed properly.

pandu-k (commented):
added comment

src/marqo/tensor_search/models/settings_object.py
```diff
@@ -128,7 +128,7 @@
                                 NsFields.hnsw_ef_construction: {
                                     "type": "integer",
                                     "minimum": 1,
-                                    "maximum": int(read_env_vars_and_defaults(EnvVars.MARQO_EF_CONSTRUCTION_MAX_VALUE)),
+                                    "maximum": read_env_vars_and_defaults_ints(EnvVars.MARQO_EF_CONSTRUCTION_MAX_VALUE),
```
vicilliar, line 131: may be the wrong PR to comment this on, but maybe we should rename `MARQO_EF_CONSTRUCTION_MAX_VALUE` to `MARQO_MAX_EF_CONSTRUCTION_VALUE`, because if you observe the other env vars, the word `MAX` always comes first. It may confuse users if we keep EF like that.

tests/tensor_search/test_create_index.py
```diff
@@ -324,24 +324,105 @@ def test_set_number_of_shards(self):
...
+        )
+        assert intended_replicas_count == int(resp.json()[self.index_name_1]['settings']['index']['number_of_replicas'])
+
+    def test_configurable_max_number_of_replicas(self):
+        maximum_number_of_replicas = 5
+        large_intended_replicas_count = 10
+        small_intended_replicas_count = 3
+        from marqo.tensor_search.models.settings_object import settings_schema
+
+        with patch.dict(settings_schema["properties"][NsField.number_of_replicas], maximum=maximum_number_of_replicas):
```
vicilliar, line 353: For the future (when configurable env vars are available in marqo API tests), it could be worth making this into an end-to-end API test

tests/tensor_search/test_create_index.py
```diff
@@ -324,24 +324,105 @@ def test_set_number_of_shards(self):
...
+            # a large value exceeding limits should not work
```
vicilliar, line 354: what about a test where input number exactly equals the maximum? (`NsField.number_of_replicas: 5`)

vicilliar: in this test case, there should be no `InvalidArgError`

vicilliar (approved):
looks good! I just mentioned 1 more possible test you could add, but it shouldn't be a big deal

vicilliar (approved):
looks good! I just mentioned 1 more possible test you could add, but it shouldn't be a big deal

src/marqo/tensor_search/models/settings_object.py
pandu-k: good thinking, to keep things consistent, but that would technically be a breaking change

PR #486, opened 2023-05-23
Basic doc changes

Jeadie:

* **Other information**:

### DIFF
--- a/src/marqo/tensor_search/models/add_docs_objects.py
+++ b/src/marqo/tensor_search/models/add_docs_objects.py
@@ -154,8 +154,7 @@ class AddDocsParams(BaseModel):
     index_name: str
     auto_refresh: bool
 
-    # this should only accept Sequences of dicts, but currently validation lies elsewhere
-    docs: List[Document]= Field(default_factory=list)
+    docs: List[Document] = Field(default_factory=list)
     
     non_tensor_fields: List = Field(default_factory=list)
     device: Optional[str] = None
--- a/src/marqo/tensor_search/models/add_docs_objects.py
+++ b/src/marqo/tensor_search/models/add_docs_objects.py
@@ -10,8 +10,11 @@ from marqo.tensor_search.models.private_models import ModelAuth
 
 
 class Weights(BaseModel):
+    """Represents a map between field names and an associated weight (e.g.multi-modal fields )."""
     class Config:
         extra = 'allow'
+        allow_mutation = False
+
 
     @root_validator(pre=True, allow_reuse=True)
     def check_keys(cls, values):
@@ -25,6 +28,10 @@ class Weights(BaseModel):
 
 
 class MappingObject(BaseModel):
+    """Field level control over a field. Currently only supports multi-modal combinations
+    
+    See: https://docs.marqo.ai/*/API-Reference/mappings/
+    """
     type: enums.MappingsObjectType
     weights: Weights = None  # Default value
 
@@ -36,6 +43,8 @@ class MappingObject(BaseModel):
 
     class Config:
         extra = 'forbid'
+        allow_mutation = False
+
 
     @validator('weights', always=True, allow_reuse=True)
     def check_weights(cls, weights, values, **kwargs):
@@ -44,12 +53,12 @@ class MappingObject(BaseModel):
         if is_multimodal and weights is None:
             raise errors.InvalidArgError(
                 f"Error validating multimodal combination object. Reason: 'weights' is not provided"
-                f"\nRead about the mappings object here: https://docs.marqo.ai/0.0.15/API-Reference/mappings/"
+                f" \nRead about the mappings object here: https://docs.marqo.ai/0.0.15/API-Reference/mappings/"
             )
         elif not is_multimodal and weights is not None:
             raise errors.InvalidArgError(
                 f"Error validating mappings. Mapping is not multimodal combination object. Reason: field is not multimodal, but weights provided."
-                f"\nRead about the mappings object here: https://docs.marqo.ai/0.0.15/API-Reference/mappings/"
+                f" \nRead about the mappings object here: https://docs.marqo.ai/0.0.15/API-Reference/mappings/"
             )
         return weights
 
@@ -59,25 +68,22 @@ class Document(BaseModel):
     class Config:
         extra = 'allow'
 
-    # @validator('*', pre=True, allow_reuse=True)
     def validate_field_names(self, v, values, field):
         if field.name == "_id":
             return
 
         validation.validate_field_name(field)
 
-    # @root_validator(allow_reuse=True)
     def validate_id(cls, values):
         if "_id" in values.keys():
             validation.validate_id(values.get("_id"))
 
         return values
 
-    # @root_validator(allow_reuse=True)
     def validate_document_size(cls, values):
         """Ensure document is less than MARQO_MAX_DOC_BYTES bytes (when set)."""
 
-        max_doc_size = utils.read_env_vars_and_defaults(var=enums.EnvVars.MARQO_MAX_DOC_BYTES)
+        max_doc_size = utils.read_env_vars_and_defaults_ints(var=enums.EnvVars.MARQO_MAX_DOC_BYTES)
         if max_doc_size is None:    
             return
 
@@ -115,19 +121,18 @@ class Document(BaseModel):
         """Gets the '_id' or generates AND sets to self, a UUID."""
         if self._id is not None:
             return self._id
-        
         self._id = id_gen()
+
         validation.validate_id(self._id)
         return self._id
     
     def get_chunk_values_for_filtering(self) -> Dict[str, Any]:
         """
         Metadata can be calculated here at the doc level. Only add chunk 
-        values which are string, boolean, numeric or dictionary. Dictionary
-        keys will be store in a list.
+        values which are string, boolean, numeric or dictionary.
         """
         return dict([
-            (k, v)for k, v in self.__dict__.items() if k != "_id" and isinstance(v, (str, float, bool, int, list, dict))
+            (k, v) for k, v in self.__dict__.items() if k != "_id" and isinstance(v, (str, float, bool, int, list, dict))
         ])
 
 
@@ -167,6 +172,13 @@ class AddDocsParams(BaseModel):
 
     class Config:
         arbitrary_types_allowed = True
+        allow_mutation = False
+
+    def __init__(self, **data):
+        try:
+            super().__init__(**data)
+        except ValidationError as e:
+            raise errors.InvalidArgError(message=e.json())
 
     @validator('docs', always=True)
     def docs_isnt_empty(cls, v, values, field):
@@ -216,5 +228,18 @@ class AddDocsParams(BaseModel):
         return doc_ids[::-1]
     
     def create_anew(self, **new_kwargs: Dict[str, Any]) -> "AddDocsParams":
+        """
+        Creates a new instance of AddDocsParams with kwargs overriding those in this
+        AddDocsParams object. Validation is performed on the new object with its kwarg overrides.
+
+        Args:
+            **new_kwargs: Any parameters to replace in the new instance.
+
+        Returns:
+            A new instance of AddDocsParams with the merged parameters.
+        Raises:
+            InvalidArgError: If AddDocsParams validation fails.
+            BadRequestError: If the new AddDocsParams has no documents.
+        """
         new_dict = {**self.dict(), **new_kwargs}
         return AddDocsParams(**new_dict)
--- a/src/marqo/tensor_search/models/add_docs_objects.py
+++ b/src/marqo/tensor_search/models/add_docs_objects.py
@@ -1,17 +1,130 @@
-from pydantic.dataclasses import dataclass
-from pydantic import Field
-from typing import Optional, Union, Any, Sequence
+import json
+from typing import Any, Callable, Dict, List, Literal, Optional, Sequence, Union
+import uuid
+
 import numpy as np
+from pydantic.dataclasses import dataclass
+from pydantic import Field, validator, BaseModel, root_validator
+
+from marqo.tensor_search import constants, enums, validation, utils
+import marqo.errors as errors
 from marqo.tensor_search.models.private_models import ModelAuth
-from typing import List
 
 
-class AddDocsParamsConfig:
-    arbitrary_types_allowed = True
+class Weights(BaseModel):
+    __root__: Dict[str, float]  # Allowing only string keys and number values
+
+    @root_validator(pre=True, allow_reuse=True)
+    def check_keys(cls, values):
+        if not all(isinstance(i, str) for i in values.keys()):
+            raise ValueError('All keys must be strings')
+        if not all(isinstance(i, (int, float)) for i in values.values()):
+            raise ValueError('All values must be numbers')
+        return values
 
 
-@dataclass(frozen=True, config=AddDocsParamsConfig)
-class AddDocsParams:
+class MappingObject(BaseModel):
+    type: enums.MappingsObjectType
+    weights: Weights = None  # Default value
+
+    class Config:
+        extra = 'allow'
+
+    @validator('weights', always=True, allow_reuse=True)
+    def check_weights(cls, weights, values, **kwargs):
+        """Ensure that weights is provided if and only if the type is 'MappingsObjectType.multimodal_combination'."""
+        is_multimodal = values['type'] == enums.MappingsObjectType.multimodal_combination
+        if is_multimodal and weights is not None:
+            raise errors.InvalidArgError(
+                f"Error validating multimodal combination object. Reason: 'weights' is not provided"
+                f"\nRead about the mappings object here: https://docs.marqo.ai/0.0.15/API-Reference/mappings/"
+            )
+        elif not is_multimodal and weights is None:
+            raise errors.InvalidArgError(
+                f"Error validating mappings. Mapping is not multimodal combination object. Reason: field is not multimodal, but weights provided."
+                f"\nRead about the mappings object here: https://docs.marqo.ai/0.0.15/API-Reference/mappings/"
+            )
+        return weights
+
+class Document(BaseModel):
+    _id: Optional[str] = None
+
+    class Config:
+        extra = 'allow'
+
+    # @validator('*', pre=True, allow_reuse=True)
+    def validate_field_names(self, v, values, field):
+        if field.name == "_id":
+            return
+
+        validation.validate_field_name(field)
+
+    # @root_validator(allow_reuse=True)
+    def validate_id(cls, values):
+        if "_id" in values.keys():
+            validation.validate_id(values.get("_id"))
+
+        return values
+
+    # @root_validator(allow_reuse=True)
+    def validate_document_size(cls, values):
+        """Ensure document is less than MARQO_MAX_DOC_BYTES bytes (when set)."""
+
+        max_doc_size = utils.read_env_vars_and_defaults(var=enums.EnvVars.MARQO_MAX_DOC_BYTES)
+        if max_doc_size is None:    
+            return
+
+        try:
+            serialized = json.dumps(values)
+        except TypeError as e:
+            raise errors.InvalidArgError(
+                f"Unable to index document: it is not serializable! Document: `{values}` "
+            )
+
+        if len(serialized) > int(max_doc_size):
+            maybe_id = f" _id:`{values['_id']}`" if '_id' in values else ''
+            raise errors.DocTooLargeError(
+                f"Document{maybe_id} with length `{len(serialized)}` exceeds "
+                f"the allowed document size limit of [{max_doc_size}]."
+            )
+    
+        return values
+
+    def _validate_field_name(self, key: str):
+        if key == "_id":
+            return 
+        
+        validation.validate_field_name(key)
+
+    def validate_document(self) -> None:
+        """
+        Validate the document object.
+        """
+        self.validate_document_size(self.dict())
+        self.validate_id(self.dict())
+        [self._validate_field_name(k) for (k, _) in self.dict().items()]
+
+    def get_or_generate_id(self, id_gen: Callable[[None], str] = lambda: str(uuid.uuid4())) -> str:
+        """Gets the '_id' or generates AND sets to self, a UUID."""
+        if self._id is not None:
+            return self._id
+        
+        self._id = id_gen()
+        validation.validate_id(self._id)
+        return self._id
+    
+    def get_chunk_values_for_filtering(self) -> Dict[str, Any]:
+        """
+        Metadata can be calculated here at the doc level. Only add chunk 
+        values which are string, boolean, numeric or dictionary. Dictionary
+        keys will be store in a list.
+        """
+        return dict([
+            (k, v)for k, v in self.__dict__.items() if k != "_id" and isinstance(v, (str, float, bool, int, list, dict))
+        ])
+
+
+class AddDocsParams(BaseModel):
     """Represents the parameters of the tensor_search.add_documents() function
 
     Params:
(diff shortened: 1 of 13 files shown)


src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -31,16 +151,70 @@ class AddDocsParams:
 
     """
 
-    # this should only accept Sequences of dicts, but currently validation lies elsewhere
-    docs: Union[Sequence[Union[dict, Any]], np.ndarray]
-
     index_name: str
     auto_refresh: bool
+
+    docs: List[Document] = Field(default_factory=list)
```
pandu-k, line 162: My preference is to separate the documents from the parameters as part of a separate-data-from-config principle. Does this change your design?

src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -1,17 +1,137 @@
...
+    class Config:
+        extra = 'allow'
+
+    @root_validator(pre=True, allow_reuse=True)
+    def check_keys(cls, values):
+        """Ensure all keys map to floats"""
+        if not all(isinstance(i, str) for i in values.keys()):
+            raise ValueError('All keys must be strings')
+        if not all(isinstance(i, (int, float)) and not isinstance(i, bool) for i in values.values()):
+            # Note: `isinstance(True, (int, float)) == True`
```
pandu-k, line 25: Thx for note!

src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -1,17 +1,137 @@
...
+
+from marqo.tensor_search import enums, validation, utils
+import marqo.errors as errors
 from marqo.tensor_search.models.private_models import ModelAuth
-from typing import List
 
 
-class AddDocsParamsConfig:
-    arbitrary_types_allowed = True
+class Weights(BaseModel):
```
pandu-k, line 12: Can you please add a docstring that adds context? For example, this is for defined for weights specifically to be used within the mappings for multimodal combination docs. You can also include a link to the docs: https://docs.marqo.ai/0.0.21/API-Reference/mappings/

src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -1,17 +1,137 @@
...
+                f"\nRead about the mappings object here: https://docs.marqo.ai/0.0.15/API-Reference/mappings/"
+            )
+        elif not is_multimodal and weights is not None:
+            raise errors.InvalidArgError(
+                f"Error validating mappings. Mapping is not multimodal combination object. Reason: field is not multimodal, but weights provided."
+                f"\nRead about the mappings object here: https://docs.marqo.ai/0.0.15/API-Reference/mappings/"
+            )
+        return weights
+
+class Document(BaseModel):
```
pandu-k, line 65: Consider logically separating this from the config-oriented objects. For example, by putting this at the top of the file and then having a section dedicated for config (`# ------ ADD DOCUMENTS CONFIG OBJECTS: ------`)

src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -1,17 +1,137 @@
...
+        extra = 'forbid'
+
+    @validator('weights', always=True, allow_reuse=True)
+    def check_weights(cls, weights, values, **kwargs):
+        """Ensure that weights is provided if and only if the type is 'MappingsObjectType.multimodal_combination'."""
+        is_multimodal = values.get('type', None) == enums.MappingsObjectType.multimodal_combination
+        if is_multimodal and weights is None:
+            raise errors.InvalidArgError(
+                f"Error validating multimodal combination object. Reason: 'weights' is not provided"
+                f"\nRead about the mappings object here: https://docs.marqo.ai/0.0.15/API-Reference/mappings/"
```
pandu-k: nit: consider adding a space before newline in these error messages. This makes it look better in when reading error messages in cURL:

`is not provided.\nRead about` vs `is not provided. \nRead about`

src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -1,17 +1,137 @@
...
+
+    class Config:
+        extra = 'forbid'
+
+    @validator('weights', always=True, allow_reuse=True)
+    def check_weights(cls, weights, values, **kwargs):
+        """Ensure that weights is provided if and only if the type is 'MappingsObjectType.multimodal_combination'."""
+        is_multimodal = values.get('type', None) == enums.MappingsObjectType.multimodal_combination
+        if is_multimodal and weights is None:
+            raise errors.InvalidArgError(
```
pandu-k, line 54: Do these properly get raised as InvalidArgErrors? Or are they caught and raised as ValidationErrors?

src/marqo/tensor_search/models/add_docs_objects.py
pandu-k, line 1: I prefer if we make objects immutable if we can, unless there is a specific reason for it to be mutable. Thoughts?

src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -1,17 +1,137 @@
...
+            )
+        return weights
+
+class Document(BaseModel):
+    _id: Optional[str] = None
+
+    class Config:
+        extra = 'allow'
+
+    # @validator('*', pre=True, allow_reuse=True)
```
pandu-k: I am guessing this is still WIP?

src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -1,17 +1,137 @@
...
+
+        return values
 
+    # @root_validator(allow_reuse=True)
+    def validate_document_size(cls, values):
+        """Ensure document is less than MARQO_MAX_DOC_BYTES bytes (when set)."""
 
-@dataclass(frozen=True, config=AddDocsParamsConfig)
-class AddDocsParams:
+        max_doc_size = utils.read_env_vars_and_defaults(var=enums.EnvVars.MARQO_MAX_DOC_BYTES)
```
pandu-k: You can use the following, which handles if an env var isn't coercable into an int: https://github.com/marqo-ai/marqo/blob/mainline/src/marqo/tensor_search/utils.py#L202

src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -1,17 +1,137 @@
...
+
+    def validate_document(self) -> None:
+        """
+        Validate the document object.
+        """
+        self.validate_document_size(self.dict())
+        self.validate_id(self.dict())
+        [self._validate_field_name(k) for (k, _) in self.dict().items()]
+
+    def get_or_generate_id(self, id_gen: Callable[[None], str] = lambda: str(uuid.uuid4())) -> str:
```
pandu-k, line 120: Document ID generation is an important piece of functionality. I would prefer this to have its own function, perhaps defined in the `add_docs.py` class, with some unit tests around it, to ensure stability between releases.

src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -1,17 +1,137 @@
...
+        
+        self._id = id_gen()
+        validation.validate_id(self._id)
+        return self._id
+    
+    def get_chunk_values_for_filtering(self) -> Dict[str, Any]:
+        """
+        Metadata can be calculated here at the doc level. Only add chunk 
+        values which are string, boolean, numeric or dictionary. Dictionary
+        keys will be store in a list.
```
pandu-k: Can you double check whether keys are indeed stored in a list? I thought dictionaries were stored as JSON objects still.

src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -31,16 +151,70 @@ class AddDocsParams:
...
+    mappings: Optional[Dict[str, MappingObject]] = None
     model_auth: Optional[ModelAuth] = None
+
+    class Config:
+        arbitrary_types_allowed = True
+
+    @validator('docs', always=True)
+    def docs_isnt_empty(cls, v, values, field):
+        if len(v) == 0:
+            raise errors.BadRequestError(message="Received empty add documents request")
```
pandu-k, line 186: Again, I have a question around whether the BadRequestError gets caught and translated into a ValidationError, so that it can be caught and added as an error for the response JSON

src/marqo/tensor_search/models/add_docs_objects.py
pandu-k: Do we call this anywhere? and if so, can we guarantee that an ID has been added? 

Not sure the best way to do this. Perhaps a subclass `DocumentWithID` (to differentiate it from the user-submitted doc), that generates an ID on initialisation and also validates the ID post initialisation.

src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -31,16 +151,70 @@ class AddDocsParams:
...
+        if self.update_mode == "replace":
+            indexing_instructions["index"]["_id"] = doc_id
+        else:
+            indexing_instructions["update"]["_id"] = doc_id
+        return indexing_instructions
+    
+    def get_doc_ids_uniquely(self) -> List[str]:
+        """Get unique, non-None, document ids."""
+        doc_ids = []
+        seen_ids = [] # Cannot use set in case doc._id is not a valid ID and is not hashable
```
pandu-k, line 220: Confused about the need for two lists and why it needs to be iterated over in reverse order. All good if WIP

src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -31,16 +151,70 @@ class AddDocsParams:
...
+        for doc in reversed(self.docs):
+            doc_id = doc._id
+            if doc_id and doc_id not in seen_ids:
+                seen_ids.append(doc_id)
+                doc_ids.append(doc_id)
+
+        # As we appended the docs in reversed order, reverse it again to maintain the original order
+        return doc_ids[::-1]
+    
+    def create_anew(self, **new_kwargs: Dict[str, Any]) -> "AddDocsParams":
```
pandu-k, line 230: It's worth adding a docstring here about behaviour. 

We should also create unit tests around this.

src/marqo/tensor_search/parallel.py
```diff
@@ -207,8 +207,8 @@ def add_documents_mp(
         IndexChunk(
             config=config,  batch_size=batch_size,
             process_id=p_id, threads_per_process=threads_per_process,
-            add_docs_params=replace(add_docs_params, docs=_docs, device=device_ids[p_id]))
-        for p_id,_docs in enumerate(np.array_split(add_docs_params.docs, n_processes))]
+            add_docs_params=add_docs_params.create_anew(docs=list(_docs), device=device_ids[p_id])
+        ) for p_id, _docs in enumerate(np.array_split(add_docs_params.docs, n_processes))]
```
pandu-k, line 211: Just curious about the difference of this change?

pandu-k: Was it to have one "official" way to do this type of operation?

src/marqo/tensor_search/models/add_docs_objects.py
pandu-k: The comment about keys being stored in a list may be outdated

src/marqo/tensor_search/tensor_search.py
```diff
@@ -738,41 +594,138 @@ def merged_doc = [:];
...
+    if add_docs_params.update_mode != "replace" or not add_docs_params.use_existing_tensors:
+        return None
+
+    matching_doc = [doc for doc in existing_docs if doc["_id"] == doc_id and doc.get("found", False)]
+    if len(matching_doc) == 1:
+        return matching_doc[0]
+    elif len(matching_doc) == 0:
+        return None
+    else:
+        raise errors.InternalError(message= f"Upsert: found {len(matching_doc)} matching docs for {doc_id} when only 1 or 0 should have been found.")
```
pandu-k: Can we change `"Upsert: .."` to `"Use existing tensors..."`

src/marqo/tensor_search/tensor_search.py
```diff
@@ -738,41 +594,138 @@ def merged_doc = [:];
...
+        raise errors.BadRequestError(
+            message=f'Problem vectorising query. Reason: {str(model_error)}',
+            link="https://marqo.pages.dev/latest/Models-Reference/dense_retrieval/"
+        )
+    except s2_inference_errors.S2InferenceError:
+        raise errors.InvalidArgError(message=f'Could not process given image: {field_content}')
+
+    if (len(vector_chunks) != len(text_chunks)):
+        raise RuntimeError(
+            f"the input content after preprocessing and its vectorized counterparts must be the same length."
```
pandu-k, line 674: nit: space after full stop in error msg

src/marqo/tensor_search/tensor_search.py
```diff
@@ -738,41 +594,138 @@ def merged_doc = [:];
...
+    if unsuccessful_docs:
+        result_dict['errors'] = True
+
+    for loc, error_info in unsuccessful_docs:
+        new_items.insert(loc, error_info)
+
+    result_dict["processingTimeMs"] = time_diff * 1000
+    result_dict["index_name"] = add_docs_params.index_name
+    result_dict["items"] = new_items
+    return result_dict
```
pandu-k, line 728: Consider making a Marqo Add Docs Response object

tests/tensor_search/test_add_documents.py
```diff
@@ -614,26 +612,6 @@ def test_add_documents_resilient_doc_validation(self):
...
-            ([{"this is a set, lmao"}, "this is a string", {"some_dict": "yep"}], [(None, 'error'), (None, 'error'), (None, 'result')]),
-            ([1234], [(None, 'error')]), ([None], [(None, 'error')]),
-            # handle invalid field names
-            ([{123: "bad"}, {"_id": "cool"}], [(None, 'error'), ("cool", 'result')]),
-            ([{"__chunks": "bad"}, {"_id": "1511", "__vector_a": "some content"}, {"_id": "cool"},
-              {"_id": "144451", "__field_content": "some content"}],
-             [(None, 'error'), ("1511", 'error'), ("cool", 'result'), ("144451", "error")]),
-            ([{123: "bad", "_id": "12345"}, {"_id": "cool"}], [("12345", 'error'), ("cool", 'result')]),
-            ([{None: "bad", "_id": "12345"}, {"_id": "cool"}], [("12345", 'error'), ("cool", 'result')]),
-            # handle bad content
```
pandu-k, line 636: why are these cases removed?

tests/tensor_search/test_add_documents.py
```diff
@@ -686,11 +660,11 @@ def test_mappings_arent_updated(self):
              ({"my_field"}, {})
              ),
             # invalid fields
-            ([{"_id": "14g", (12, 14): "some content"}, {"_id": "1511", None: "some content"},
+            ([{"_id": "14g"}, {"_id": "1511"},
               {"_id": "1511", "__vector_a": "some content"}, {"_id": "1234f", "__chunks": "some content"},
               {"_id": "144451", "__field_content": "some content"},
               {"_id": "sv4124", "good_field_3": "some content 2 " , "good_field_4": 3.65}],
-             ({"good_field_3", "good_field_4"}, {(12, 14), None, "__vector_a", "__chunks", "__field_content"})
```
pandu-k, line 693: Why are these cases removed?

tests/tensor_search/test_validation.py
```diff
@@ -764,36 +738,16 @@ def test_validate_mappings_invalid(self):
...
-                }
-            },
-            {
-                "my_combination_field": { # this dict is OK
-                    "type": "multimodal_combination",
-                    "weights": {
-                        "some_text": 0.5
-                    }
-                },
-                "other_field": {
```
pandu-k, line 781: This is removed because each mapping is validated individually?

src/marqo/tensor_search/models/add_docs_objects.py
Jeadie: InvalidArgErrors, https://github.com/marqo-ai/marqo/pull/486/files#diff-d9198de14a706fb00e7bd7e2ef56ebf00e0cab33e7de688690f3ddbd57ba33c9R748

src/marqo/tensor_search/models/add_docs_objects.py
Jeadie: We currently throw BadRequestError, https://github.com/marqo-ai/marqo/blob/ba2e99aacfc47b4c180e8d12cc94fbf01b2cbc4a/src/marqo/tensor_search/tensor_search.py#L376

src/marqo/tensor_search/models/add_docs_objects.py
Jeadie: This is the equivalent of a set inclusion operation. We can use a python set because _id may not be hashable.

src/marqo/tensor_search/models/add_docs_objects.py
Jeadie: We can add tests to it. I don't really see why it needs a separate function and file. The Id is a product of a document and should be associated with the data structure. It might be worth distinguishing between Document and DocumentWithID but in the current flow because they come in interchangeably, you'd have a need to split `List[Union[Document, DocumentWithID]]`

Jeadie: Worth adding some tests too

src/marqo/tensor_search/models/add_docs_objects.py
Jeadie: Good find.

src/marqo/tensor_search/models/add_docs_objects.py
Jeadie: I'd prefer to parse the HTTP request's Dict/Json to a predefined structure immediately.

src/marqo/tensor_search/parallel.py
Jeadie: Made AddDocsParams a pydantic model. replace is no longer available.

Jeadie: But also, it does help centralise these changes. For instance, now validation is done on the new AddDocsParam object.

src/marqo/tensor_search/tensor_search.py
Jeadie: Agree, but not here.

tests/tensor_search/test_add_documents.py
Jeadie: Same as above

tests/tensor_search/test_add_documents.py
Jeadie: Sorry, it doesn't seem I added some updates to the PR description. Because we validate immediately after the HTTP request comes in, there are some validations (generally just Typing validation) that cannot be caught and then returned as an error. These are those tests. May be a problem with this change.

tests/tensor_search/test_validation.py
Jeadie: Because we have refactored into a MappingObject, and then `mappings: Optional[Dict[str, MappingObject]]`, the first test no longer makes sense (in essence it not longer gets validated in MappingObject, it gets validated in AddDocsParams

Jeadie: The second case is similar. In essence, we are testing two separate cases the first (my_combination_field) passes, and the second (other_field) fails but is already a test case.

src/marqo/tensor_search/models/add_docs_objects.py
Jeadie: I should remove. Essentially we don't currently run this validation at DOcument creation (in the current marqo). So we cannot use a @validator. Instead we use these methods in `validate_document`

Jeadie:
https://github.com/marqo-ai/marqo/actions/runs/5149896919

src/marqo/tensor_search/models/add_docs_objects.py
Jeadie: We need Document above AddDocsParam for typing

src/marqo/tensor_search/models/add_docs_objects.py
Jeadie: Added explicit unit tests for this

src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -25,6 +28,10 @@ def check_keys(cls, values):
 
 
 class MappingObject(BaseModel):
+    """Field level control over a field. Currently only supports multi-modal combinations
+    
+    See: https://docs.marqo.ai/*/API-Reference/mappings/
```
pandu-k, line 33: That link doesn't work. Use `https://docs.marqo.ai/latest/API-Reference/mappings/` for this purpose. For better future proofing, link it to a specific version: `https://docs.marqo.ai/0.0.21/API-Reference/mappings/`

src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -216,5 +228,18 @@ def get_doc_ids_uniquely(self) -> List[str]:
         return doc_ids[::-1]
     
     def create_anew(self, **new_kwargs: Dict[str, Any]) -> "AddDocsParams":
+        """
+        Creates a new instance of AddDocsParams with kwargs overriding those in this
+        AddDocsParams object. Validation is performed on the new object with its kwarg overrides.
```
pandu-k, line 233: Great docstring!

src/marqo/tensor_search/models/add_docs_objects.py
Jeadie: I didn't think that, but it also means the docs are instantly out of date. At least with this is it's a regex of a link (we should also get `latest` to work).

src/marqo/tensor_search/models/add_docs_objects.py
pandu-k: It doesn't get parsed immediately, because requests can come in with batch_size and processes

src/marqo/tensor_search/models/add_docs_objects.py
pandu-k: Yep, that requirement shouldn't conflict with my comment

src/marqo/tensor_search/models/add_docs_objects.py
pandu-k: Test it out. It works with `latest` but not with `*`. I think it's best if users can copy-paste the link directly into their browser. 
 
Latest vs fixed version is a tradeoff with between these two scenarios:

1. We use `latest`. If we refactor our docs, this link will be dead
2. We use a specific version. This may seem out of date, but link will always work. There will also be a "you are looking at an outdated version of the docs message"

`latest` should be fine as it's just for developers' reference

src/marqo/tensor_search/models/add_docs_objects.py
pandu-k: Ok we can leave a the separate function for the future. The tests around this are good start. We already have a dedicated add documents file, by the way: `add_docs.py`

src/marqo/tensor_search/models/add_docs_objects.py
pandu-k: Sorry, I meant I wanted to make sure a BadRequestError is returned to the user, rather than a ValidationError (which would be a breaking change)

tests/tensor_search/test_add_documents.py
pandu-k: Is there any way around this? Like doing pydantic validation on each doc individually? and then adding the relevant "doc failed to be indexed" message to the response?

src/marqo/tensor_search/models/add_docs_objects.py
Jeadie: hmm, `latest` did not work originally, but does now.

farshidz:
Closing as stale

PR #508, opened 2023-06-13
Consolidating default device to CUDA when available

vicilliar:
Feature

Many different device defaults exist in different functions. 
- Config.py has search_device and indexing_device which are sometimes used
- get_default_device is also used for some functions
- CLIP/ONNX models have their own default device parameter
- vectorise has its own default device parameter
- etc

All device defaults are completely removed except 3:
- tensor_search/bulk_search
- tensor_search/search
- add_documents_orchestrator

No. All the changes are internal. But Py-marqo will have a breaking change.

In progress

In progress

In progress

* **Other information**:
TODO:
1. fixing some add_documents tests
2. Adding new unit tests for this feature
3. Adding py-marqo and API tests for this feature


src/marqo/tensor_search/tensor_search.py
```diff
@@ -239,6 +239,10 @@ def add_documents_orchestrator(
         config: Config, add_docs_params: AddDocsParams,
         batch_size: int = 0, processes: int = 1,
     ):
+    
+    # Default device calculated here and not in add_documents call
+    if add_docs_params.device is None:
+        add_docs_params.device = utils.read_env_vars_and_defaults("_MARQO_BEST_AVAILABLE_DEVICE")
```
wanliAlex: Why do we have an underline for this environment variable?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -249,13 +253,14 @@ def add_documents_orchestrator(
         _check_and_create_index_if_not_exist(config=config, index_name=add_docs_params.index_name)
 
         try:
+            # using search instead of _vector_text_search since search calculates default device
```
wanliAlex: The comment seems inconsistent with the code. you still use _vector_text_search here.

src/marqo/tensor_search/tensor_search.py
wanliAlex: actually one more thing, we should put this default assignment in pydantic @Jeadie  is my understanding correct?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -366,6 +374,9 @@ def add_documents(config: Config, add_docs_params: AddDocsParams):
     t0 = timer()
     bulk_parent_dicts = []
 
+    if not add_docs_params.device:
+        raise errors.InvalidArgError(message=f"add_documents (internal function) cannot be called without setting device!")
+
```
wanliAlex: similarly, this check should be done in pydantic

src/marqo/tensor_search/tensor_search.py
```diff
@@ -924,7 +933,7 @@ def refresh_index(config: Config, index_name: str):
 
 
 @add_timing
-def bulk_search(query: BulkSearchQuery, marqo_config: config.Config, verbose: bool = True, device=None):
+def bulk_search(query: BulkSearchQuery, marqo_config: config.Config, verbose: bool = True, device: str = None):
     """Performs a set of search operations in parallel.
```
wanliAlex, line 955: Optional[str] = None

src/marqo/tensor_search/tensor_search.py
```diff
@@ -950,7 +959,8 @@ def bulk_search(query: BulkSearchQuery, marqo_config: config.Config, verbose: bo
     if len(query.queries) == 0:
         return {"result": []}
 
-    selected_device = marqo_config.indexing_device if device is None else device
+    selected_device = utils.read_env_vars_and_defaults("_MARQO_BEST_AVAILABLE_DEVICE") \
+        if device is None else device
```
wanliAlex: same comment, put it in pydantic.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1101,11 +1112,14 @@ def search(config: Config, index_name: str, text: Union[str, dict],
         args=(config, index_name, REFRESH_INTERVAL_SECONDS))
     cache_update_thread.start()
 
+    selected_device = utils.read_env_vars_and_defaults("_MARQO_BEST_AVAILABLE_DEVICE") \
+        if device is None else device
+
```
wanliAlex, line 1148: pydantic

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1723,7 +1736,7 @@ def create_bulk_search_response(queries: List[BulkSearchQueryEntity], query_to_b
 
 def _vector_text_search(
         config: Config, index_name: str, query: Union[str, dict], result_count: int = 5, offset: int = 0,
-        searchable_attributes: Iterable[str] = None, verbose=0, filter_string: str = None, device=None,
+        searchable_attributes: Iterable[str] = None, verbose=0, filter_string: str = None, device: str = None,
         attributes_to_retrieve: Optional[List[str]] = None, boost: Optional[Dict] = None,
```
wanliAlex, line 1777: Optional[str] = None

src/marqo/s2_inference/clip_utils.py
```diff
@@ -192,7 +192,7 @@ class CLIP:
     conveniance class wrapper to make clip work easily for both text and image encoding
     """
 
-    def __init__(self, model_type: str = "ViT-B/32", device: str = 'cpu',  embedding_dim: int = None,
+    def __init__(self, model_type: str = "ViT-B/32", device: str = None,  embedding_dim: int = None,
                             truncate: bool = True, **kwargs) -> None:
```
wanliAlex, line 204: Optional[str] = None. 

something from chatgpt

Yes, using `device: Optional[str] = None` is generally better than `device: str = None` when you want to indicate that the `device` parameter is optional and can accept `None` as a valid value.

By using `Optional[str]`, you explicitly annotate that the parameter can either be a string or `None`. This conveys the intention that the parameter is optional and provides better clarity to other developers reading the code or using the function.

Using `Optional` from the `typing` module is a more precise and expressive way to indicate optional parameters in Python. It enhances readability and helps convey the intended contract of the function.

Here's an example demonstrating the usage of `Optional[str]`:

```python
from typing import Optional

def some_function(device: Optional[str] = None):
    # Perform operations based on the device
    if device is None:
        print("No device specified")
    else:
        print(f"Using device: {device}")

# Example usage
some_function()  # No device specified
some_function(device="cuda:0")  # Using device: cuda:0
some_function(device=None)  # No device specified
```

In the example above, the `device` parameter is annotated as `Optional[str]`, indicating that it can accept a string or `None`. Inside the function, the presence of `None` can be checked using an `if` statement to handle the case where no device is specified.

Using `Optional[str]` provides better clarity and avoids potential confusion about the intention of the parameter compared to simply using `str`.

src/marqo/s2_inference/s2_inference.py
```diff
@@ -55,6 +54,9 @@ def vectorise(model_name: str, content: Union[str, List[str]], model_properties:
         VectoriseError: if the content can't be vectorised, for some reason.
     """
 
+    if not device:
+        raise InvalidArgError(message=f"vectorise (internal function) cannot be called without setting device!")
+
```
wanliAlex, line 58: Raise a 500 error (runtimeerror) here. Based on this PR, under no circumstance, the device can be None? @pandu-k can i get your confirmation about this comment?

src/marqo/s2_inference/s2_inference.py
```diff
@@ -412,7 +413,9 @@ def _float_tensor_to_list(output: FloatTensor, device: str = get_default_device(
     Returns:
         List[List[float]]: _description_
     """
-
+    
+    # Hardcoded since get_default_device() used to return "cpu" anyway
+    device = "cpu"
     return output.detach().to(device).tolist()
```
wanliAlex: this one must be hardcoded as "CPU", as we can only convert a tensor in "CPU" to list.

src/marqo/tensor_search/parallel.py
```diff
@@ -184,13 +184,15 @@ def add_documents_mp(
         _type_: _description_
     """
 
-    selected_device = add_docs_params.device if add_docs_params.device is not None else config.indexing_device
-
+    if not add_docs_params.device:
+        raise InvalidArgError("You cannot call add_documents_mp without device set!")
+
```
wanliAlex, line 204: pydantic

wanliAlex (changes requested):

wanliAlex:
One general comment, can you double check if you should use `device: str` or `device: Optional[str] = None` in this PR to make sure it is clear what the arguments should be?

src/marqo/tensor_search/tensor_search.py
vicilliar: my mistake, will remove that comment.

src/marqo/tensor_search/tensor_search.py
vicilliar: Not sure about validating the search and bulk search parameters with pydantic, what are your thoughts @Jeadie ?

src/marqo/tensor_search/tensor_search.py
Jeadie: As discussed in slack, this would force a breaking change to py-marqo and a larger systematic change if we did it at the higher, request level. For now this is an improvement on the current situation.

Jeadie: Also yes, I don't believe we want to underscore our marqo variables

src/marqo/s2_inference/s2_inference.py
vicilliar: Will make this a 500 error, thanks!
Yes, under no circumstance can the device be none here. I just did not make `device` a required parameter because it would require rearranging the parameters, thus making big downstream changes.

src/marqo/s2_inference/clip_utils.py
vicilliar: I do not want to indicate that device is optional, because it really shouldn't be. I only left a default `None` because otherwise it would require rearranging the parameters, thus making big downstream changes. Maybe this can be done in a separate PR. cc-ing @pandu-k , let me know if this is alright.

src/marqo/s2_inference/s2_inference.py
vicilliar: done!

src/marqo/tensor_search/tensor_search.py
vicilliar: This error should never be seen by a user (only if a dev calls `add_documents` directly). To have this check done in Pydantic, the solution structure would have to change, either:
1. calculate the default device with the env var inside the Pydantic object definition
2. Just make the Pydantic object device required

src/marqo/tensor_search/tensor_search.py
vicilliar: I just put in the underscore to indicate it was an env var for internal use only, but I can remove it

vicilliar:
unit tests: https://github.com/marqo-ai/marqo/actions/runs/5265622093

src/marqo/s2_inference/clip_utils.py
pandu-k: Yep I think this is the right call @vicilliar, especially if we throw errors if we detect it is None

pandu-k: Actually, if we consider this function to be in a bad state if it is called with `device=None` (or not specified), can we raise error if we detect it?

src/marqo/s2_inference/clip_utils.py
```diff
@@ -504,7 +506,7 @@ def encode_text(self, sentence: Union[str, List[str]], normalize=True) -> FloatT
 
 
 class MULTILINGUAL_CLIP(CLIP):
-    def __init__(self, model_type: str = "multilingual-clip/ViT-L/14", device: str = 'cpu',  embedding_dim: int = None,
+    def __init__(self, model_type: str = "multilingual-clip/ViT-L/14", device: str = None,  embedding_dim: int = None,
                             truncate: bool = True, **kwargs) -> None:
```
pandu-k, line 522: This doesn't contain `super().__init__()`, so we need to validate device isn't None

src/marqo/s2_inference/onnx_clip_utils.py
```diff
@@ -56,7 +56,7 @@ class CLIP_ONNX(object):
     Load a clip model and convert it to onnx version for faster inference
     """
 
-    def __init__(self, model_name="onnx32/openai/ViT-L/14", device="cpu", embedding_dim: int = None,
+    def __init__(self, model_name: str ="onnx32/openai/ViT-L/14", device: str = None, embedding_dim: int = None,
                  truncate: bool = True,
                  load=True, **kwargs):
         self.model_name = model_name
```
pandu-k, line 63: Validation on device=None, unless we are happy with `self.device.startswith("cuda")` indicating that method was badly called.

src/marqo/s2_inference/processing/DINO_utils.py
```diff
@@ -82,15 +82,15 @@ def _get_DINO_transform(image_size: Tuple = (224, 224)) -> Any:
...
     Args:
         model (Any): ('vit_small', 'vit_base')
         transform (Any): _get_DINO_transform
         img (ImageType, optional): the image to infer on. Defaults to None.
         patch_size (int, optional): the patch size the model architecture uses. Defaults to None.
-        device (str, optional): device for the model to run on. Defaults to "cpu".
+        device (str, optional): device for the model to run on. Required to be set
 
     Returns:
         FloatTensor: returns N x w x h tensor
```
pandu-k, line 97: Add validation for device name

src/marqo/s2_inference/processing/image.py
```diff
@@ -151,13 +151,13 @@ def process(self):
...
                 nms: bool = True, replace_small: bool = True, top_k: int = 10, 
                 filter_bb: bool = True, min_area_replace: float = 60*60, **kwargs):
         """_summary_
 
         Args:
-            device (str, optional): the device to run the model on. Defaults to 'cpu'.
+            device (str): the device to run the model on. Required to be set.
             size (Tuple, optional): the final image size to go to the model. Defaults to (224, 224).
             min_area (float, optional): the min area (pixels) that a box must meet to be kept. 
                 areas lower than this are removed. Defaults to 60*60.
```
pandu-k, line 165: validation for device

src/marqo/s2_inference/s2_inference.py
pandu-k: Yes this should be a 500

src/marqo/s2_inference/s2_inference.py
```diff
@@ -402,18 +402,18 @@ def _check_output_type(output: List[List[float]]) -> bool:
...
         output (FloatTensor): _description_
 
     Returns:
         List[List[float]]: _description_
     """
-
-    return output.detach().to(device).tolist()
+    
+    # Hardcoded to CPU always
+    return output.detach().to("cpu").tolist()
```
pandu-k, line 416: @wanliAlex what's the impact of this change?

src/marqo/s2_inference/sbert_utils.py
```diff
@@ -11,7 +11,7 @@
 class Model:
     """ generic model wrapper class
     """
-    def __init__(self, model_name: Optional[str] = None, device: str = 'cpu', batch_size: int = 2048, embedding_dim=None, max_seq_length=None , **kwargs) -> None:
+    def __init__(self, model_name: Optional[str] = None, device: str = None, batch_size: int = 2048, embedding_dim=None, max_seq_length=None , **kwargs) -> None:
```
pandu-k, line 15: Can we validate device can't be None?

src/marqo/tensor_search/on_start_script.py
```diff
@@ -84,8 +84,17 @@ def id_to_device(id):
...
+
+        """
+            This is set once at startup time. We assume it will NOT change,
+            if it does, health check should throw a warning.
+        """
+        device_count = 0
+        os.environ["MARQO_BEST_AVAILABLE_DEVICE"] = "cpu"
+        if torch.cuda.is_available():
+            device_count = torch.cuda.device_count()
+            os.environ["MARQO_BEST_AVAILABLE_DEVICE"] = "cuda"
```
pandu-k: I'd prefer this being only set once (rather than set to CPU and then being overridden by CUDA):
```
if cuda is available:
    "MARQO_BEST_AVAILABLE_DEVICE" -> "cuda"
else:
    "MARQO_BEST_AVAILABLE_DEVICE" -> "cpu"

```

for multithreading reasons, so there isn't  a path where this env var ever changes (even though this is unlikely to be a problem).

pandu-k: Can we also break this out into its own function/static method? something like `on_start_script.set_best_available_device`

src/marqo/tensor_search/tensor_search.py
pandu-k: There's another way to do this in pydantic. Make a subclass of AddDocsParams (like AddDocsParamsWithDevice) where device must not be None

pandu-k: Good to get @Jeadie 's input

src/marqo/tensor_search/tensor_search.py
pandu-k: 👍

src/marqo/tensor_search/tensor_search.py
```diff
@@ -950,7 +963,13 @@ def bulk_search(query: BulkSearchQuery, marqo_config: config.Config, verbose: bo
...
         return {"result": []}
 
-    selected_device = marqo_config.indexing_device if device is None else device
+    if device is None:
+        selected_device = utils.read_env_vars_and_defaults("MARQO_BEST_AVAILABLE_DEVICE")
+        if selected_device is None:
+            raise errors.InternalError("Best available device was not properly determined on Marqo startup.")
+        logger.debug(f"No device given for bulk_search. Defaulting to best available device: {selected_device}")
+    else:
+        selected_device = device
```
pandu-k, line 986: This seems to be repeated logic. Consider moving it into its own utils function.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1751,6 +1779,7 @@ def _vector_text_search(
         ridiculous number of attributes
         - Should not be directly called by client - the search() method should
         be called. The search() method adds syncing
+        - device should ALWAYS be set
```
pandu-k, line 1805: Validation if not set

tests/tensor_search/test_bulk_search.py
```diff
@@ -398,20 +404,6 @@ def test_bulk_search_with_no_max_searchable_attributes(self):
...
-        add_docs_caller(
-            config=self.config, index_name=self.index_name_1, docs=[
-                {"abc": "Exact match hehehe", "other field": "baaadd", "_id": "id1-first"},
-                {"abc": "random text", "other field": "Close match hehehe", "_id": "id1-second"},
-            ], auto_refresh=True
-        ) 
-        api.bulk_search(BulkSearchQuery(queries=[{
-            "index": self.index_name_1,
-            "q": "title about some doc",
-        }]), marqo_config=self.config, device="cpu")
```
pandu-k, line 412: why was this deleted?

src/marqo/tensor_search/tensor_search.py
pandu-k: then you can also create unit tests around it

src/marqo/tensor_search/on_start_script.py
pandu-k: it then becomes easier to create unit tests for it

tests/tensor_search/test_search.py
```diff
@@ -574,27 +581,6 @@ def test_filtering_bad_syntax(self):
         except InvalidArgError:
             pass
 
-    def test_set_device(self):
```
pandu-k, line 577: This is a good test to have, but can be refactored so that it overrides the BEST_DEVICE env vr

vicilliar:
Updated unit test run: https://github.com/marqo-ai/marqo/actions/runs/5275160720

src/marqo/s2_inference/clip_utils.py
vicilliar: Ok, I can add that condition for all these functions to manually raise an error when device is None!

src/marqo/s2_inference/processing/DINO_utils.py
vicilliar: Do you just mean checking if it's `None` here? Or the entire device validation process? I'd go for the former, as these errors are more just to correct developers who make mistakes

src/marqo/s2_inference/s2_inference.py
vicilliar: I talked with him about this already, the behavior was actually already like this. `device` was based on `get_default_device()`, which was hardcoded to `cpu`

src/marqo/tensor_search/on_start_script.py
vicilliar: Can do! Although I kept it in `CUDAAvailable` to reuse the `if torch.cuda.is_available()` call. It's fine tbh since it only happens ocne on startup anyway

pandu-k (commented):
👍

src/marqo/s2_inference/processing/DINO_utils.py
pandu-k: Yes, just checking if it's None, apologies for the confusion!

src/marqo/s2_inference/processing/image.py
pandu-k: (as in checking whether device=None)

src/marqo/s2_inference/s2_inference.py
pandu-k: Makes sense. Thanks for clarifying!

src/marqo/tensor_search/tensor_search.py
pandu-k: (to clarify, if device=None)

pandu-k (commented):
-

pandu-k (commented):
looks good.

tests/s2_inference/test_sbert_utils.py
```diff
@@ -20,12 +20,3 @@ def test_sbert_with_no_device(self):
             raise AssertionError
         except InternalError as e:
             pass
-    
-    def test_model_with_no_device(self):
```
pandu-k: put test back

tests/tensor_search/test_model_auth.py
```diff
@@ -2030,6 +2066,7 @@ def test_model_loads_from_all_add_docs_derivatives(self):
 
         for add_docs_method, kwargs in [
                 (tensor_search.add_documents_orchestrator, {'batch_size': 10}),
+                # TODO: add add_documents and add_documents_mp ?
```
pandu-k, line 2068: add_docs_mp can crash CI pipelines, that's why it is excluded

tests/tensor_search/test_parallel.py
```diff
@@ -45,4 +55,16 @@ def test_add_documents_parallel(self) -> None:
         res = tensor_search.add_documents_orchestrator(config=self.config, add_docs_params=AddDocsParams(
                 index_name=self.index_name_1, docs=data, auto_refresh=True),
             batch_size=10, processes=1)
-        res = tensor_search.search(config=self.config, text='something', index_name=self.index_name_1)
\ No newline at end of file
+        res = tensor_search.search(config=self.config, text='something', index_name=self.index_name_1)
+    
+    def test_add_documents_mp_no_device(self) -> None:
```
pandu-k, line 60: Multiprocessing can crash Marqo. Run different CI pipelines a few times. If this causes Marqo to crash, just remove it.

pandu-k (approved):

wanliAlex (approved):

PR #516, opened 2023-06-26
Disable auto create index

farshidz:
Feature

When adding documents to an index that does not exist, Marqo automatically and silently creates an index with a default index configuration. 

Attempting to add documents to an index that does not exist will fail with an error. Indices must be created explicitly before adding documents.

Yes. Users must ensure an index exists before documents are added, or must handle the error. Code that relies on the removed automatic index creation behaviour will fail.

Yes. Only unit tests.

None

None

* **Other information**:


tests/tensor_search/test_bulk_search.py
```diff
@@ -312,6 +312,7 @@ def setUp(self) -> None:
         self.index_name_2 = "my-test-index-2"
         self.index_name_3 = "my-test-index-3"
         self._delete_test_indices()
+        self._create_test_indices()
```
pandu-k, line 316: Let's also delete the test indices when the test suite is torn down, see here for an example: https://github.com/marqo-ai/marqo/blob/2b14a6ad941dc5ddacbd80479b6edd8da018fc37/tests/marqo_test.py#L21

Deleting common test index names in the test super class is a simpler way to do this. This is done in py-marqo tests: https://github.com/marqo-ai/py-marqo/blob/mainline/tests/marqo_test.py#L136

We're moving towards being a bit stricter with cleaning up environments after testing.

pandu-k (approved):

PR #523, opened 2023-06-30
Remove Parallel (Processes) in Marqo

wanliAlex:
remove Parallel processes in marqo

we have a `processes` parameter in marqo `add_or_replace_documents` API that uses multiprocess to index documents.
We decide to remove the support of the multiprocess feature and remove the support of this parameter in the API, for the following reasons:

1. It blows out memory (it seems that loaded models are replicated for each process)

2. The (time) overhead from starting multiple processes makes it unviable for ~30 second HTTP requests. It only makes sense for much longer running requests (which is an unusual pattern for a REST API)

yes, this is a breaking change

passes

on going

parameter processes has already been removed from the doc

* **Other information**:
not


src/marqo/tensor_search/api.py
```diff
@@ -185,6 +185,10 @@ def add_or_replace_documents(
         ),
         mappings: typing.Optional[dict] = Depends(api_utils.decode_mappings)):
     """add_documents endpoint (replace existing docs with the same id)"""
+    if processes is not None:
+        raise InvalidArgError("The parameter `process` is no longer supported anymore. Please check the documentation for more information. \n"
```
farshidz: Parameter is `processes` (plural). Also I don't think backticks will be picked up in error message i.e. won't affect formatting. 

Could probably do without the line break. Error messages are generally one paragraph

src/marqo/tensor_search/tensor_search.py
```diff
@@ -251,40 +250,12 @@ def add_documents_orchestrator(
...
-        if add_docs_params.index_name in get_cache():
-            logger.info(f'deleting cache entry for {add_docs_params.index_name} after parallel add documents')
-            del get_cache()[add_docs_params.index_name]
-
-        return results
     else:
         if batch_size < 0:
             raise errors.InvalidArgError("Batch size can't be less than 1!")
-        logger.debug(f"batch_size={batch_size} and processes={processes} - batching using a single process")
+        logger.debug(f"batch_size={batch_size} - batching using a single process")
```
farshidz: I think `batching using a single process` is superfluous now that there's no other option

src/marqo/tensor_search/api.py
```diff
@@ -185,6 +185,10 @@ def add_or_replace_documents(
         ),
         mappings: typing.Optional[dict] = Depends(api_utils.decode_mappings)):
     """add_documents endpoint (replace existing docs with the same id)"""
+    if processes is not None:
```
farshidz: This needs a test

tests/tensor_search/test_parallel.py
```diff
@@ -14,57 +14,5 @@ class TestAddDocumentsPara(MarqoTestCase):
...
-        
-        data = [{'text':f'something {str(i)}', '_id': str(i)} for i in range(100)]
-        try:
-            res = parallel.add_documents_mp(config=self.config, add_docs_params=AddDocsParams(
-                    index_name=self.index_name_1, docs=data, auto_refresh=True),
-                batch_size=10, processes=2)
-            raise AssertionError
-        except InternalError:
-            pass
+    pass
```
farshidz: Why keep the empty file?

farshidz (changes requested):

src/marqo/tensor_search/api.py
wanliAlex: updated

pandu-k: @farshidz We use backticks as quotation marks in our error messages, following the style found in other projects, like Meilisearch: https://www.meilisearch.com/docs/reference/errors/overview#error-format

tests/tensor_search/test_parallel.py
wanliAlex: deleted now

src/marqo/tensor_search/tensor_search.py
wanliAlex: updated

src/marqo/tensor_search/api.py
wanliAlex: We just delete this error and make it easier.

farshidz (approved):

PR #538, opened 2023-07-13
Add tensor_fields parameter to add_documents

farshidz:
Feature

To add new documents to an index, `POST index_name/documents` receives `non_tensor_fields` and treats all fields except those in `non_tensor_fields` as tensor fields.

`POST index_name/documents` will now accept `tensor_fields`. When this is provided, all fields not in `tensor_fields` are treated at non-tensor fields. 

Yes. While the change is backwards compatible for requests that specify `non_tensor_fields`, requests that specify neither `tensor_fields` nor `non_tensor_fields` will now lead to an error response. 

Yes

In progress

In progress

* **Other information**:


src/marqo/tensor_search/add_docs.py
```diff
@@ -44,10 +49,16 @@ def threaded_download_images(allocated_docs: List[dict], image_repo: dict,
         metric_obj = RequestMetricsStore.for_request()
         RequestMetricsStore.set_in_request(metrics=metric_obj)
 
+    def is_non_tensor_field(f: str) -> bool:
```
vicilliar: would it be better to have this function at the file level, so it's easier to make unit tests for?

src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -24,6 +24,7 @@ class Config:
         extra = "forbid" # Raise error on unknown fields
 
     nonTensorFields: List = Field(default_factory=list)
+    tensorFields: List = Field(default_factory=list)
```
vicilliar: Why is this not `Field(default_factory=None)` just like what you did for `AddDocsParams`?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -370,17 +373,23 @@ def add_documents(config: Config, add_docs_params: AddDocsParams):
                 continue
             chunk_values_for_filtering[key] = value
 
+        def is_non_tensor_field(field: str) -> bool:
```
vicilliar: since this is repeated logic, it may be worth making it a util func that accepts: `field`, `tensor_fields (which may be None)`, and `non_tensor_fields (which may be None)`

src/marqo/tensor_search/add_docs.py
pandu-k: I agree, I prefer to avoid nested functions and to avoid impure functions (functions that mutate/depends on an external state) if possible, unless absolutely necessary.

If we decide to move the function to the file level we could also add some straightforward tests for it

src/marqo/tensor_search/add_docs.py
```diff
@@ -20,8 +22,10 @@ def threaded_download_images(allocated_docs: List[dict], image_repo: dict,
         allocated_docs: docs with images to be downloaded by this thread,
         image_repo: dictionary that will be mutated by this thread. It will add PIL images
             as values and the URLs as keys
+        tensor_fields: A tuple of tensor_fields. Images will be downloaded for these fields only. Cannot be provided
+            at the same time as `non_tensor_fields`.
         non_tensor_fields: A tuple of non_tensor_fields. No images will be downloaded for
-            these fields
+            these fields. Cannot be provided at the same time as `tensor_fields`.
```
pandu-k, line 29: We don't validate that `non_tensor_fields ` and `tensor_fields` aren't allowed to co-occur in this function. We should:
a) if we expected validation to be done beforehand we should explicitly state this in the docstring, along with the function/code that is expected to do this validation. This is to inform future devs in case they want to refactor
b) validate it here, and raise an internal error because this is an unusual state. If this is an unusual state because a user error should have been raised earlier, this should be stated in the docstring

src/marqo/tensor_search/add_docs.py
```diff
@@ -77,13 +88,16 @@ def threaded_download_images(allocated_docs: List[dict], image_repo: dict,
...
     Args:
         docs: docs with images to be downloaded. These will be allocated to each thread
         thread_count: number of threads to spin up
+        tensor_fields: A tuple of tensor_fields. Images will be downloaded for these fields only. Cannot be provided
+            at the same time as `non_tensor_fields`.
         non_tensor_fields: A tuple of non_tensor_fields. No images will be downloaded for
-            these fields
+            these fields. Cannot be provided at the same time as `tensor_fields`.
         image_download_headers: A dict of image download headers for authentication.
     This should be called only if treat URLs as images is True
```
pandu-k, line 105: ditto regarding communication about `non_tensor_fields` and `tensor_fields` validation

src/marqo/tensor_search/tensor_search.py
pandu-k: I like this idea!

src/marqo/tensor_search/web/api_utils.py
```diff
@@ -161,19 +161,36 @@ def add_docs_params_orchestrator(index_name: str, body: Union[AddDocsBodyParams,
...
         image_download_headers = body.imageDownloadHeaders
 
+        if tensor_fields and non_tensor_fields:
+            raise BadRequestError('Cannot provide nonTensorFields when tensorFields is defined. '
+                                  'nonTensorFields has been deprecated and will be removed in Marqo 2.0.0. '
+                                  'Its use is discouraged.')
+
+        if not tensor_fields and not non_tensor_fields:
+            raise BadRequestError('You must include the tensorFields parameter. '
+                                  'Use tensorFields=[] to index for lexical-only search.')
```
pandu-k: The pattern we are moving to now is to relegate validation to move validation to AddDocsParams validation methods

src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -61,14 +62,25 @@ class Config:
...
+    tensor_fields: Optional[List] = Field(default_factory=None)
     image_download_thread_count: int = 20
     image_download_headers: dict = Field(default_factory=dict)
     use_existing_tensors: bool = False
     mappings: Optional[dict] = None
     model_auth: Optional[ModelAuth] = None
 
-    def __init__(self, **data: dict):
+    @root_validator
+    def validate_fields(cls, values):
```
pandu-k, line 74: We are moving towards the primary validation being included in the AddDocsParams object (so raising user errors  rather than internal errors here)

src/marqo/tensor_search/web/api_utils.py
pandu-k: Good error messages! 

Nit on the wording: 
```
You must include the tensorFields parameter.
````
 -> 
```
Required parameter `tensorFields` is missing from the request body.
```

File: src/marqo/tensor_search/models/add_docs_objects.py (hunk 2 above)
Comment (farshidz) [reply src/marqo/tensor_search/models/add_docs_objects.py]: Thanks. Fixed

File: src/marqo/tensor_search/add_docs.py (hunk 4 above)
Comment (farshidz) [reply src/marqo/tensor_search/add_docs.py:29]: Addressed

File: src/marqo/tensor_search/models/add_docs_objects.py (hunk 7 above)
Comment (farshidz) [reply src/marqo/tensor_search/models/add_docs_objects.py:74]: As discussed this will be refactored as a separate PR and will remain unchanged for this one

File: src/marqo/tensor_search/web/api_utils.py (hunk 6 above)
Comment (farshidz) [reply src/marqo/tensor_search/web/api_utils.py]: Changed

File: src/marqo/tensor_search/utils.py [hunk 8]
```diff
@@ -335,3 +335,17 @@ def get_best_available_device() -> str:
                            f"Invalid device: {device}. Must be either 'cpu' or start with 'cuda'.")
     return device
 
+
+def is_tensor_field(field: str,
+                    tensor_fields: Optional[List[str]] = None,
+                    non_tensor_fields: Optional[List[str]] = None
+                    ) -> bool:
+    """Determine whether a field is a tensor field or not."""
```
pandu-k: nit:  `tensor field or not. `->`tensor field or not, for add_documents calls.` (this function doesn't make sense outside this context, for example: during search/any other endpoint). In particular the validation doesn't make

pandu-k (approved):

PR #552, opened 2023-07-25
Revise health check API to index level

wanliAlex:
add an index level health check API

we do not take index-name in the heath check API. We use `"/health"` API.

The health check API is moved to `"/indexes/{index_name}/health"`

we add a new API, however it's still backward compatible

running

working on 

working on

* **Other information**:
no

### DIFF
--- a/src/marqo/tensor_search/enums.py
+++ b/src/marqo/tensor_search/enums.py
@@ -146,6 +146,17 @@ class ModelProperties:
 class InferenceParams:
     model_auth = "model_auth"
 # Perhaps create a ThrottleType to differentiate thread_count and data_size throttling mechanisms
+
+
+class HealthStatuses(str, Enum):
+    green = "green"
+    yellow = "yellow"
+    red = "red"
+
+    def __lt__(self, other):
+        # Define the custom comparison logic
+        status_order = [self.green, self.yellow, self.red]
+        return status_order.index(self) < status_order.index(other)
     
 
 
--- a/src/marqo/tensor_search/health.py
+++ b/src/marqo/tensor_search/health.py
@@ -51,7 +51,8 @@ def get_marqo_os_status(config: Config, index_name: Optional[str] = None) -> Uni
     return marqo_os_status
 
 
-def aggregate_status(marqo_status: Union[str, HealthStatuses], marqo_os_status: Union[str, HealthStatuses]) -> Tuple[Union[str,HealthStatuses], Union[str, HealthStatuses]]:
+def aggregate_status(marqo_status: Union[str, HealthStatuses], marqo_os_status: Union[str, HealthStatuses]) \
+        -> Tuple[Union[str,HealthStatuses], Union[str, HealthStatuses]]:
     """Aggregate the Marqo instance and Marqo-os backend status."""
     aggregated_marqo_status = max(marqo_status, marqo_os_status)
     return aggregated_marqo_status, marqo_os_status
--- a/src/marqo/tensor_search/health.py
+++ b/src/marqo/tensor_search/health.py
@@ -51,7 +51,7 @@ def get_marqo_os_status(config: Config, index_name: Optional[str] = None) -> Uni
     return marqo_os_status
 
 
-def aggregate_status(marqo_status: str, marqo_os_status: str) -> Tuple[Union[str,HealthStatuses], Union[str, HealthStatuses]]:
+def aggregate_status(marqo_status: Union[str, HealthStatuses], marqo_os_status: Union[str, HealthStatuses]) -> Tuple[Union[str,HealthStatuses], Union[str, HealthStatuses]]:
     """Aggregate the Marqo instance and Marqo-os backend status."""
     aggregated_marqo_status = max(marqo_status, marqo_os_status)
     return aggregated_marqo_status, marqo_os_status
--- a/src/marqo/tensor_search/health.py
+++ b/src/marqo/tensor_search/health.py
@@ -6,6 +6,7 @@ from marqo import errors
 from marqo._httprequests import HttpRequests
 from marqo.tensor_search.enums import HealthStatuses
 
+
 def generate_heath_check_response(config: Config, index_name: Optional[str] = None) -> dict:
     """Generate the health check response for check_heath(), check_index_health() APIs in tensor_search"""
     marqo_status = get_marqo_status()
--- a/src/marqo/tensor_search/health.py
+++ b/src/marqo/tensor_search/health.py
@@ -1,0 +1,6 @@
+from typing import Optional
+from marqo.config import Config
+
+
+def generate_heath_check_response(config: Config, index_name: Optional[str] = None) -> dict:
+    pass
--- a/src/marqo/tensor_search/health.py
+++ b/src/marqo/tensor_search/health.py
@@ -40,7 +40,8 @@ def get_marqo_os_status(config: Config, index_name: Optional[str] = None) -> str
         marqo_os_health_check_response = None
 
     if marqo_os_health_check_response is not None:
-        if "status" in marqo_os_health_check_response:
+        if "status" in marqo_os_health_check_response and marqo_os_health_check_response['status'] \
+                in list(HealthStatuses):
             marqo_os_status = HealthStatuses[marqo_os_health_check_response['status']]
         else:
             marqo_os_status = HealthStatuses.red
--- a/src/marqo/tensor_search/health.py
+++ b/src/marqo/tensor_search/health.py
@@ -4,7 +4,7 @@ import copy
 from marqo.config import Config
 from marqo import errors
 from marqo._httprequests import HttpRequests
-
+from marqo.tensor_search.enums import HealthStatuses
 
 def generate_heath_check_response(config: Config, index_name: Optional[str] = None) -> dict:
     """Generate the health check response for check_heath(), check_index_health() APIs in tensor_search"""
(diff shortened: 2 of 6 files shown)


src/marqo/tensor_search/tensor_search.py
```diff
@@ -1807,7 +1807,55 @@ def sort_chunks(docs: dict) -> List:
     return sorted(as_list, key=lambda x: x["chunks"][0]["_score"], reverse=True)
 
 
+def check_index_health(config: Config, index_name: str) -> dict:
+    """Checks the health of an index
```
wanliAlex, line 1812: A question about coding practice here. There is a lot of duplicated code with the old API. Would it be good to wrap the duplicated code to avoid duplication here? Or should I keep the duplication so we can remove the old API easily in the future?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1807,7 +1807,55 @@ def sort_chunks(docs: dict) -> List:
...
+
+    if marqo_os_health_check is not None:
+        if "status" in marqo_os_health_check:
+            marqo_os_status = marqo_os_health_check['status']
+        else:
+            marqo_os_status = "red"
+    else:
+        marqo_os_status = "red"
+
+    marqo_status = marqo_status if statuses[marqo_status] >= statuses[marqo_os_status] else marqo_os_status
```
Mehul-Porwal: Can this be simply written as `status = max(statuses[marqo_status], statuses[marqo_os_status])`?

Mehul-Porwal (approved):

src/marqo/tensor_search/tensor_search.py
wanliAlex: I don't think so. We want to return "green", "yellow", and "red" instead of value. We will still need a reverse mapping from value to textual status.

src/marqo/tensor_search/tensor_search.py
pandu-k: You can think about breaking out some repeated code into its own dedicated function.

For example:
```
def calculate_status (...) -> dict:
```

This also makes the logic contained in the function much easier to test

src/marqo/tensor_search/tensor_search.py
pandu-k: @wanliAlex  Maybe the reverse mappings is better? 

The logic within this code is numbers, and at the end we translate it to colours. It may make the logic more straightforward

src/marqo/tensor_search/utils.py
```diff
@@ -349,3 +349,27 @@ def is_tensor_field(field: str,
         return field in tensor_fields
     else:
         return field not in non_tensor_fields
+
+
+def calculate_health_status(marqo_os_health_check_response: Optional[Dict]) -> dict:
```
pandu-k, line 354: let's move this to a dedicated `health.py` file

src/marqo/tensor_search/utils.py
```diff
@@ -349,3 +349,27 @@ def is_tensor_field(field: str,
...
+    """Calculate the health status of Marqo based on the response API from Marqo-os ."""
```
pandu-k, line 355: This function actually has 2 purposes:
1. To handle the OpenSearch health check response 
2. To handle calculating the Marqo health

Perhaps we should break this into more functions: 
i. One for calculating the Marqo-os healthcheck, and handling errors from Marqo-os
ii. One for calculating this instance's health 
iii. One for calculating the aggregated system health
iv. One for formatting the health response for the user 

We can probably combine ii and iii into a single function for now

src/marqo/tensor_search/enums.py
```diff
@@ -146,6 +146,17 @@ class ModelProperties:
...
+
+class HealthStatuses(str, Enum):
+    green = "green"
+    yellow = "yellow"
+    red = "red"
+
+    def __lt__(self, other):
+        # Define the custom comparison logic
+        status_order = [self.green, self.yellow, self.red]
+        return status_order.index(self) < status_order.index(other)
```
pandu-k, line 159: This is a nice way of doing this.

It also might makes sense to define an equality:
```
def __eq__(): 
```

src/marqo/tensor_search/health.py
```diff
@@ -0,0 +1,57 @@
...
+
+    return {
+        "status": marqo_status,
+        "backend": {
+            "status": marqo_os_status
+        }
+    }
+
+
+def get_marqo_status() -> str:
```
pandu-k: nit: should the return type be Union `Union[str, HealthStatuses]`

src/marqo/tensor_search/health.py
```diff
@@ -0,0 +1,57 @@
...
+    """Check the Marqo instance status."""
+    return HealthStatuses.green
+
+
+def get_marqo_os_status(config: Config, index_name: Optional[str] = None) -> str:
```
pandu-k: ditto on  return type

src/marqo/tensor_search/health.py
```diff
@@ -0,0 +1,57 @@
...
+from marqo import errors
+from marqo._httprequests import HttpRequests
+from marqo.tensor_search.enums import HealthStatuses
+
+
+def generate_heath_check_response(config: Config, index_name: Optional[str] = None) -> dict:
+    """Generate the health check response for check_heath(), check_index_health() APIs in tensor_search"""
+    marqo_status = get_marqo_status()
+    marqo_os_status = get_marqo_os_status(config, index_name=index_name)
+    marqo_status, marqo_os_status = aggregate_status(marqo_status, marqo_os_status)
```
pandu-k: nit: to better reflect what is happening in the code you may want to rename `marqo_status`

```
aggregated_status, marqo_os_status = aggregate_status(...
```

src/marqo/tensor_search/health.py
```diff
@@ -0,0 +1,57 @@
...
+            marqo_os_status = HealthStatuses.red
+    else:
+        marqo_os_status = HealthStatuses.red
+
+    return marqo_os_status
+
+
+def aggregate_status(marqo_status: str, marqo_os_status: str) -> Tuple[str, str]:
+    """Aggregate the Marqo instance and Marqo-os backend status."""
+    marqo_status = max(marqo_status, marqo_os_status)
```
pandu-k: let's clarify this by calling `marqo_status` `aggregate_status` or something

src/marqo/tensor_search/health.py
```diff
@@ -0,0 +1,57 @@
...
+            marqo_os_status = HealthStatuses[marqo_os_health_check_response['status']]
+        else:
+            marqo_os_status = HealthStatuses.red
+    else:
+        marqo_os_status = HealthStatuses.red
+
+    return marqo_os_status
+
+
+def aggregate_status(marqo_status: str, marqo_os_status: str) -> Tuple[str, str]:
```
pandu-k: ditto on the return type

src/marqo/tensor_search/health.py
```diff
@@ -0,0 +1,57 @@
...
+    """Check the Marqo-os backend status."""
+    TIMEOUT = 3
+    marqo_os_health_check_response = None
+    path = f"_cluster/health/{index_name}" if index_name else "_cluster/health"
+
+    try:
+        timeout_config = copy.deepcopy(config)
+        timeout_config.timeout = TIMEOUT
+        marqo_os_health_check_response = HttpRequests(timeout_config).get(path=path)
+    except errors.BackendCommunicationError:
```
pandu-k: We may also need to handle timeout errors : https://github.com/marqo-ai/py-marqo/blob/d22fe41e9cf7efb10ef1a9a3ed055746612041a2/src/marqo/_httprequests.py#L77C34-L77C34 

or internal errors, which is the super class of these errors : https://github.com/marqo-ai/py-marqo/blob/d22fe41e9cf7efb10ef1a9a3ed055746612041a2/src/marqo/errors.py#L145C15-L145C15

tests/tensor_search/test_healthcheck.py
```diff
@@ -47,3 +49,64 @@ def run():
...
+    def test_index_health_check_unknown_backend_response(self):
+        mock__get = mock.MagicMock()
+        mock__get.return_value = dict()
+
+        # Ensure the index does not exist
+        with self.assertRaises(IndexNotFoundError):
+            tensor_search.delete_index(index_name=self.index_name, config=self.config)
+        @mock.patch("marqo._httprequests.HttpRequests.get", mock__get)
+        def run():
+            health_check_status = tensor_search.check_index_health(index_name=self.index_name, config=self.config)
```
pandu-k, line 93: is the opensearch behaviour to be red, if an index is not found? or would it throw an IndexNotFound type error?

src/marqo/tensor_search/health.py
wanliAlex: updated!

src/marqo/tensor_search/health.py
wanliAlex: Internal errors, yes

tests/tensor_search/test_healthcheck.py
wanliAlex: It will raise an IndexNotFound error because it is not caught by internal errors.

src/marqo/tensor_search/enums.py
wanliAlex: Just confirmed, this is not needed.

pandu-k (approved):

PR #555, opened 2023-07-27
Opensearch disk watermark breach handling

vicilliar:
Feature

1. When OpenSearch returns a 429 error, Marqo automatically converts it to `TooManyRequests`, when sometimes it can be a `rejectedRequest` due to breached flood stage disk watermark.
2. Health check is unaffected by breached flood stage disk watermark.

1. OpenSearch errors with the error type `cluster_block_exception` are now returned as 507 Insufficient Storage errors, as they are caused by a breached flood stage disk watermark.
2. Health check will now return "red" if available OpenSearch disk space is below the flood stage disk watermark.

No

In progress

* **Other information**:

### DIFF
--- a/src/marqo/_httprequests.py
+++ b/src/marqo/_httprequests.py
@@ -15,7 +15,8 @@ from marqo.errors import (
     IndexAlreadyExistsError,
     InvalidIndexNameError,
     HardwareCompatabilityError,
-    IndexMaxFieldsError, TooManyRequestsError
+    IndexMaxFieldsError, TooManyRequestsError,
+    DiskWatermarkBreachError
 )
 from urllib3.exceptions import InsecureRequestWarning
 import warnings
@@ -144,11 +145,6 @@ def convert_to_marqo_web_error_and_raise(response: requests.Response, err: reque
         pprint.pprint(response_dict)
     except JSONDecodeError:
         raise_catchall_http_as_marqo_error(response=response, err=err)
-    if response.status_code == 429:
-        # TODO, raise a different marqo error here.
-        raise TooManyRequestsError(
-            message="Marqo-OS received too many requests! "
-                    "Please try reducing the frequency of add_documents and update_documents calls.")
 
     try:
         open_search_error_type = response_dict["error"]["type"]
@@ -172,9 +168,21 @@ def convert_to_marqo_web_error_and_raise(response: requests.Response, err: reque
             if "limit of total fields" in reason and "exceeded" in reason:
                 raise IndexMaxFieldsError(message="Exceeded maximum number of "
                                                   "allowed fields for this index.")
+        elif open_search_error_type == "cluster_block_exception":
+            # TODO: change the open search error type depending on experiment results
+            raise DiskWatermarkBreachError(message="OpenSearch cluster indexing has been blocked, most likely due to breach \
+                                           of the `flood_stage` disk watermark. For more information, please refer \
+                                           to your cluster settings: `cluster.routing.allocation.disk.watermark` at \
+                                           https://opensearch.org/docs/2.7/api-reference/cluster-api/cluster-settings/")
     except KeyError:
         pass
 
+    # for throttling
+    if response.status_code == 429:
+        raise TooManyRequestsError(
+            message="Marqo-OS received too many requests! "
+                    "Please try reducing the frequency of add_documents and update_documents calls.")
+    
     try:
         if response_dict["found"] is False:
             raise DocumentNotFoundError(
--- a/src/marqo/_httprequests.py
+++ b/src/marqo/_httprequests.py
@@ -139,10 +139,9 @@ def convert_to_marqo_web_error_and_raise(response: requests.Response, err: reque
     Raises:
         MarqoWebError - some type of Marqo Web error
     """
+
     try:
         response_dict = response.json()
-        print("DEBUG: OpenSearch error response: ")
-        pprint.pprint(response_dict)
     except JSONDecodeError:
         raise_catchall_http_as_marqo_error(response=response, err=err)
 
@@ -169,7 +168,6 @@ def convert_to_marqo_web_error_and_raise(response: requests.Response, err: reque
                 raise IndexMaxFieldsError(message="Exceeded maximum number of "
                                                   "allowed fields for this index.")
         elif open_search_error_type == "cluster_block_exception":
-            # TODO: change the open search error type depending on experiment results
             raise DiskWatermarkBreachError(message="OpenSearch cluster indexing has been blocked, most likely due to breach \
                                            of the `flood_stage` disk watermark. For more information, please refer \
                                            to your cluster settings: `cluster.routing.allocation.disk.watermark` at \
--- a/src/marqo/_httprequests.py
+++ b/src/marqo/_httprequests.py
@@ -168,10 +168,10 @@ def convert_to_marqo_web_error_and_raise(response: requests.Response, err: reque
                 raise IndexMaxFieldsError(message="Exceeded maximum number of "
                                                   "allowed fields for this index.")
         elif open_search_error_type == "cluster_block_exception":
-            raise DiskWatermarkBreachError(message="OpenSearch cluster indexing has been blocked, most likely due to breach \
-                                           of the `flood_stage` disk watermark. For more information, please refer \
-                                           to your cluster settings: `cluster.routing.allocation.disk.watermark` at \
-                                           https://opensearch.org/docs/2.7/api-reference/cluster-api/cluster-settings/")
+            raise DiskWatermarkBreachError(message="OpenSearch cluster indexing has been blocked, most likely due to breach "
+                                           "of the `flood_stage` disk watermark. Try freeing up disk space on your OpenSearch instances. "
+                                           "For more information, please refer to cluster settings: `cluster.routing.allocation.disk.watermark` at "
+                                           "https://opensearch.org/docs/2.7/api-reference/cluster-api/cluster-settings/")
     except KeyError:
         pass
 
--- a/src/marqo/_httprequests.py
+++ b/src/marqo/_httprequests.py
@@ -143,6 +143,7 @@ def convert_to_marqo_web_error_and_raise(response: requests.Response, err: reque
     except JSONDecodeError:
         raise_catchall_http_as_marqo_error(response=response, err=err)
     if response.status_code == 429:
+        # TODO, raise a different marqo error here.
         raise TooManyRequestsError(
             message="Marqo-OS received too many requests! "
                     "Please try reducing the frequency of add_documents and update_documents calls.")
--- a/src/marqo/_httprequests.py
+++ b/src/marqo/_httprequests.py
@@ -140,6 +140,8 @@ def convert_to_marqo_web_error_and_raise(response: requests.Response, err: reque
     """
     try:
         response_dict = response.json()
+        print("DEBUG: OpenSearch error response: ")
+        pprint.pprint(response_dict)
     except JSONDecodeError:
         raise_catchall_http_as_marqo_error(response=response, err=err)
     if response.status_code == 429:
--- a/src/marqo/tensor_search/constants.py
+++ b/src/marqo/tensor_search/constants.py
@@ -28,5 +28,7 @@ NON_OFFICIAL_LUCENE_SPECIAL_CHARS = {
     ' '
 }
 
-# If the OpenSearch cluster is this percent full, health should return red
-OPENSEARCH_DISK_BREACH_WATERMARK = .85
+# Order is important here: transient, persistent, then defaults.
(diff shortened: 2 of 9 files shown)


src/marqo/_httprequests.py
```diff
@@ -169,9 +167,20 @@ def convert_to_marqo_web_error_and_raise(response: requests.Response, err: reque
             if "limit of total fields" in reason and "exceeded" in reason:
                 raise IndexMaxFieldsError(message="Exceeded maximum number of "
                                                   "allowed fields for this index.")
+        elif open_search_error_type == "cluster_block_exception":
+            raise DiskWatermarkBreachError(message="OpenSearch cluster indexing has been blocked, most likely due to breach \
+                                           of the `flood_stage` disk watermark. For more information, please refer \
```
pandu-k: Nit: backslashes may be unnecessary here

src/marqo/_httprequests.py
```diff
@@ -169,9 +167,20 @@ def convert_to_marqo_web_error_and_raise(response: requests.Response, err: reque
...
+                                           to your cluster settings: `cluster.routing.allocation.disk.watermark` at \
+                                           https://opensearch.org/docs/2.7/api-reference/cluster-api/cluster-settings/")
```
pandu-k: We can also offer a suggestions: try freeing up disk space on your OpenSearch instances.

src/marqo/tensor_search/health.py
```diff
@@ -1,10 +1,108 @@
...
+    if watermark is None:
+        raise errors.InternalError("OpenSearch disk watermark cannot be None.")
+    watermark = watermark.replace(" ", "")
+    if watermark == "":
+        raise errors.InternalError("OpenSearch disk watermark cannot be empty string.")
+    
+    if total_in_bytes is not None and total_in_bytes < 0:
+        raise errors.InternalError("OpenSearch cluster stats fs: total_in_bytes cannot be negative.")
+
+    if watermark[-2:].lower() in constants.BYTE_SUFFIXES:
```
pandu-k: Do we need any validation here? For example if len(watermark) == 1

src/marqo/tensor_search/health.py
```diff
@@ -1,10 +1,108 @@
...
+
+    Returns: red if watermark is breached, green otherwise.
+    """
+
+    # Query opensearch for watermark
+    raw_flood_stage_watermark = None
+    for settings_type in constants.OPENSEARCH_CLUSTER_SETTINGS_TYPES:
+        try:
+            # Check for transient, persistent, then defaults settings
+            opensearch_disk_settings = HttpRequests(config).get(path="_cluster/settings?include_defaults=true&filter_path=**.disk*")\
```
pandu-k: We should move the HTTP request out of the for loop so that it only gets called once

src/marqo/tensor_search/health.py
```diff
@@ -48,6 +146,9 @@ def get_marqo_os_status(config: Config, index_name: Optional[str] = None) -> Uni
     else:
         marqo_os_status = HealthStatuses.red
```
pandu-k, line 154: Nit: In the doc string of this method we can record the different requests to OpenSearch that will be made. This will help us be conscious of the number of OpenSearch requests.

src/marqo/tensor_search/health.py
pandu-k: We can also do make a test asserting only the expected type of calls, and number of calls, to OpenSearch from the health endpoint

src/marqo/_httprequests.py
vicilliar: done

src/marqo/_httprequests.py
vicilliar: done

vicilliar: done

src/marqo/tensor_search/health.py
vicilliar: done

src/marqo/tensor_search/health.py
vicilliar: done

src/marqo/tensor_search/health.py
vicilliar: I've made validation for all the edge cases I can think of. If if `len(watermark) == 1`, this line just returns the whole watermark. For example, `"a"[-2:]` is just `"a"`.

src/marqo/tensor_search/enums.py
```diff
@@ -153,10 +153,10 @@ class HealthStatuses(str, Enum):
     yellow = "yellow"
     red = "red"
 
-    def __lt__(self, other):
+    def __gt__(self, other):
         # Define the custom comparison logic
         status_order = [self.green, self.yellow, self.red]
-        return status_order.index(self) < status_order.index(other)
+        return status_order.index(self) > status_order.index(other)
```
pandu-k: Are we 100% using __gt__ rather than __lt__ won't break @wanliAlex 's changes? Otherwise we can just define both?

src/marqo/tensor_search/health.py
pandu-k: Thanks for clarifying!

src/marqo/tensor_search/health.py
pandu-k: Do you need to push changes? I don't see it updated yet

src/marqo/tensor_search/health.py
pandu-k: Do you need to push changes? I don't see it updated yet

src/marqo/tensor_search/enums.py
vicilliar: defined both!

vicilliar:
All pipelines passing:
image

src/marqo/tensor_search/health.py
```diff
@@ -1,10 +1,108 @@
...
+from marqo.config import Config
+from marqo.tensor_search.tensor_search_logging import get_logger
 from marqo.tensor_search.enums import HealthStatuses
+from marqo.tensor_search import constants
+from marqo.tensor_search import validation
+
+logger = get_logger(__name__)
+
+
+def convert_watermark_to_bytes(watermark: str, total_in_bytes: int = None):
```
farshidz: Add return value type annotation

src/marqo/tensor_search/health.py
```diff
@@ -1,10 +1,108 @@
...
+    
+    if total_in_bytes is not None and total_in_bytes < 0:
+        raise errors.InternalError("OpenSearch cluster stats fs: total_in_bytes cannot be negative.")
+
+    if watermark[-2:].lower() in constants.BYTE_SUFFIXES:
+        # Watermark in KB/MB/GB/TB format
+        # Bytes represent MIN disk space AVAILABLE
+        # TODO: Do we use 1000 or 1024 for conversion?
+        numeric_watermark = validation.validate_nonnegative_number(watermark[:-2], "OpenSearch disk watermark value")
+        multiplier = 1024 ** constants.BYTE_SUFFIXES.index(watermark[-2:].lower())
```
farshidz: I suggest adding a comment in constants on the significance of the order (i.e. index used as the power)

src/marqo/tensor_search/health.py
```diff
@@ -1,10 +1,108 @@
...
+    opensearch_settings = HttpRequests(config).get(path="_cluster/settings?include_defaults=true&filter_path=**.disk*")
+    for settings_type in constants.OPENSEARCH_CLUSTER_SETTINGS_TYPES:
+        try:
+            # Check for transient, persistent, then defaults settings
+            opensearch_disk_settings = opensearch_settings[settings_type]["cluster"]["routing"]["allocation"]["disk"]
+            raw_flood_stage_watermark = opensearch_disk_settings["watermark"]["flood_stage"]
+            logger.debug(f"Found disk flood stage watermark in {settings_type} settings: {raw_flood_stage_watermark}")
+            break
+        except KeyError:
+            logger.debug(f"Flood stage watermark not found in {settings_type} settings.")
```
farshidz: Is this expected to happen under normal circumstances? Otherwise you should raise it

src/marqo/tensor_search/validation.py
```diff
@@ -621,3 +621,27 @@ def validate_delete_docs_request(delete_request: MqDeleteDocsRequest, max_delete
 
     return delete_request
 
+
+def validate_nonnegative_number(input_string: str, field_description: str = "Input"):
+    """Validates that a string is a non-negative number
+
+    Args:
+        input_string: the string to validate
```
farshidz, line 629: add docstring for field_description

src/marqo/_httprequests.py
```diff
@@ -169,9 +167,20 @@ def convert_to_marqo_web_error_and_raise(response: requests.Response, err: reque
...
+                                           "of the `flood_stage` disk watermark. Try freeing up disk space on your OpenSearch instances. "
+                                           "For more information, please refer to cluster settings: `cluster.routing.allocation.disk.watermark` at "
+                                           "https://opensearch.org/docs/2.7/api-reference/cluster-api/cluster-settings/")
     except KeyError:
         pass
 
+    # for throttling
+    if response.status_code == 429:
+        raise TooManyRequestsError(
+            message="Marqo-OS received too many requests! "
```
farshidz, line 181: We refer to both OpenSearch as both OpenSearch and also Marqo-OS. Perhaps beyond the scope of this PR, but it's confusing

tests/tensor_search/test_healthcheck.py
```diff
@@ -17,38 +18,776 @@ def tearDown(self) -> None:
...
+            nonlocal mock_health, mock_settings, mock_stats
+            # mock settings should return GREEN, flood stage watermark NOT breached
+            # Inserting default mocks and also making deep copies so that the original mocks aren't modified each run
+            returned_mock_health = copy.deepcopy(mock_health) if mock_health is not None else {'status': 'green'}
+            returned_mock_settings = copy.deepcopy(mock_settings) if mock_settings is not None else {
+                "transient": {
+                    "cluster": {"routing": {"allocation": {"disk": {"watermark": {"flood_stage": "95%"}}}}}
+                },
+                "persistent": {},
+                "defaults": {}
```
farshidz, line 39: Can we have a test that has both all of defaults, transient and persistent, and make sure if honours the priority order? Perhaps also one where we have only defaults and persistent

tests/tensor_search/test__httprequests.py
```diff
@@ -76,8 +76,8 @@ def run():
                     )
                 )
                 raise AssertionError
-            except DiskWatermarkBreachError:
-                pass
+            except DiskWatermarkBreachError as e:
+                print(e.message)
```
pandu-k: nit: remove print statement

pandu-k (approved):

src/marqo/tensor_search/constants.py
```diff
@@ -27,3 +27,8 @@
 NON_OFFICIAL_LUCENE_SPECIAL_CHARS = {
     ' '
 }
+
+# Order is important here: transient, persistent, then defaults.
+OPENSEARCH_CLUSTER_SETTINGS_TYPES = ["transient", "persistent", "defaults"]
+
+BYTE_SUFFIXES = ['b', 'kb', 'mb', 'gb', 'tb', 'pb', 'eb', 'zb', 'yb']
```
farshidz: This is really elegant, but at the same time the use of index as exponent worries me a bit, since if we later changed this list for any reason, the bug introduced would be really hard to catch (we'd get no error, just unexpected behaviour). I suggest replacing this with a `dict` even though the code won't look as nice as your approach.

farshidz (changes requested):

src/marqo/tensor_search/health.py
vicilliar: Yes, this is actually just a debug message, it will indicate every time watermark is not found (let's say in transient or persistent settings), then it will just check the next level of settings. I can remove this debug statement.

src/marqo/_httprequests.py
vicilliar: Will make a ticket for this!

tests/tensor_search/test_healthcheck.py
vicilliar: Yep, actually I have test cases for every combination of settings, as well as the watermark being in none/all of them. You can check them out here: https://github.com/marqo-ai/marqo/blob/joshua/opensearch-disk-breach/tests/tensor_search/test_healthcheck.py#L375

src/marqo/tensor_search/health.py
vicilliar: changed the byte suffixes to a dict instead

src/marqo/tensor_search/constants.py
vicilliar: done, implemented your suggestion!

farshidz (approved):

vicilliar:
Manual health check and Pipelines passing:
image

PR #562, opened 2023-08-02
Use a dummy search request to replace zero vector search

wanliAlex:
Use a dummy search query to replace a zero-vector search query in vector search.

A zero-vector is a vector that all the elements are 0s, e.g., `[0.0, 0.0, 0.0]`.

we send the zero vectors to Marqo-OS as a normal knn search. However, this will result in an error in Marqo-OS when `number_of_shards` is set to a specific value.

For example, sometimes an error is returned from Marqo-OS. Sometimes a 0 hits search result is returned in Marqo-OS.

We use a `match none` search query to replace the zero-vector search query so Marqo-OS will return consistent results (0 hits) regardless of `number_of_shards`

no

* **Other information**:


pandu-k (approved):

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1746,18 +1750,17 @@ def _vector_text_search(
...
-                for vec in list(q["query"]["nested"]["query"]["knn"].keys()):
-                    readable_body[i]["query"]["nested"]["query"]["knn"][vec]["vector"] = \
-                        readable_body[i]["query"]["nested"]["query"]["knn"][vec]["vector"][:5]
+                try:
+                    for vec in list(q["query"]["nested"]["query"]["knn"].keys()):
+                        if "vector" in q["query"]["nested"]["query"]["knn"][vec]:
+                            readable_body[i]["query"]["nested"]["query"]["knn"][vec]["vector"] = \
+                                readable_body[i]["query"]["nested"]["query"]["knn"][vec]["vector"][:5]
+                except KeyError:
+                    pass
```
farshidz: When does this happen?

wanliAlex: This will happen when `verbose == 1` and we use a dummy query to replace the original query. 

I am not exactly sure about the best solution here, try except, or do`if "key" in query` check?

farshidz: An `if` check would be much better. Swallowing an exception is bad practice and can cause difficult to debug issues.

farshidz (approved):

PR #569, opened 2023-08-10
Reduce Docker layers

pandu-k:
Dockerfile optimisation

The Dockerfile narrates a build process that results in a Docker image with many layers

The Dockerfile now specifies a build process that results in fewer layers, grouping together the dependency setup code and separating it from the steps that prepare the Marqo application code. 

No

Running API tests against this Dockerfile


pandu-k:
Reducing the number of run commands has reduced layers during download:
image
Ideas for further optimisation:
1. Perhaps we can combine the run commands for the following into a single command:
```
COPY requirements.txt requirements.txt
RUN pip3 install --no-cache-dir -r requirements.txt

COPY scripts scripts
RUN bash scripts/install_onnx_gpu_for_amd.sh && \
    bash scripts/install_torch_amd.sh && \
    # redis installation for throttling
    bash scripts/install_redis.sh && \
    # redis config lines
    echo "echo never > /sys/kernel/mm/transparent_hugepage/enabled" >> /etc/rc.local && \
    echo "save ''" | tee -a /etc/redis/redis.conf

COPY dind_setup dind_setup
RUN bash dind_setup/setup_dind.sh
```
2. restructure the  `dind_setup/` to be inside `scripts/`, to get rid of one COPY command

pandu-k:
image

After using the marqo_base registry ^

We end up with 14 layers (as opposed to 12 when using a single repo, but combining run steps).

Marqo build time has been reduced to 6.5 min.

Can also confirm that Marqo starts as expected?

Dockerfile
```diff
@@ -1,42 +1,48 @@
...
-RUN apt-get install unzip -y
-RUN mkdir -p /root/nltk_data/tokenizers
-RUN curl https://raw.githubusercontent.com/nltk/nltk_data/gh-pages/packages/tokenizers/punkt.zip -o /root/nltk_data/tokenizers/punkt.zip
-RUN unzip /root/nltk_data/tokenizers/punkt.zip  -d /root/nltk_data/tokenizers/
-
-# TODO: up the RAM
-RUN echo Target platform is "$TARGETPLATFORM"
 COPY requirements.txt requirements.txt
 RUN pip3 install --no-cache-dir -r requirements.txt
+
```
farshidz, line 8: Do we still need everything in `scripts` in this image?

Dockerfile
```diff
@@ -1,42 +1,48 @@
-ARG CUDA_VERSION=11.4.3
-FROM nvidia/cuda:${CUDA_VERSION}-cudnn8-runtime-ubuntu20.04 as cuda_image
-FROM ubuntu:20.04
+# ARG CUDA_VERSION=11.4.3
+# FROM nvidia/cuda:${CUDA_VERSION}-cudnn8-runtime-ubuntu20.04 as cuda_image
+# FROM ubuntu:20.04 as base_image
+FROM marqoai/marqo-base:test as base_image
 VOLUME /var/lib/docker
```
farshidz, line 2: What is this for?

pandu-k: This is allows Docker-in-Docker to work. It also allows Marqo to transfer its state from container to container

pandu-k:
After modifying the marqo-base image, it seems to have re-increased the number of Marqo layers: 
image

pandu-k:
Log: Did a DIND performance test, using CUDA, against 1.2.0 and saw no regression.

pandu-k:
Also did arm64, local OS API tests with no failures

pandu-k:
Did the volume transfer check (between 1.2.0 and test tag) and Marqo behaved as expected.

pandu-k:
Test suites running:
- arm64: https://github.com/marqo-ai/marqo/actions/runs/5921786617
- CUDA DIND: https://github.com/marqo-ai/marqo/actions/runs/5921789878
- DIND: https://github.com/marqo-ai/marqo/actions/runs/5921791422
- Large model unit tests: https://github.com/marqo-ai/marqo/actions/runs/5923989099
- Local OS: https://github.com/marqo-ai/marqo/actions/runs/5921793274
- Unit tests: https://github.com/marqo-ai/marqo/actions/runs/5923988025

.github/workflows/s2search_CI.yml
pandu-k, line 1: taking this opportunity to remove this deprecated test workflow

.github/workflows/unit_test_CI.yml
pandu-k, line 1: This workflow has been superseded by the unit_test_200GB workflow. Taking this opportunity to remove this workflow.

Dockerfile
pandu-k: Thanks - removed unnecessary  kept necessary scripts.

scripts/dind_setup/setup_dind.sh
pandu-k, line 1: Docker-in-Docker setup still needs to be done in final image

.github/workflows/largemodel_unit_test_CI.yml
```diff
@@ -35,66 +35,51 @@ jobs:
     name: Run Marqo Test Suite
     needs: Start-Runner # required to start the main job when the runner is ready
     runs-on: ${{ needs.start-runner.outputs.label }} # run the job on the newly created runner
-
-    environment: marqo-test-suite
-
+
```
farshidz: There seems to be extra whitespace in this file. Reformatting with your IDE should clean it up.

requirements.txt
```diff
@@ -1,44 +1,10 @@
-# tensor search:
+# This is a reduced set of requirements used for Marqo's tensor search module
+# If you are running this on your local machine you will need to install additional
+# requirements specified here: https://github.com/marqo-ai/marqo-base/blob/main/requirements.txt
```
farshidz, line 3: This will be a bit difficult to manage. It might confuse the IDE (I have tested though). Also when checking out an old commit we don't really know which `marqo-base` commit corresponds to this for the right version.

How about including a requirements.dev.txt in the repo with all dependencies listed?

Dockerfile
```diff
@@ -1,42 +1,15 @@
...
-# redis installation for throttling
-RUN bash scripts/install_redis.sh
-# redis config lines
-RUN echo "echo never > /sys/kernel/mm/transparent_hugepage/enabled" >> /etc/rc.local
-RUN echo "save ''" | tee -a /etc/redis/redis.conf
-COPY dind_setup dind_setup
-RUN bash dind_setup/setup_dind.sh
+RUN bash scripts/dind_setup/setup_dind.sh
+
+FROM base_image
```
farshidz, line 12: I'm guessing this is superfluous and won't change the final image.

farshidz (changes requested):

requirements.txt
pandu-k: Good point. Added the new file and updated dev README

pandu-k: Also, keep in mind that if you needed to, you can trace the marqo-base version via the marqo-base number (but it isn't super straightforward)

.github/workflows/largemodel_unit_test_CI.yml
pandu-k: Thanks. Removed.

Dockerfile
pandu-k: If this is simply superfluous do you mind if we keep this as-is for now? The current image went through quite a long testing process, and I do vaguely recall that removing this caused some issues (though I could be misremembering)

farshidz: Fair enough. Yes I don't think it'll have any major effect (good or bad). Let's keep it for now then.

farshidz (approved):

PR #575, opened 2023-08-17
Update m2 developer instructions

pandu-k:
Docs update. As a developer with an M2 Mac, the dev setup instructions don't currently work. 

 M2 Mac, the dev setup instructions updated to install ONNX and protobuf properly.

### DIFF
--- a/src/marqo/README.md
+++ b/src/marqo/README.md
@@ -139,7 +139,7 @@ onnx
 protobuf
 ```
 
-9. Install marqo dependencies. 
+8. Install marqo dependencies. 
 ```
 pip install -r requirements.txt
 ```
@@ -149,7 +149,8 @@ pip install -r requirements.txt
 CWD=$(pwd)
 cd src/marqo/tensor_search/
 ```
-9. Run Marqo,
+
+10. Run Marqo,
 ```
 export OPENSEARCH_URL="https://localhost:9200" && 
     export PYTHONPATH="${PYTHONPATH}:${CWD}/src" &&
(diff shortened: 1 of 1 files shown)


src/marqo/README.md
```diff
@@ -133,17 +133,24 @@ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs/ | sh;
 
 6. Set up redis (follow instructions in Option A)
 
-7. Install marqo dependencies,
+7. If you are using an M2 Mac specifically, then edit `requirements.txt`. Find the `onnx` and `protobuf` entries, and remove their versions. Then, move onnx to be above protobuf. These two lines should end up looking like this:
```
vicilliar: the issue with this is their future PRs will have the versions removed as well. Maybe a more permanent fix for this would be to have a separate script for installing `onnx` and `protobuf` for M2, similar to this: https://github.com/marqo-ai/marqo/blob/mainline/scripts/install_onnx_gpu_for_amd.sh

vicilliar (changes requested):

src/marqo/README.md
pandu-k: Good point. We don't want them to accidentally git commit the wrong version of things.

vicilliar: Or we could encourage them to make a new `m2_requirements.txt` and `.gitignore` it, so the original requirements file stays unchanged.

pandu-k: I don't want to be to prescriptive in this sense. If that is a pattern of usage we see emerge (m2-using devs will really want to have their own dedicated requirements file) then we can support it, but I don't want to do this is if we don't have to. These are also devs, so they should be able to figure out a pattern that works for them.

There is also the gitignored `local_only/` dir that they can use

vicilliar: Fair point. We can be less prescriptive in what to do then, as long as they upgrade `onnx` and `protobuf`.

PR #601, opened 2023-09-12
Set add docs default refresh to false

vicilliar:
Feature

The default value for `add_documents` parameter `refresh` is true.

  - API
    - The default value for `add_documents` parameter `refresh` is now false.
    - The default value for `delete_docs` parameter `refresh` is now false.
  - API UTILS
    - The default value for `auto_refresh` in orchestrator is now false.
  - README
    - added explanation for when to set `auto_refresh`.
  - EXAMPLES
    - added `auto_refresh=True` to add docs calls with small number of documents.

Yes. To keep old behavior, users will have to set `refresh` to true manually.

Yes

[In progress](https://github.com/marqo-ai/py-marqo/pull/175)

[In progress](https://github.com/marqo-ai/marqodocs/commit/ac4fc02efb4989821c4ef6b1bd6a80bb25b08d60)

* **Other information**:


examples/MultiLingual/article.md
```diff
@@ -83,7 +83,7 @@ Here we set the device argument as `"cuda"`. This tells Marqo to use the GPU it
 If you don't have a GPU, remove this argument or set it to `"cpu"`. We encourage using a GPU as it will make the `add_documents` 
 process significantly faster (our testing showed a 6–12x speed up). 
 
-We also set the `auto_refresh` argument to `False`. When indexing large volumes of data we encourage you to set this to False, as it optimises the `add_documents` process. 
+We do not set the `auto_refresh` argument, so it defaults to `False`. When indexing large volumes of data we encourage you to keep this `False``, as it optimises the `add_documents` process.
```
pandu-k: it looks like False has an extra backtick: 
```
`False``
```

src/marqo/tensor_search/api.py
```diff
@@ -172,7 +172,7 @@ def add_or_replace_documents(
         request: Request,
         body: typing.Union[AddDocsBodyParams, List[Dict]],
         index_name: str,
-        refresh: bool = True,
+        refresh: bool = False,
```
pandu-k, line 175: In this case, is it possible to move the default for this value into the `AddDocsParams`?

pandu-k: (but no need to consolidate other add_docs params)

tests/tensor_search/test_api.py
```diff
@@ -90,3 +90,60 @@ def test_add_or_replace_documents_fields_undefined_query_param(self):
             )
             self.assertEqual(response.status_code, 400)
             mock_add_documents.assert_not_called()
+    
+    def test_add_or_replace_documents_defaults(self):
```
pandu-k, line 94: this is a good test. nice

pandu-k, line 94: is it also worth testing that manually setting `auto_refresh=True` and `auto_refresh=False`, propagates this value in the expected way?

tests/tensor_search/test_api.py
```diff
@@ -90,3 +90,60 @@ def test_add_or_replace_documents_fields_undefined_query_param(self):
...
+            self.assertEqual(response.status_code, 200)
+            mock_add_documents.assert_called_once()
+            args, kwargs = mock_add_documents.call_args
+
+            # Assert that add documents is called with the correct default arguments
+            assert kwargs["add_docs_params"].auto_refresh == False
+            assert kwargs["add_docs_params"].use_existing_tensors == False
+
+
+class ApiTestsDeleteDocs(MarqoTestCase):
```
pandu-k, line 149: ditto on my extra test suggestions on `test_add_or_replace_documents_defaults()`

tests/tensor_search/test_api_utils.py
```diff
@@ -113,6 +113,31 @@ def test_add_docs_params_orchestrator(self):
         assert result.docs == [{"test": "doc"}]
         assert result.image_download_headers == {"header1": "value1"}
 
+    def test_add_docs_params_orchestrator_defaults(self):
+        """
+        Ensures that certain defaults are correct when not specified upon calls to 
+        add_docs_params_orchestrator (like use_existing_tensors, auto_refresh, etc)
+        """
```
pandu-k, line 120: thanks for proactively filling gaps in our current set of tests

tests/tensor_search/test_api.py
vicilliar: yep, will update this.

tests/tensor_search/test_api_utils.py
vicilliar: no worries! this will be a good template to build upon when future parameters/features are implemented.

examples/MultiLingual/article.md
vicilliar: great catch, thanks

src/marqo/tensor_search/api.py
vicilliar: definitely

vicilliar: Actually, the only way I see this happening is by setting the default for the `refresh` parameter here to `None`, passing that to the orchestrator, then handling logic for `auto_refresh is None` inside `addDocsParams`. It feels a bit roundabout.

Another way to do it is to add logic to the API and the orchestrator to not send the `refresh` parameter at all if not provided by the API call, so that `addDocsParams` fills it in by itself, but that goes against our principle of having less logic in the API.

src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -41,7 +41,11 @@ class AddDocsParamsConfig:
 
 
 class AddDocsBodyParams(BaseModel):
-    """The parameters of the body parameters of tensor_search_add_documents() function"""
+    """
+    Representation of the body parameters of the API add_or_replace_documents() function.
```
vicilliar, line 45: Revamped the description for this class, as the old one was vague and confusing. Feedback would be appreciated on the clarity of the new docstring!

pandu-k: I'd prefer keeping this class and `tensor_search.add_documents()` decoupled from the API layer. `tensor_search.add_documents()`'s logic should be independent of whatever API (including a non REST protocols, such as gRPC) we choose.

vicilliar: But isn't that the point of this class? To my understanding, it's called `AddDocsBodyParams` because it represents the parameters of the post request body. It then internally gets converted to a `AddDocsParams`.

pandu-k: the point of this class was to consolidate the `add_documents` parameters, it wasn't meant to represent the body of post request. 

This class was a thing before the post request configuration was moved from the query string param to the body

pandu-k (approved):

PR #610, opened 2023-10-04
Custom Vector field type for add_documents

vicilliar:
Feature

Marqo does not accept vectors directly input into documents.

    - Main change:
        - New field type: "custom_vector" implemented, allowing for vectors (with optional content attached) to be directly added as a field. 

```
# Mapping format
mappings={
    "my_custom_vector_field": {
        "type": "custom_vector"
    }
}

# Document field format
document={
  "my_custom_vector_field": {
      "content": "Going to the shops",
      "vector": [0.141, 1.414, ..., -1.4, 0.1]
  },
}
```
    - Minor changes:
        - Refactor: separate function to create chunk metadata
        - Refactor: separate function to determine document field type (standard, multimodal, custom vector)
        - Removed/combined redundant mappings and multimodal mappings validation functions
        - Multimodal fields use `field_chunks_to_append` instead of adding to `doc_chunks` directly (for uniformity)
        - Chunk metadata ended at the end of document processing (since custom vector logic changes content of document)
        - `validate_dict` tests moved to `test_validation.py` file.

No. Old field types should work exactly the same.

In progress.

In progress.

In progress.

* **Other information**:


src/marqo/tensor_search/add_docs.py
```diff
@@ -182,4 +183,46 @@ def reduce_thread_metrics(data):
...
+    metadata = {}
+    metadata_field_types = {str, bool, int, float, list, dict}
+    for key, value in raw_document.items():
+        if type(value) in metadata_field_types:
+            metadata[key] = value
+
+    return metadata
+
+
+def determine_document_field_type(field_name, field_content, mappings):
```
wanliAlex: nit: type for these parameters?

src/marqo/tensor_search/add_docs.py
```diff
@@ -182,4 +183,46 @@ def reduce_thread_metrics(data):
...
+
+    3 Options:
+    1. standard (str, int, float, bool, list)
+    2. multimodal_combination (dict)
+    3. custom_vector (dict)
+    """
+
+    if isinstance(field_content, dict):
+        if field_name not in mappings:
+            raise errors.InternalError(f"Invalid dict field {field_name}. Could not find field in mappings object.")
```
wanliAlex, line 221: Should this be an internal error or should this be a user side error? Maybe you did the validation before so it should be an internal error here. Just to confirm with you.

src/marqo/tensor_search/validation.py
```diff
@@ -501,7 +510,66 @@ def validate_multimodal_combination(field_content, is_non_tensor_field, field_ma
...
+    
+    # Must be a tensor_field
+    if is_non_tensor_field:
+        raise InvalidArgError(
+            f"Cannot create custom_vector field (given field content: `{field_content}`) as a non-tensor field. "
+            f"Add this field to `tensor_fields` to fix this problem."
+        )
+    
+    try:
+        custom_vector_schema = {
```
wanliAlex: Should we move this to models.mappings_object module?

src/marqo/tensor_search/validation.py
```diff
@@ -501,7 +510,66 @@ def validate_multimodal_combination(field_content, is_non_tensor_field, field_ma
...
+    Args:
+        field_content: the field content
+        is_non_tensor_field: whether this is a non-tensor-field
+        index_model_dimensions: the `dimensions` property of the index to be added to
+
+    Returns:
+        field_content if the validation passes
+        "content" key will be added to the field_content as empty string if it is not provided.
+    '''
+    if index_model_dimensions is None:
```
wanliAlex: maybe check if the dimension is a `int`
```suggestion
if not isinstance(index_model_dimension, int): 
```

wanliAlex (changes requested):
Well done Joshua!

src/marqo/tensor_search/add_docs.py
vicilliar: Done, but omitted for field_content, as there is no straightforward way to use `Union` with `constants.ALLOWED_CUSTOMER_FIELD_TYPES`. I also don't want to hardcode the list.

src/marqo/tensor_search/add_docs.py
vicilliar: Leaving this as is. But adding `InvalidArgError` checks in `validate_dict`, which will be called before this.

src/marqo/tensor_search/validation.py
vicilliar: I would prefer not to, since this is validation for the custom vector object, not its mapping. The custom vector mapping schema is already in the `models.mappings_object`

src/marqo/tensor_search/add_docs.py
```diff
@@ -182,4 +184,46 @@ def reduce_thread_metrics(data):
...
+    Creates a chunk metadata dictionary for a given document.
+    This metadata will be put in each OpenSearch child document (chunk) to be used for filtering.
+
+    We will only add values which are string, boolean, int, float, list or dictionary.
+    """
+
+    metadata = {}
+    metadata_field_types = {str, bool, int, float, list, dict}
+    for key, value in raw_document.items():
+        if type(value) in metadata_field_types:
```
pandu-k: It might be worth looking into `type(blah)` vs `isintance(blah)` as they may different characteristics, especially around inheritance - just worth double checking/quickly reading/testing on

src/marqo/tensor_search/add_docs.py
```diff
@@ -182,4 +184,46 @@ def reduce_thread_metrics(data):
...
+def determine_document_field_type(field_name: str, field_content, mappings: dict):
+    """
+    Determines the type of a document field
+    using its name, content, and the add docs mappings object.
+
+    3 Options:
+    1. standard (str, int, float, bool, list)
+    2. multimodal_combination (dict)
+    3. custom_vector (dict)
+    """
```
pandu-k, line 217: awesome docstring

src/marqo/tensor_search/add_docs.py
```diff
@@ -182,4 +184,46 @@ def reduce_thread_metrics(data):
...
+    metadata = {}
+    metadata_field_types = {str, bool, int, float, list, dict}
+    for key, value in raw_document.items():
+        if type(value) in metadata_field_types:
+            metadata[key] = value
+
+    return metadata
+
+
+def determine_document_field_type(field_name: str, field_content, mappings: dict):
```
pandu-k: nit: could be good having the return type in the signature, too

src/marqo/tensor_search/add_docs.py
```diff
@@ -182,4 +184,46 @@ def reduce_thread_metrics(data):
                     result[new_key] = [result[new_key], value]
             else:
                 result[new_key] = value
-    return result
\ No newline at end of file
+    return result
+
+
+def create_chunk_metadata(raw_document: dict):
```
pandu-k: nit: ditto on the return type

src/marqo/tensor_search/enums.py
```diff
@@ -119,9 +119,15 @@ class RequestType:
     DELETE = "DELETE"
     CREATE = "CREATE"
 
+# Each type has different add_documents behavior. More can be added to represent special types.
```
pandu-k, line 122: thanks for comment!

src/marqo/tensor_search/enums.py
```diff
@@ -119,9 +119,15 @@ class RequestType:
...
+class DocumentFieldType:
+    standard = "standard"       # str, int, float, bool, or list
+    multimodal_combination = "multimodal_combination"   # dict
+    custom_vector = "custom_vector"     # dict
 
-class MappingsObjectType:
+class MappingsObjectType(str, Enum):
     multimodal_combination = "multimodal_combination"
+    custom_vector = "custom_vector"
```
pandu-k, line 130: nit: alternatively you could make MappingsObjectType subclass DocumentFieldType. Then you may get some nice polymorphism properties. But not too concerned if Python doesn't play nice with this arrangement.

src/marqo/tensor_search/models/mappings_object.py
```diff
@@ -46,3 +48,15 @@
...
+  "type": "object",
+  "properties": {
+    "type": {
+      "type": "string",
+      "enum": [MappingsObjectType.custom_vector]
+    }
+  },
+  "required": ["type"],
+  "additionalProperties": False
+}
```
pandu-k, line 62: FYI: we are moving towards PyDantic validation over JsonSchema, but this is a long term goal, and would not be worth rewriting this code if it is a lot of effort.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -482,22 +470,52 @@ def add_documents(config: Config, add_docs_params: AddDocsParams):
...
 
-                # chunks generated by processing this field for this doc:
-                chunks_to_append = []
-                # Check if content of this field changed. If no, skip all chunking and vectorisation
-                if (add_docs_params.use_existing_tensors
+                # 4 current options for chunking/vectorisation behavior:
+                # A) field type is custom_vector -> no chunking or vectorisation
+                # B) use_existing_tensors=True and field content hasn't changed -> no chunking or vectorisation
+                # C) field type is standard -> chunking and vectorisation
+                # D) field type is multimodal -> use vectorise_multimodal_combination_field (does chunking and vectorisation)
```
pandu-k, line 494: this comments are helpful. In the future we can structure the code to communicate this:

```
if field type is custom_vector:
    result = no_chunking_or_vectorisation( ... )
elif use_existing_tensors=True and field_content_not_changed:
     result = no_chunking_or_vectorisation( ... )
 ...
```

src/marqo/tensor_search/tensor_search.py
```diff
@@ -596,43 +614,51 @@ def add_documents(config: Config, add_docs_params: AddDocsParams):
...
+                # Add field_chunks_to_append to total document chunk list
+                for chunk in field_chunks_to_append:
+                    doc_chunks.append(chunk)
 
             if document_is_valid:
                 new_fields = new_fields.union(new_fields_from_doc)
-                copied[TensorField.chunks] = chunks
+
+                # Create metadata to put in doc chunks (from altered doc)
+                chunk_values_for_filtering = add_docs.create_chunk_metadata(raw_document=copied)
```
pandu-k, line 656: What was the decision behind using the copied doc rather than the original one?

src/marqo/tensor_search/validation.py
```diff
@@ -427,57 +431,65 @@ def validate_settings_object(settings_object):
...
 
     Args:
         field: the field name
         field_content: the field when it is a dict, especially used for multimodal tensor combination field
         is_non_tensor_field: for multimodal tensor combination field, this should be True
         mappings: a dictionary to help validate the object field content
+        index_model_dimensions: the dimensions of the model of the index. used to validate custom vector field.
 
     Returns:
         True or raise an error
```
pandu-k: This Return value seems to be outdated

src/marqo/tensor_search/validation.py
```diff
@@ -427,57 +431,65 @@ def validate_settings_object(settings_object):
...
-        validate_multimodal_combination(field_content, is_non_tensor_field, mappings[field])
+        field_content = validate_multimodal_combination(field_content, is_non_tensor_field, mappings[field])
+    elif mappings[field]["type"] == "custom_vector":
+        field_content = validate_custom_vector(field_content, is_non_tensor_field, index_model_dimensions)
+    else:
+        raise InvalidArgError(
+            f"The field {field} is of invalid type in the `mappings` parameter. The only object field types supported "
+            f"are `multimodal_combination` and `custom_vector. However, the `mappings` is {mappings}. Please change the "
+            f"type of {field} to one of these."
+            f"See `https://marqo.pages.dev/1.4.0/API-Reference/mappings/` for more info on object fields. "
```
pandu-k: nit. this will be printed as `"... one of these.See ..."`

src/marqo/tensor_search/validation.py
```diff
@@ -427,57 +431,65 @@ def validate_settings_object(settings_object):
...
         field_mapping: the mapping to help validate this field content
 
     Returns:
-
+        The field content
     '''
     if len(field_content) < 1:
         raise InvalidArgError(
-            f"The multimodal_combination_field `{field_content}` is an empty dictionary. "
+            f"The multimodal_combination field `{field_content}` is an empty dictionary. "
```
pandu-k, line 490: nice update 👍

src/marqo/tensor_search/validation.py
pandu-k: This schema is may be worth moving to into it's own file as it specify a logical model. Perhaps `models.user_vector`

src/marqo/tensor_search/validation.py
```diff
@@ -501,7 +513,67 @@ def validate_multimodal_combination(field_content, is_non_tensor_field, field_ma
...
+                    }
+                }
+            },
+            "required": ["vector"],     # content is optional.
+            "additionalProperties": False
+        }
+        jsonschema.validate(instance=field_content, schema=custom_vector_schema)
+    except jsonschema.ValidationError as e:
+        raise InvalidArgError(
+            f"Invalid custom_vector field format. Reason: \n{str(e)}"
```
pandu-k, line 549: nit: could be worth adding a space after the reason.

src/marqo/tensor_search/validation.py
```diff
@@ -501,7 +513,67 @@ def validate_multimodal_combination(field_content, is_non_tensor_field, field_ma
...
+            f"\nFor info on how to use custom_vector, please see: `https://docs.marqo.ai/1.4.0/Advanced-Usage/document_fields/#custom-vectors`"
+        )
+
+    # Fill in default content as empty string if not provided.
+    if "content" not in field_content:
+        field_content["content"] = ""
```
pandu-k, line 555: is this OK if the content type is an int or a float? OpenSearch might throw an error if the content field is already defined as an int field, for example

pandu-k: It looks like the type of a custom vector's content can be anything (unless I am misreading)

src/marqo/tensor_search/validation.py
```diff
@@ -523,67 +598,69 @@ def validate_mappings_object(mappings_object: Dict):
...
     try:
-        jsonschema.validate(instance=multimodal_mappings, schema=multimodal_combination_schema)
-        return multimodal_mappings
+        jsonschema.validate(instance=mappings_object, schema=multimodal_combination_mappings_schema)
     except jsonschema.ValidationError as e:
         raise InvalidArgError(
-            f"Error validating multimodal combination object. Reason: \n{str(e)}"
-            f"\nRead about the mappings object here: https://docs.marqo.ai/0.0.15/API-Reference/mappings/"
+            f"Error validating multimodal combination mappings object. Reason: \n{str(e)}"
+            f"\nRead about the mappings object here: https://docs.marqo.ai/1.4.0/API-Reference/mappings/"
```
pandu-k: nit: ditto on space after reason

src/marqo/tensor_search/validation.py
```diff
@@ -523,67 +598,69 @@ def validate_mappings_object(mappings_object: Dict):
...
 
+def validate_custom_vector_mappings_object(mappings_object: Dict):
+    """Validates the multimodal mappings object
+
+    Args:
+        mappings_object:
+
+    Returns:
+        The original object, if it passes validation
+    Raises InvalidArgError if the object is badly formatted
```
pandu-k, line 629: docstring seems unrelated

pandu-k: (at least the line about multimodal mappings)

src/marqo/tensor_search/validation.py
```diff
@@ -523,67 +598,69 @@ def validate_mappings_object(mappings_object: Dict):
...
+    }
+    """
+    try:
+        jsonschema.validate(instance=mappings_object, schema=custom_vector_mappings_schema)
+    except jsonschema.ValidationError as e:
         raise InvalidArgError(
-            f"The multimodal_combination mapping `{field_mapping}` does not contain `weights`"
-            f"Please check `https://docs.marqo.ai/0.0.15/Advanced-Usage/document_fields/#multimodal-combination-object` for more info."
+            f"Error validating custom vector mappings object. Reason: \n{str(e)}"
+            f"\nRead about the mappings object here: https://docs.marqo.ai/1.4.0/API-Reference/mappings/"
```
pandu-k: on the new documentation, it looks like the URL path is this: `new-marqo-docs/API-Reference/Documents/mappings/`

tests/tensor_search/test_custom_vector_field.py
```diff
@@ -0,0 +1,1061 @@
...
+            )
+        )
+
+        # Confirm get_document_by_id returns correct content
+        res = tensor_search.get_document_by_id(
+            config=self.config, index_name=self.index_name_1,
+            document_id="0", show_vectors=True)
+        
+        # Check content is correct
+        print(f"DEBUG: res: {res}")
```
pandu-k: nit: rm print statement

tests/tensor_search/test_multimodal_tensor_combination.py
```diff
@@ -62,13 +62,6 @@ def test_add_documents(self):
         tensor_search.add_documents(config=self.config, add_docs_params=AddDocsParams(
             index_name=self.index_name_1, docs=[
                 expected_doc,
-                # this is just a dummy one
-                {
-                    "Title": "Horse rider",
-                    "text_field": "A rider is riding a horse jumping over the barrier.",
-                    "image_field": "https://marqo-assets.s3.amazonaws.com/tests/images/image1.jpg",
-                    "_id": "1"
-                },
```
pandu-k: why was this removed?

pandu-k (commented):
.

tests/tensor_search/test_add_documents.py
```diff
@@ -1371,3 +1371,83 @@ def test_params_validation_non_tensor_fields_only(self):
 
         assert params.tensor_fields is None
         assert params.non_tensor_fields == ["field1", "field2"]
+
+
+class TestAddDocumentsUtils(MarqoTestCase):
+    def test_create_chunk_metadata(self):
```
pandu-k, line 1377: It is worth having a test to ensure that the error message renders into a string properly (i.e. `abc.__name__` gets interpreted correctly

src/marqo/tensor_search/enums.py
```diff
@@ -119,9 +119,15 @@ class RequestType:
     DELETE = "DELETE"
     CREATE = "CREATE"
 
+# Each type has different add_documents behavior. More can be added to represent special types.
+class DocumentFieldType:
```
pandu-k, line 123: Do we have tests about whether these are translate-able into strings, if that is desired behaviour?

src/marqo/tensor_search/add_docs.py
vicilliar: Did some research on it, the difference is just that `type` ignores inheritance, so any subclass of str, bool, int, etc, won't be considered. I'll change this back to `isinstance` to keep parity with our old code.

src/marqo/tensor_search/add_docs.py
vicilliar: thanks! :)

src/marqo/tensor_search/validation.py
vicilliar: No, the content vector's content can only be str at the moment. The allowed types are defined in constants here:
https://github.com/marqo-ai/marqo/blob/joshua/add-docs-custom-vectors/src/marqo/tensor_search/constants.py#L21
as `ALLOWED_CUSTOM_VECTOR_CONTENT_TYPES`.

src/marqo/tensor_search/validation.py
vicilliar: thank you!

src/marqo/tensor_search/validation.py
vicilliar: sorry, what do you mean by this? The next line starts with a `\n` as well.

src/marqo/tensor_search/validation.py
vicilliar: nice catch, fixed!

src/marqo/tensor_search/validation.py
vicilliar: Will it actually be called `new-marqo-docs`? Or will it just be `https://marqo.pages.dev/1.4.0-n/API-Reference/Documents/mappings/`

src/marqo/tensor_search/enums.py
vicilliar: No, but that isn't really their intended behavior. It is just meant to be an Enum to classify the 3 types of fields cleanly, and so far is only used internally (inside add docs)

src/marqo/tensor_search/models/mappings_object.py
vicilliar: Got it, I'd say leave this is as in interest of time. We can come back for the refactor later on.

src/marqo/tensor_search/tensor_search.py
vicilliar: We were already doing this originally, I just kept it for the sake of parity. Not exactly sure why we use the copied document for metadata

src/marqo/tensor_search/tensor_search.py
vicilliar: got this!

src/marqo/tensor_search/validation.py
vicilliar: good catch fixed

tests/tensor_search/test_custom_vector_field.py
vicilliar: removed

src/marqo/tensor_search/validation.py
vicilliar: fixed thanks!

src/marqo/tensor_search/validation.py
pandu-k: because if you are looking at the raw json it will look like this:
```
... Reason: \nThe reason is a bug\nFor info on how...
```

This could be more readable:
```
... Reason: \nThe reason is a bug\n For info on how...
```

pandu-k: Not a huge issue

src/marqo/tensor_search/validation.py
pandu-k: Got it. Thanks!

src/marqo/tensor_search/validation.py
pandu-k: As in the path will be `.../API-Reference/Documents/mappings/` rather than `.../API-Reference/mappings/`

src/marqo/tensor_search/validation.py
vicilliar: Ah got it, will change.

src/marqo/tensor_search/validation.py
vicilliar: ah yep, understood. Although I really was asking if new docs will use `1.4.0-n` as the path.

tests/tensor_search/test_multimodal_tensor_combination.py
vicilliar: it doesn't actually add anything to the test. Just seems like an extraneous doc. But I can put it back, it may just be to check if the correct document was fetched.

src/marqo/tensor_search/enums.py
vicilliar: Tried it, Python doesn't like it. It doesn't like extending enums with new elements added. Also I think it hurts readability, as we want to see each Enum as a distinct set of values.

tests/tensor_search/test_add_documents.py
vicilliar: What error message are you referring to?

src/marqo/tensor_search/validation.py
vicilliar: Done. moved to `models.custom_vector_object.py`

src/marqo/tensor_search/validation.py
vicilliar: done, fixed all references

tests/tensor_search/test_add_documents.py
pandu-k: Basically the new validation error messages. For example:
- https://github.com/marqo-ai/marqo/blob/61ec1cc9829411ad6c1e35227925974ba10fe4bb/src/marqo/tensor_search/validation.py#L466
- https://github.com/marqo-ai/marqo/blob/61ec1cc9829411ad6c1e35227925974ba10fe4bb/src/marqo/tensor_search/validation.py#L601
- https://github.com/marqo-ai/marqo/blob/61ec1cc9829411ad6c1e35227925974ba10fe4bb/src/marqo/tensor_search/validation.py#L567

There are a few reason to test these:
1. Ensure that the formatting is correct (and the variables are being replaced with the correct string)
2. Ensure that formatting code doesn't have any errors. This is quite important. For example, if you are trying to generate an error message for a 4xx error, but you have some error in the generation code, this will return a confusing 500 error to the user
3. Some users will have logic based on the error message (against best practices). For example:  `if "index not found" in error.message`. For these users, changes to error messages are technically breaking changes. We may decide to make these breaking changes, but we at least need to detect a change to the error message. Our error messages are an integral part of our public API

vicilliar: Understood, will add these tests!

pandu-k (approved):

wanliAlex (approved):
Good job!

PR #612, opened 2023-10-04
 Implement create index

farshidz:


src/marqo/core/index_management/index_management.py
```diff
@@ -0,0 +1,101 @@
...
+    FieldFeature, TensorField
+from marqo.core.vespa_index import for_marqo_index as vespa_index_factory
+from marqo.vespa.models import VespaDocument
+from marqo.vespa.vespa_client import VespaClient
+
+logger = marqo.logging.get_logger(__name__)
+
+
+class IndexManagement:
+    _MARQO_SETTINGS_SCHEMA_NAME = 'marqo__settings'
```
pandu-k, line 19: Is there any chance of this conflicting with a user's defined schema? If so, how do we validate against this?

src/marqo/core/index_management/index_management.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def _index_exists(self, name: str) -> bool:
+        return False
+
+    def _ensure_marqo_settings_schema(self, app: str) -> None:
+        schema_path = os.path.join(app, 'schemas', f'{self._MARQO_SETTINGS_SCHEMA_NAME}.sd')
+        if not os.path.exists(schema_path):
+            with open(schema_path, 'w') as f:
+                f.write(self._MARQO_SETTINGS_SCHEMA)
+            self._add_schema_to_services(app, self._MARQO_SETTINGS_SCHEMA_NAME)
```
pandu-k, line 88: Why do we prefer writing the schema first and then reading it from disk, parsing it, mutating it and writing it again, vs loading it into memory editing it and saving it? 

My first thought is that` loading it into memory, editing it and saving it`, has less room for things to go wrong

src/marqo/core/typed_vespa_index.py
```diff
@@ -0,0 +1,271 @@
...
+    @classmethod
+    def to_marqo_document(cls, vespa_document: Dict[str, Any], marqo_index: MarqoIndex) -> Dict[str, Any]:
+        pass
+
+    @classmethod
+    def to_vespa_query(cls, query: MarqoQuery) -> Dict[str, Any]:
+        pass
+
+    @classmethod
+    def _generate_document(cls, marqo_index: MarqoIndex, schema: List[str]) -> None:
```
pandu-k, line 70: this method name is easy to confuse with user documents. Perhaps something like `_generate_document_section_of_schema` makes it clearer

src/marqo/core/typed_vespa_index.py
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    _DISTANCE_METRIC_MAP = {
+        DistanceMetric.PrenormalizedAnguar: 'prenormalized-angular'
+    }
+
+    _INDEX_FIELD_PREFIX = 'marqo__lexical_'
+    _FILTER_FIELD_PREFIX = 'marqo__filter_'
+    _CHUNKS_FIELD_PREFIX = 'marqo__chunks_'
+    _EMBEDDING_FIELD_PREFIX = 'marqo__embeddings_'
```
pandu-k, line 30: if we are going to add validation that prevents users from using these prefixes, then we should move these out into a static data type of file. Then these prefixes can be used in both these contexts

src/marqo/vespa/vespa_client.py
```diff
@@ -206,59 +219,58 @@ def _gzip_compress(self, directory: str) -> io.BytesIO:
...
 
-        response = httpx.post(endpoint)
+        response = httpx_client.post(endpoint)
 
         self._raise_for_status(response)
 
         return response.json()['session-id']
 
-    def _download_application(self, session_id: int) -> str:
+    def _download_application(self, session_id: int, httpx_client: httpx.Client) -> str:
```
pandu-k, line 237: nit: can we use asyncio/threading to speed up downloading the app package, rather than serially? It could speed things up. Also, this suggestion could be premature optimisation. It depends on the user experience that we identify

src/marqo/vespa/vespa_client.py
```diff
@@ -320,4 +332,10 @@ def _raise_for_status(self, resp):
...
             resp.raise_for_status()
         except httpx.HTTPStatusError as e:
-            raise VespaStatusError(e) from e
+            response = e.response
+            try:
+                json = response.json()
+                message = f'{json["error-code"]}: {json["message"]}'
+                raise VespaStatusError(message=message, cause=e) from e
+            except Exception:
+                raise VespaStatusError(message=response.text, cause=e) from e
```
pandu-k, line 345: this exception is being raised in case of a JSON decode error right? This could be a repeated issue throughout all error parsing, and it may be worth splitting out the error handling of JSON decoding to prevent code duplication

src/marqo/core/models/marqo_index.py
```diff
@@ -0,0 +1,82 @@
...
+
+from pydantic import BaseModel
+
+
+class IndexType(Enum):
+    Typed = 'typed'
+    Dynamic = 'dynamic'
+
+
+class FieldType(Enum):
```
pandu-k, line 12: are these Marqo or Vespa fieldtypes? it would be worth making this explicit (e.g.: `MarqoFieldType`)

src/marqo/core/typed_vespa_index.py
```diff
@@ -0,0 +1,271 @@
...
+
+    @classmethod
+    def _get_distance_metric(cls, marqo_distance_metric: DistanceMetric) -> str:
+        try:
+            return cls._DISTANCE_METRIC_MAP[marqo_distance_metric]
+        except KeyError:
+            raise ValueError(f'Unknown Marqo distance metric: {marqo_distance_metric}')
+
+    @classmethod
+    def _get_model_dimension(cls, marqo_index: MarqoIndex) -> int:
```
pandu-k, line 247: nit: I don't feel that this logic belongs to this class. I feel that it makes more sense being part of the MarqoIndex class

src/marqo/core/models/marqo_index.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class MarqoIndex(BaseModel):
+    name: str
+    type: IndexType
+    model: Model
+    distance_metric: DistanceMetric
+    vector_numeric_type: VectorNumericType
+    hnsw_config: HnswConfig
+    fields: Optional[List[Field]]
```
pandu-k, line 71: does `fields` include tensor fields, or not? can we rename this attribute to make it clearer? e.g.: `all_fields` or `all_non_tensor_fields`

src/marqo/core/typed_vespa_index.py
```diff
@@ -0,0 +1,271 @@
...
+            schema.append('}}')
+            schema.append('}')
+
+            field.chunk_field_name = chunks_field_name
+            field.embeddings_field_name = embedding_field_name
+
+        schema.append('}')
+
+    @classmethod
+    def _generate_summaries(cls, marqo_index: MarqoIndex, schema: List[str]) -> None:
```
pandu-k, line 130: Is it possible to keep this method as pure as possible? So not mutating `schema`, and instead returning a new one? I am not a huge fan of these mutating-style methods

src/marqo/core/typed_vespa_index.py
pandu-k, line 70: Is it possible to keep this method as pure as possible? So not mutating `schema`, and instead returning a new one? I am not a huge fan of these mutating-style methods

src/marqo/core/typed_vespa_index.py
pandu-k, line 130: ditto on the mutating paramater comment

pandu-k (commented):
.

src/marqo/core/index_management/index_management.py
farshidz: `marqo__` will be reserved and not allowed as a prefix for index name or field name

src/marqo/core/index_management/index_management.py
farshidz: This is the Marqo settings schema. It's hard-coded and we don't change once created. Not sure if I'm misunderstanding your comment?

src/marqo/core/typed_vespa_index.py
farshidz: It's an internal method, but I agree. I'll rename it

src/marqo/core/typed_vespa_index.py
farshidz: I'm thinking we'll just reject anything starting with `marqo__`

src/marqo/vespa/vespa_client.py
farshidz: Good idea. Though to be honest this is fragile as it is, seeing as we're counting on the single connection pool to help route all downloads to the same config server (but connection pooling is best effort and not guaranteed). So I'd rather run it like this at least for a while, before adding extra complexity (in case we encounter unexpected behaviour). Also this is strictly for index creation/modification, which isn't a common operation.

src/marqo/core/typed_vespa_index.py
farshidz: You are right. Though I don't think it belongs in `MarqoIndex` either. I don't like to have heavy logic in models. For one thing, it increases the chance of hitting circular dependencies (e.g. if model registry needs `MarqoIndex` at some point). 

This operation should be in a separate module that takes care of all model logic. This will be a refactored version of `model_registry` and pull all related functions from `s2_search` and other places into once place. 

I've added a TODO for this

src/marqo/core/models/marqo_index.py
farshidz: `fields` is all fields. There will be validation to make sure `tensor_fields` are already in `fields`.

src/marqo/core/typed_vespa_index.py
farshidz: I don't like side effects either. This is an optimisation strategy. Appending to a list is considered an efficient way to build long strings in Python. Note I'm strictly limiting this behaviour to internal methods and not the public interface.

farshidz: The alternative would be to clone that list, a linear time operation every time (linear in total length of all strings combined).

src/marqo/core/typed_vespa_index.py
farshidz: Same as above

src/marqo/core/typed_vespa_index.py
farshidz: Same as above

src/marqo/core/typed_vespa_index.py
pandu-k: Thanks Farshid! Having good internal method names still helps out developers understand the code faster in the future

src/marqo/core/typed_vespa_index.py
pandu-k: That makes sense

src/marqo/core/index_management/index_management.py
pandu-k: Ok all good - this approach makes sense in this context

src/marqo/vespa/vespa_client.py
pandu-k: makes sense

src/marqo/core/typed_vespa_index.py
pandu-k: makes sense

src/marqo/core/models/marqo_index.py
pandu-k: it would be good to add a comment for this, as it the nature of whether `tensor_fields` are truly `fields` can be ambiguous

src/marqo/core/typed_vespa_index.py
pandu-k: But this operation only is used to create an index schema right? so it should only happen on index-creation/updates? These are slow processes anyway thanks to network IO. Any gains from efficiency here may be eaten up by the network latency. Correctness, on the other hand, is more critical, as a malformed index could be pretty bad

pandu-k (approved):

src/marqo/core/models/marqo_index.py
farshidz: Done

farshidz:
Covered by another PR and merged in

PR #617, opened 2023-10-16
`no_model` index and optional search query

vicilliar:
Feature

1. Every index must be made with a model (with encoding capability)
2. Text query is required for all searches

1. New model `no_model` loader introduced
    - only has 1 required property: `dimensions`. user-defined on index creation.
    - this model is not in registry, as no other property is relevant

2. Text query for search now optional if: search is tensor and context vector is provided

No, provides new options without taking any away.

Yes.

Upcoming

Upcoming

* **Other information**:


src/marqo/s2_inference/model_registry.py
pandu-k, line 1: We will need to make sure we add this to the marqo-commons repo. The longer we take to integrate these two, the more work we need to do to keep them in sync

src/marqo/s2_inference/s2_inference.py
```diff
@@ -77,6 +77,9 @@ def vectorise(model_name: str, content: Union[str, List[str]], model_properties:
                 raise RuntimeError(f"Vectorise created an empty list of batches! Content: {content}")
             else:
                 vectorised = np.concatenate(vector_batches, axis=0)
+    except IllegalVectoriseError as e:
```
wanliAlex, line 81: Under what circumstances will this error be raised? Should this be an internal error?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1156,7 +1158,7 @@ def _lexical_search(
     """
     if not isinstance(text, str):
         raise errors.InvalidArgError(
```
wanliAlex, line 1160: Will this error ever be raised with the pedantic validation?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1517,56 +1525,66 @@ def get_query_vectors_from_jobs(
...
+        if isinstance(q.q, str):
             result[qidx] = get_content_vector(
                 possible_jobs=qidx_to_job.get(qidx, []),
                 jobs=jobs,
                 job_to_vectors= job_to_vectors,
                 treat_urls_as_images=index_info.index_settings[NsField.index_defaults][NsField.treat_urls_and_pointers_as_images],
                 content=q.q
             )
+            continue
+
```
wanliAlex: Do we need an elif here to check the type of q and an else to raise an error?

wanliAlex (changes requested):

src/marqo/s2_inference/s2_inference.py
vicilliar: The `IllegalVectoriseError` will be raised when the `encode` function is called of `no_model`. This will happen when a user calls add docs or search with content, thus raising a 400.

src/marqo/tensor_search/tensor_search.py
vicilliar: Yes, because `_lexical_search` never actually uses pydantic validation. This validation happens in the `SearchQuery` api model, which is only used by `_vector_text_search`. We should definitely change this in the future.

src/marqo/tensor_search/tensor_search.py
vicilliar: Could you elaborate on this? Currently this method has handling for if `q` is a `str`, `None`, or `dict`.

src/marqo/s2_inference/s2_inference.py
wanliAlex: Got it. Thanks!

src/marqo/tensor_search/tensor_search.py
wanliAlex: Got it!

src/marqo/tensor_search/tensor_search.py
wanliAlex: Merge the q type check, `None`, `str`, `dict`, into a combined if else branch to make the logic clearer?
if isinstance(q.q, str): ..., elif isinstance(q.q, None): ..., etc

wanliAlex (changes requested):

src/marqo/tensor_search/tensor_search.py
vicilliar: I can do a hybrid of that, sure! I won't make separate if branches for None and dict, as they share some of the same logic (using context vectors)

vicilliar: Done! Could you check if this is clearer? @wanliAlex

src/marqo/s2_inference/s2_inference.py
```diff
@@ -185,7 +188,15 @@ def _update_available_models(model_cache_key: str, model_name: str, validated_mo
...
     """validate model_properties, if not given then return model_registry properties
     """
-    if model_properties is not None:
+
+    # no_model should always have model_properties (with dimensions only)
+    if model_name == SpecialModels.no_model:
+        if model_properties is None:
+            raise InternalError("If your index is using `no_model`, you must have `model_properties` set.")
+        if "dimensions" not in model_properties:
+            raise InvalidModelPropertiesError("If your index is using `no_model`, your `model_properties` must have `dimensions` set.")
```
pandu-k: nit: probably out of scope for this PR: we should move model specific validation into the model file or class itself

src/marqo/s2_inference/s2_inference.py
```diff
@@ -331,8 +342,14 @@ def _load_model(
...
-    print(f"loading for: model_name={model_name} and properties={model_properties}")
-    loader = _get_model_loader(model_properties.get('name', None), model_properties)
+    if model_name == SpecialModels.no_model:
+        # Using model_name as to not require unnecessary fields in model_properties
+        print(f"Model-less index selected, with dimensions={model_properties['dimensions']}")
+        loader = _get_model_loader(model_name, model_properties)
+    else:
+        # Normal models should use model_properties["name"]
+        print(f"loading for: model_name={model_name} and properties={model_properties}")
+        loader = _get_model_loader(model_properties.get('name', None), model_properties)
```
pandu-k, line 350: nit: this logic is a bit messy

src/marqo/s2_inference/s2_inference.py
pandu-k, line 1: How come we can't use similar logic to the random model? is it because the no_model requires dimensions?

src/marqo/s2_inference/s2_inference.py
```diff
@@ -500,11 +517,14 @@ def _get_model_loader(model_name: str, model_properties: dict) -> Any:
     TODO: standardise these dicts
 
     Returns:
-        dict: a dictionary describing properties of the model.
+        A helper object of class `Model` (subclass depends on the type of model)
     """
 
+    # `no_model` is a special case, we use the name instead of type.
+    if model_name == SpecialModels.no_model:
+        return MODEL_PROPERTIES['loaders'][model_name]
```
pandu-k, line 523: why can't we make it consistent with other models?

src/marqo/tensor_search/models/api_models.py
```diff
@@ -43,9 +44,28 @@ def validate_search_method(cls, value):
...
+        search_method = values.get("searchMethod")
+        query = values.get("q")
+        context = values.get("context")
+
+        if search_method == SearchMethod.LEXICAL:
+            if query is None:
+                raise errors.InvalidArgError("Query must be provided when using lexical search.")
+        elif search_method == SearchMethod.TENSOR:
+            if query is None and context is None:
+                raise errors.InvalidArgError("Either query or context vectors must be provided when using tensor search.")
```
pandu-k: nit: By definition this error message is correct, but we can make it even more clear (in English `or` is sometimes vague): `"Either query or context vectors or both must be provided when using tensor search."`

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1703,7 +1723,7 @@ def create_bulk_search_response(queries: List[BulkSearchQueryEntity], query_to_b
     return results
 
 def _vector_text_search(
-        config: Config, index_name: str, query: Union[str, dict], result_count: int = 5, offset: int = 0,
+        config: Config, index_name: str, query: Optional[Union[str, dict]] = None, result_count: int = 5, offset: int = 0,
```
pandu-k: why is query set to None by default here? I think we should be really careful about adding default params

tests/tensor_search/test_custom_vector_field.py
```diff
@@ -1051,7 +1051,460 @@ def test_lexical_search_with_custom_vector_field_searchable_attributes(self):
...
 
+        assert res["hits"][0]["_id"] == "custom_vector_doc"
+        assert res["hits"][0]["_score"] == 1.0
+        assert res["hits"][0]["_highlights"]["my_custom_vector"] == "custom content is here!!"
+
+    def test_search_no_query(self):
+        """
+        Tests that tensor search is possible with no text query, as long as context vector of correct length is given.
+        Vectorise should not be called at all.
+        """
```
pandu-k, line 1322: This is a good test - can this same test be applied to an index with a normal model?

wanliAlex (dismissed):

src/marqo/s2_inference/model_registry.py
vicilliar: Actually, this will be fine, as there are 2 parts of this model registry:
1. Model properties - This is the one in marqo commons, and `no_model` doesn't have this at all. It was designed such that if a user uses `no_model`, they **must** specify `model_properties`, thus marqo commons does not need to change.
2. Model loaders - This is still handled in the marqo repo, and is what I added for `no_model`.

src/marqo/s2_inference/s2_inference.py
vicilliar: Because this could lead to users accidentally generating random vectors. If a user creates a `no_model` index, it should be so that no marqo vectors are generated with it. If we silently generate random vectors for them, they may misunderstand that add docs and search can still vectorise. It's better that we fail explicitly when they try to vectorise.

src/marqo/s2_inference/s2_inference.py
vicilliar: noted on this!

src/marqo/s2_inference/s2_inference.py
vicilliar: Because other models always have a `type` in their model properties. I elected not to have it for this format as it would be redundant to have any other property aside from dimensions. If we want to keep everything consistent, the user would input model properties as:
```
"model": "no_model",
"model_properties": {
   "dimensions": 123,
   "name": "no_model",
   "type": "no_model"
}
```

which I don't think is good UX.

vicilliar: Instead, `no_model` just takes the `model` key and uses it for everything

src/marqo/tensor_search/models/api_models.py
vicilliar: updated!

tests/tensor_search/test_custom_vector_field.py
vicilliar: yep, this exact test is also in the test class above `test_search_no_query`, which uses a normal model :)

src/marqo/tensor_search/tensor_search.py
vicilliar: you're right, `search` should already make query `None` if it doesn't exist, then pass that `None` to `_vector_text_search`. Fixed.

src/marqo/tensor_search/validation.py
```diff
@@ -722,4 +723,22 @@ def validate_model_dimensions(input_dimensions: Any):
     raise InternalError(f"Model dimensions must be a positive integer! Given model dimensions = {input_dimensions}.")
 
 
+def validate_model_properties_no_model(model_properties: dict):
+    """
+    model_properties dict must have exactly 1 key: "dimensions".
```
pandu-k, line 728: Nit: it could be nice to have a more explicit docstring name, as `no_model` in the function name may get misinterpreted.  For example:
`model_properties dict must have exactly 1 key: "dimensions". Validation when an index's model is set to no_model. `

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1517,56 +1525,68 @@ def get_query_vectors_from_jobs(
     """
```
pandu-k, line 1525: the logic of `get_query_vectors_from_jobs` changes no insignificantly to allow `q.q==None`. I think there should be test cases for this scenario

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1577,10 +1576,10 @@ def get_query_vectors_from_jobs(
                                             f"This causes the error when we do `numpy.mean()` over all the vectors.\n"
                                             f"The original error is `{e}`.\n"
                                             f"Please check `https://docs.marqo.ai/0.0.16/API-Reference/search/#context`.")
-
             if index_info.index_settings['index_defaults']['normalize_embeddings']:
                 norm = np.linalg.norm(merged_vector, axis=-1, keepdims=True)
                 if norm > 0:
+                    # TODO: Why are we calculating norm twice?
```
pandu-k, line 1582: nit: this is not ideal but norm is idempotent

pandu-k (approved):

src/marqo/tensor_search/tensor_search.py
vicilliar: but why can't we use the `norm` variable we calculated like so: `merged_vector /= norm`

PR #623, opened 2023-10-23
Vitus/fix/mq 216/add retries to opensearch requests

VitusAcabado:
This PR introduces a bug fix for MQ-216, add N retries between marqo and opensearch communication for search and indexing requests if ever these requests fail due to transient network errors.

Some requests fail between Marqo and Opensearch due to transient network errors.

 This is done by introducing 4 Configurable environment variables:
- `MARQO_OPENSEARCH_MAX_SEARCH_RETRY_ATTEMPTS`: is used to set the maximum number of attempts to retry if ever a search requests fails due to BackendCommunicationError. It has a default value of 0.

- `MARQO_OPENSEARCH_MAX_SEARCH_RETRY_BACKOFF`: Retries are implemented using an exponential back off interval. This value is used to set the maximum backoff interval in which we send retries. It has a default value of 1 (second).

- `MARQO_OPENSEARCH_MAX_ADD_DOCS_RETRY_ATTEMPTS`: is used to set the maximum number of attempts to retry if ever a add_docs requests fails due to BackendCommunicationError. It has a default value of 0.

- `MARQO_OPENSEARCH_MAX_ADD_DOCS_RETRY_BACKOFF`: similar behavior as for `MARQO_OPENSEARCH_MAX_SEARCH_RETRY_BACKOFF`. It has a default value of 1 (second).

No

yes

None

* **Other information**:


src/marqo/_httprequests.py
```diff
@@ -38,7 +42,14 @@ def send_request(
         path: str,
         body: Optional[Union[Dict[str, Any], List[Dict[str, Any]], List[str], str]] = None,
         content_type: Optional[str] = None,
+        max_retry_attempts: Optional[int] = None,
+        max_retry_backoff_seconds: Optional[int] = None
     ) -> Any:
+        if not max_retry_attempts:
+            max_retry_attempts = 0
```
vicilliar: not sure about having the defaults defined here as well as having defaults for `MARQO_OPENSEARCH_MAX_ADD_DOCS_RETRY_BACKOFF` etc. Another option is to store the defaults here under `constants.py`

src/marqo/_httprequests.py
```diff
@@ -52,64 +63,106 @@ def send_request(
...
+                            verify=to_verify
+                        )
+                    return self.__validate(response)
+                except requests.exceptions.Timeout as err:
+                    raise BackendTimeoutError(str(err)) from err
+                except requests.exceptions.ConnectionError as err:
+                    if (attempt == max_retry_attempts):
+                        raise BackendCommunicationError(str(err)) from err
+                    else:
+                        logger.info(f"BackendCommunicationError encountered... Retrying request to {request_path}.")
```
vicilliar: We could have more detail like `Retry attempt 2 of 3...`

src/marqo/_httprequests.py
```diff
@@ -129,6 +182,19 @@ def __validate(
...
+
+        Args:
+            attempt (int): the attempt number
+
+        Returns:
+            float: the backoff sleep time
+        """
+        return min(
+            cap * 1000, # convert to milliseconds
+            (2 ** attempt) * 10 # start at 10ms for first attempt
```
vicilliar, line 198: Is there any particular basis for doing time intervals of 10ms, 20ms, 30ms, etc?

src/marqo/tensor_search/backend.py
```diff
@@ -62,7 +62,9 @@ def get_index_info(config: Config, index_name: str) -> IndexInfo:
 
 def add_customer_field_properties(config: Config, index_name: str,
                                   customer_field_names: Iterable[Tuple[str, enums.OpenSearchDataType]],
-                                  multimodal_combination_fields: Dict[str, Iterable[Tuple[str, enums.OpenSearchDataType]]]):
+                                  multimodal_combination_fields: Dict[str, Iterable[Tuple[str, enums.OpenSearchDataType]]],
+                                  max_retry_attempts: int = 0,
```
vicilliar: Another instance of defining the defaults. You could also get this from `constants.py` instead.

src/marqo/tensor_search/enums.py
```diff
@@ -111,6 +111,10 @@ class EnvVars:
     MARQO_MAX_NUMBER_OF_REPLICAS = "MARQO_MAX_NUMBER_OF_REPLICAS"
     MARQO_BEST_AVAILABLE_DEVICE = "MARQO_BEST_AVAILABLE_DEVICE"
     MARQO_MAX_ADD_DOCS_COUNT = "MARQO_MAX_ADD_DOCS_COUNT"
+    MARQO_OPENSEARCH_MAX_SEARCH_RETRY_ATTEMPTS = "MARQO_OPENSEARCH_MAX_SEARCH_RETRY_ATTEMPTS"
```
vicilliar: To be uniform with the other env var names, it may be better to start with `MARQO_MAX`. Also since we're moving to vespa, we could also be backend-agnostic in naming. So something like `MARQO_MAX_BACKEND_RETRY_ATTEMPTS`. Any feedback @pandu-k ?

vicilliar (changes requested):

src/marqo/tensor_search/enums.py
pandu-k: I don't mind too much, the important thing is to clearly communicate what the env var does

src/marqo/_httprequests.py
VitusAcabado: that's a good idea, I can move it there instead if that's what's prefered

src/marqo/_httprequests.py
VitusAcabado: Yep I had that discussion with @pandu-k as well and it didn't really matter too much for the initial implementation and most retry mechanisms are executed silently in terms of a user's perspective.

If that is what is preferred, I can implement that as well.

src/marqo/tensor_search/enums.py
vicilliar: So can we use `MARQO_MAX_BACKEND_SEARCH_RETRY_ATTEMPTS` and `MARQO_MAX_BACKEND_ADD_DOCS_RETRY_ATTEMPTS`?

src/marqo/_httprequests.py
VitusAcabado: the value for the initial sleep interval is arbitrary but I decided for a small start since our target is for search requests to execute within < 300ms ideally and the backoff exponentially grows to one rapidly (just needs 7 retries to reach >1s).

src/marqo/tensor_search/enums.py
VitusAcabado: Wouldn't `MARQO_MAX_BACKEND_RETRY_ATTEMPTS` imply that it will apply to all backend requests as well? I mean we can be specific about it in the documentation but not quite sure if it will be immediately apparent for a user that it only applies to `add_docs()` and `search()` requests for now. Thoughts?

VitusAcabado:
Screenshot of implementation working locally
image

src/marqo/tensor_search/enums.py
vicilliar: @VitusAcabado Yep `MARQO_MAX_BACKEND_RETRY_ATTEMPTS` was a typo, I really meant `MARQO_MAX_BACKEND_SEARCH_RETRY_ATTEMPTS` and `MARQO_MAX_BACKEND_ADD_DOCS_RETRY_ATTEMPTS` to specify that it is for add docs and search.

src/marqo/_httprequests.py
vicilliar: Yep, this reasoning makes sense to me, thanks!

src/marqo/tensor_search/tensor_search.py
```diff
@@ -672,17 +672,23 @@ def add_documents(config: Config, add_docs_params: AddDocsParams):
...
-                multimodal_combination_fields=new_obj_fields)
+                multimodal_combination_fields=new_obj_fields, max_retry_attempts=max_add_docs_retry_attempts,
+                max_retry_backoff_seconds=max_add_docs_retry_backoff)
 
             # ADD DOCS TIMER-LOGGER (5)
             start_time_5 = timer()
             with RequestMetricsStore.for_request().time("add_documents.opensearch._bulk"):
                 serialised_body = utils.dicts_to_jsonl(bulk_parent_dicts)
+                
                 index_parent_response = HttpRequests(config).post(
```
vicilliar, line 700: The retry attempts and backup parameters were added to add docs mappings update and `_bulk` calls, but not the call to `_refresh` if `add_docs_params.auto_refresh is True`. It would be good to add it there as well.

src/marqo/_httprequests.py
```diff
@@ -38,7 +43,14 @@ def send_request(
         path: str,
         body: Optional[Union[Dict[str, Any], List[Dict[str, Any]], List[str], str]] = None,
         content_type: Optional[str] = None,
+        max_retry_attempts: Optional[int] = None,
+        max_retry_backoff_seconds: Optional[int] = None
     ) -> Any:
+        if max_retry_attempts is None:
```
vicilliar, line 50: Can you add a unit test for this function that tests these defaults? Try not passing `max_retry_attempts` and `max_retry_backoff_seconds` and ensure that it defaults to the constants you set.

tests/tensor_search/test__httprequests.py
```diff
@@ -86,5 +89,58 @@ def run():
...
+                config=self.config, index_name=self.index_name_1, text="cool match",
+                search_method=SearchMethod.LEXICAL)
+                raise AssertionError
+            except BackendCommunicationError as e:
+                assert e.code == "backend_communication_error"
+                assert e.status_code == HTTPStatus.INTERNAL_SERVER_ERROR
+                assert "Max retries exceeded with url" in e.message
+
+            try:
+                res = tensor_search.search(
```
vicilliar, line 514: We should also add a similar test for `bulk_search`.

tests/tensor_search/test__httprequests.py
```diff
@@ -86,5 +89,58 @@ def run():
...
+        def run():
+            try:
+                res = tensor_search.add_documents(
+                    config=self.config, add_docs_params=AddDocsParams(
+                        index_name=self.index_name_1,
+                        docs=[{"some ": "doc"}], auto_refresh=True, device="cpu"
+                    )
+                )
+                raise AssertionError
+            except BackendCommunicationError as e:
```
vicilliar, line 400: Does this test actually check that the retry happens 3 times? Perhaps we can use `call_args_list` of the mocked HTTP requests to count the retries and ensure it is correct.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1221,9 +1227,16 @@ def _lexical_search(
     total_preprocess_time = RequestMetricsStore.for_request().stop("search.lexical.processing_before_opensearch")
     logger.debug(f"search (lexical) pre-processing: took {(total_preprocess_time):.3f}ms to process query.")
 
+    max_search_retry_attempts = utils.read_env_vars_and_defaults_ints(EnvVars.MARQO_MAX_BACKEND_SEARCH_RETRY_ATTEMPTS)
+    max_search_retry_backoff = utils.read_env_vars_and_defaults_ints(EnvVars.MARQO_MAX_BACKEND_SEARCH_RETRY_BACKOFF)
```
vicilliar: nit: are we sure the backoff values will always be `int`? We may want to accept a backoff of 1.5 seconds, for example. I have no strong opinion though.

src/marqo/_httprequests.py
```diff
@@ -52,64 +64,106 @@ def send_request(
...
-                        data=json.dumps(body) if body else None,
-                        verify=to_verify
-                    )
-                return self.__validate(response)
-            except requests.exceptions.Timeout as err:
-                raise BackendTimeoutError(str(err)) from err
-            except requests.exceptions.ConnectionError as err:
-                raise BackendCommunicationError(str(err)) from err
+            
+            for attempt in range(max_retry_attempts + 1):
```
vicilliar, line 69: It would be better to have more test cases for edge cases here in general, as the `send_request` method is used for almost everything. We want to make sure this code change is quite stable.

src/marqo/_httprequests.py
```diff
@@ -52,64 +64,106 @@ def send_request(
...
+                            timeout=self.config.timeout,
+                            headers=req_headers,
+                            data=json.dumps(body) if body else None,
+                            verify=to_verify
+                        )
+                    return self.__validate(response)
+                except requests.exceptions.Timeout as err:
+                    raise BackendTimeoutError(str(err)) from err
+                except requests.exceptions.ConnectionError as err:
+                    if (attempt == max_retry_attempts):
```
vicilliar, line 92: Add unit tests for `send_request` with varying `max_retry_attempts` and `max_retry_backoff_seconds`

src/marqo/_httprequests.py
```diff
@@ -129,6 +183,19 @@ def __validate(
         except requests.exceptions.HTTPError as err:
             convert_to_marqo_web_error_and_raise(response=request, err=err)
 
+    def calculate_backoff_sleep(self, attempt: int, cap: int) -> float:
```
vicilliar, line 187: It's worth making simple unit tests for this method.

src/marqo/tensor_search/backend.py
```diff
@@ -62,7 +63,9 @@ def get_index_info(config: Config, index_name: str) -> IndexInfo:
 
 def add_customer_field_properties(config: Config, index_name: str,
                                   customer_field_names: Iterable[Tuple[str, enums.OpenSearchDataType]],
-                                  multimodal_combination_fields: Dict[str, Iterable[Tuple[str, enums.OpenSearchDataType]]]):
+                                  multimodal_combination_fields: Dict[str, Iterable[Tuple[str, enums.OpenSearchDataType]]],
+                                  max_retry_attempts: int = DEFAULT_MARQO_MAX_BACKEND_RETRY_ATTEMPTS,
```
vicilliar: Since the codepath for this is  `add_customer_field_properties` -> `put` -> `send_request`, wouldn't it be better to set the default `max_retry_attempts` and `max_retry_backoff_seconds` here to `None`, so that `send_request` catches it and fills it in with the constant?

vicilliar (changes requested):

src/marqo/tensor_search/tensor_search.py
VitusAcabado: Having floating point backoff values doesn't really matter since it just changes the growth rate of the values by a tiny bit. Keeping it as `int` makes it simple.

pandu-k: what happens if an admin tries setting this as a float? a parse error?

src/marqo/tensor_search/backend.py
pandu-k: I agree

src/marqo/tensor_search/constants.py
```diff
@@ -47,3 +47,6 @@
     'zb': 7, 
     'yb': 8
 }
+
+DEFAULT_MARQO_MAX_BACKEND_RETRY_ATTEMPTS = 0
+DEFAULT_MARQO_MAX_BACKEND_RETRY_BACKOFF = 1
```
pandu-k: nit: if we are going to define these separately, perhaps we should just define them as default env vars, in `default_env_vars()`

pandu-k: This will give devs more ways to experiment with Marqo

src/marqo/tensor_search/tensor_search.py
pandu-k: Should we instead add `"?refresh=true" if add_docs_params.auto_refresh is True else "?refresh=false"` to the call to the `_bulk` endpoint? It will reduce the number of requests to OS and solve this problem.

https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-refresh.html#docs-refresh

src/marqo/tensor_search/tensor_search.py
pandu-k: At least we get this error message, which is readable to a user: https://github.com/marqo-ai/marqo/blob/c607ee3bd8483be0fd7e019a5d28a219977e4093/src/marqo/tensor_search/utils.py#L189

VitusAcabado: yup, since it expects it to be parsed as an int.

VitusAcabado: Yup, `read_env_vars_and_defaults_ints` has some validation and returns a parse error since it expects a string version of an int value

src/marqo/tensor_search/constants.py
VitusAcabado: In `default_env_vars`, I've added them as what the environment names are, would you reckon I add them as well?

so from:
```
EnvVars.MARQO_MAX_BACKEND_SEARCH_RETRY_ATTEMPTS: 0,
EnvVars.MARQO_MAX_BACKEND_SEARCH_RETRY_BACKOFF: 1,
EnvVars.MARQO_MAX_BACKEND_ADD_DOCS_RETRY_ATTEMPTS: 0,
EnvVars.MARQO_MAX_BACKEND_ADD_DOCS_RETRY_BACKOFF: 1
```

to:
```
EnvVars.DEFAULT_MARQO_MAX_BACKEND_RETRY_ATTEMPTS = 0
EnvVars.DEFAULT_MARQO_MAX_BACKEND_RETRY_BACKOFF = 1
EnvVars.MARQO_MAX_BACKEND_SEARCH_RETRY_ATTEMPTS: EnvVars.DEFAULT_MARQO_MAX_BACKEND_RETRY_ATTEMPTS,
EnvVars.MARQO_MAX_BACKEND_SEARCH_RETRY_BACKOFF: EnvVars.DEFAULT_MARQO_MAX_BACKEND_RETRY_BACKOFF ,
EnvVars.MARQO_MAX_BACKEND_ADD_DOCS_RETRY_ATTEMPTS: EnvVars.DEFAULT_MARQO_MAX_BACKEND_RETRY_ATTEMPTS,
EnvVars.MARQO_MAX_BACKEND_ADD_DOCS_RETRY_BACKOFF: EnvVars.DEFAULT_MARQO_MAX_BACKEND_RETRY_BACKOFF 
``

File: src/marqo/_httprequests.py [hunk 16]
```diff
@@ -38,7 +43,14 @@ def send_request(
         path: str,
         body: Optional[Union[Dict[str, Any], List[Dict[str, Any]], List[str], str]] = None,
         content_type: Optional[str] = None,
+        max_retry_attempts: Optional[int] = None,
+        max_retry_backoff_seconds: Optional[int] = None
     ) -> Any:
+        if max_retry_attempts is None:
+            max_retry_attempts = read_env_vars_and_defaults_ints("DEFAULT_MARQO_MAX_BACKEND_RETRY_ATTEMPTS")
```
vicilliar: nit: you can use the enum instead of the raw string here.

tests/tensor_search/test__httprequests.py
```diff
@@ -85,6 +95,517 @@ def run():
...
+                    assert e.code == "backend_communication_error"
+                    assert e.status_code == HTTPStatus.INTERNAL_SERVER_ERROR
+                    assert "Max retries exceeded with url" in e.message
+                    assert method.call_count == 1
+            return True
+        assert run()
+
+    def test_httprequest_send_request_variable_default_retry_and_backoff(self):
+        retry_backoff_list = [
+            {'retry_attempts': 5, 'backoff_seconds': 3},
```
vicilliar, line 252: nit: backoff_seconds isn't actually doing anything in this test right?

vicilliar (changes requested):

tests/tensor_search/test__httprequests.py
VitusAcabado: done

tests/tensor_search/test__httprequests.py
VitusAcabado: done

tests/tensor_search/test__httprequests.py
VitusAcabado: good catch! i forgot to add them as default env vars

tests/tensor_search/test__httprequests.py
```diff
@@ -116,6 +117,44 @@ def test_backoff_sleep_with_zero_cap(self):
...
+        mock_post = mock.MagicMock()
+        mock_put = mock.MagicMock()
+        mock_delete = mock.MagicMock()
+
+        mock_allowed_operations = {mock_post, mock_get, mock_put, mock_delete}
+
+
+        mock_response = requests.Response()
+        mock_response.status_code = 200
+        mock_response.json = lambda: {'succes': {'hits': ['test']}}
```
vicilliar: nit: `success` misspelled. But good test! Thanks Vitus.

tests/tensor_search/test__httprequests.py
```diff
@@ -213,6 +254,11 @@ def run():
                         assert e.status_code == HTTPStatus.INTERNAL_SERVER_ERROR
                         assert "Max retries exceeded with url" in e.message
                         assert method.call_count == mock_retry_pair['retry_attempts'] + 1 # +1 since the first call is not a retry
+                        expected_total_wait_time = 0
+                        for i in range(mock_retry_pair['retry_attempts']):
+                            expected_total_wait_time += self.httprequest_object.calculate_backoff_sleep(i, mock_retry_pair['backoff_seconds'])
```
vicilliar, line 304: great job on testing this interval!

vicilliar (changes requested):

vicilliar (approved):

PR #628, opened 2023-10-30
Vespa running marqo locally.

wanliAlex:
Allow running Marqo and vespa locally. No docker in docker.


src/marqo/tensor_search/index_meta_cache.py
```diff
@@ -179,6 +179,7 @@ def populate_cache(config: Config):
     """
     global index_info_cache
     index_management = IndexManagement(config.vespa_client)
+    index_management.vespa_client.wait_for_application_convergence()
```
farshidz: Why do we need this? If it's for local Vespa, can we wait on convergence in entrypoint?

farshidz (approved):

PR #630, opened 2023-10-30
Implement tensor search, get docs, delete docs, delete index

farshidz:
Implement basic tensor search, get docs (including batch), delete docs (including batch) and delete index with Vespa.


src/marqo/core/search/search_filter.py
```diff
@@ -0,0 +1,274 @@
...
+
+class Or(Operator):
+    greedy = False
+
+    def __str__(self):
+        return 'OR'
+
+
+class SearchFilter:
+    """
```
wanliAlex, line 51: Impressive work!

wanliAlex (approved):

src/marqo/core/search/search_filter.py
farshidz: Thank you

PR #631, opened 2023-10-31
Patch multimodal so broken document is not indexed

wanliAlex:
patch a bug in multimodal that indexes the broken documents

a broken document is still indexed

broken document will not be indexed

no

running

no

no

* **Other information**:
no


pandu-k (commented):
Thanks Li! 2 comments:
1. There seems to be some formatting your IDE applies to the test file, which adds a lot of diffs, making it a bit harder to read
2. Great job adding new tests. Do these new unit tests fail, without the bug fix?
3. For the new tests: should we check the response body of the add_documents call, where a document isn't indexed properly, to make sure the error message appears as expected?
4. For the new tests: can we have a case where some documents get indexed successfully but other documents, that have broken URL or vectorise breaks, don't get indexed successfully?

pandu-k (approved):

PR #632, opened 2023-11-01
Lexical search and add docs use_existing_tensors 

farshidz:
* Implement basic lexical search (no filters or score modifiers).
* Implement `use_existing_tensors` for add docs
* Store and return multimodal combination weights. Use these with `use_existing_tensors` to update vectors if weights have changed but values are the same


src/marqo/core/structured_vespa_index.py
```diff
@@ -279,7 +285,50 @@ def _to_vespa_tensor_query(cls, marqo_query: MarqoTensorQuery, marqo_index: Marq
...
+            else cls._RANK_PROFILE_BM25
+
+        query_inputs = {}
+        query_inputs.update({
+            f: 1 for f in fields_to_search
+        })
+        if score_modifiers:
+            query_inputs.update(score_modifiers)
+
+        query = {
```
wanliAlex, line 320: Would it be good to have a pedantic class to better manage the Vespa query dict?

wanliAlex (approved):

src/marqo/core/structured_vespa_index.py
farshidz: Yes, I think there are a few things that require pydantic models. We'll keep this as tech debt for now and add later

PR #633, opened 2023-11-02
`search_model` added to index settings

vicilliar:
feature

An index only has `model`, which is used to encode vectors for both add documents and search.

Index settings defaults can now have `search_model` and `search_model_properties`, which will determine the model used for encoding search queries.
Notes (user-facing):
- If not explicitly set, search_model will always assimilate value of model.
- search_model cannot be set without setting model.
- search_model dimensions must match model dimensions.

Notes (internal):
- new module `tensor_search/create_index.py` created to contain all utils for creating indexes.
- Tests for index settings and IndexInfo all updated to include search_model.

no

yes (but still adding more)

to follow

to follow

* **Other information**:


src/marqo/tensor_search/validation.py
```diff
@@ -742,3 +742,25 @@ def validate_model_properties_no_model(model_properties: dict):
...
+    3. if search_model_properties is given, then search_model must be given as well.
+    """
+    model_name = index_settings[IndexSettingsField.index_defaults].get(IndexSettingsField.model)
+    model_properties = index_settings[IndexSettingsField.index_defaults].get(IndexSettingsField.model_properties)
+    search_model_name = index_settings[IndexSettingsField.index_defaults].get(IndexSettingsField.search_model)
+    search_model_properties = index_settings[IndexSettingsField.index_defaults].get(IndexSettingsField.search_model_properties)
+
+
+    if search_model_name is not None and model_name is None:
+        raise UnknownModelError(f"`search_model` cannot be specified without also specifying `model`. Received search_model={search_model_name} but found no `model`")
```
farshidz: All three look like they should be `InvalidArgError` as they are user input errors, not inference errors?

src/marqo/tensor_search/backend.py
```diff
@@ -47,12 +47,21 @@ def get_index_info(
...
+    # Identify index `model` from mapping metadata
     if "model" in res[index_name]["mappings"]["_meta"]:
         model_name = res[index_name]["mappings"]["_meta"]["model"]
     else:
         raise errors.NonTensorIndexError(
             "get_index_info: couldn't identify embedding model name "
             F"in index mappings! Mapping: {res}")
+    
+    # Identify index `search_model` from mapping metadata
+    if "search_model" in res[index_name]["mappings"]["_meta"]:
```
farshidz, line 59: This doesn't appear to be backwards compatible for existing indexes?

farshidz (changes requested):
In addition to inline comments, please add tests to verify existing indexes will continue to work. I also suggest adding debug level log messges to indicate whether a different search model is being utilised

src/marqo/tensor_search/validation.py
vicilliar: You're right. I had handling in `tensor_search` to convert these `UnknownModelError` to `InvalidArgError`, but it's better to just have them here in the validation function itself.

src/marqo/tensor_search/backend.py
vicilliar: fixed, setting `search_model_name` to `None` if it doesn't appear in OS response.

vicilliar:
Added:
1. Debug log message when `search_model` is used for search
2. Tests that `IndexInfo.search_model_name` set to `None` if `search_model` not included in backend mappings response
3. Tests that `determine_model_for_search_vectorisation` returns `model` and `model_properties` if `IndexInfo.search_model_name is None`.
4. Tests that `search` will use `model` for `vectorise` if `search_model` is NOT in backend response.

vicilliar:
Manual tests also added and passed: 
image

farshidz (approved):

Issue #585, opened 2023-08-24
[ENHANCEMENT] Support for prefixing and suffixing of text


OwenElliottDev:
**Is your feature request related to a problem? Please describe.**
Many new SOTA models for retrieval are trained with prefixes on the queries and documents, thus they expect these at inference as well. Failure to provide these degrades performance. Currently there is no way in Marqo to add a prefix or suffix to every chunk in a document as text pre-processing occurs internally.

I don't know of any models with a requirement for a specific suffix so the suffixing part can be removed if we don't want to pre-empt a use case that doesn't exist yet.

**Describe the solution you'd like**
New optional parameters are added to `search` and `add_documents` to enable prefixes like so:
```python
import marqo

mq = marqo.Client()

mq.index("my-index").add_documents(
    [{"_id": 1, "text": "lorum ipsum"}],
    text_chunk_prefix="passage: ",
    text_chunk_suffix="",
    tensor_fields=["text"]
)

mq.index("my-index").search(
    "lorum ipsum",
    query_prefix="query: ",
    query_suffix=""
)

mq.index("my-index").search(
    {"lorum": 1.0, "ipsum": 0.2},
    query_prefix="query: ",
    query_suffix=""
)
```

When the text is chunked internally, each chunk should have the prefix added to it so that it can be tokenised and passed to the model to match the prefixing applied in training.

For queries the prefix is added to the front of every query that isn't a URL. It should be added to the front of every query in a weighted query.

**Describe alternatives you've considered**
These could be part of the index defaults however if a model appears in the future with different prefixes for different sub tasks (or something similar) then this won't be a good approach because it locks you in. One nice thing though is that this approach means users won't forget to add them and the correct prefix can me included in the default models.

```
settings = {
    "index_defaults": {
        "treat_urls_and_pointers_as_images": False, 
        "text_preprocessing": {
            "split_method": "sentence", 
            "split_length": 3, 
            "split_overlap": 1,
        }, 
        "model_properties": {
            "name": "intfloat/e5-base-v2", 
            "type": "hf", 
            "dimensions": 768,
            "query_prefix": "query: ",
            "query_suffix": "",
            "text_chunk_prefix": "passage: ",
            "text_chunk_suffix": ""
        }, 
        "model": "hf-e5-base-v2", 
        "normalize_embeddings": True, 
    },
}
```

**Additional context**
[...]

PR #643, opened 2023-11-15
Prefix for field chunks and search queries

vicilliar:
Feature

No ability to add prefixes to chunks. Needed to improve some models' performance (E5). Full issue: https://github.com/marqo-ai/marqo/issues/585

2 prefixes introduced:
1. `text_chunk_prefix`
2. `text_query_prefix`

These can be set in 3 locations:
1. Inside `model_properties` object (in registry)
2. Inside override fields in `text_processing` in index settings
3. As parameters in `add_documents`, `search`, or `bulk_search` request

Request-level prefixes take priority, then override fields, the registry properties.
`text_query_prefix` is read from `model_properties` if `search_model_properties` doesn't exist, ensuring backwards compatibility with pre-v1.5.0 indexes.

No

To follow.

To follow.

To follow.

* **Other information**:


src/marqo/s2_inference/model_registry.py
```diff
@@ -672,6 +696,16 @@ def _get_sbert_test_properties() -> Dict:
                 "tokens":128,
                 "type":"test",
                 "notes": ""},
+            "prefix-test-model":
```
farshidz, line 699: Is this here just to be used in tests? Could a user pick this when creating an index?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -563,9 +571,15 @@ def add_documents(config: Config, add_docs_params: AddDocsParams):
...
-                        content_chunks = text_processor.split_text(field_content, split_by=split_by,
+                        
+                        
+                        # text chunks: WITHOUT prefix, stored in backend chunk list
+                        text_chunks = text_processor.split_text(field_content, split_by=split_by,
                                                                    split_length=split_length, split_overlap=split_overlap)
-                        text_chunks = content_chunks
+                        
+                        # content chunks: WITH prefix, used to generate vectors (text prefix not actually stored in backend)
+                        content_chunks = text_processor.prefix_text_chunks(text_chunks, text_chunk_prefix)
```
farshidz, line 581: You could incorporate prefixing into splitting to avoid an extra pass over the text. Not sure how significant the difference will be though.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -583,8 +597,11 @@ def add_documents(config: Config, add_docs_params: AddDocsParams):
                                         f"Could not find image found at `{field_content}`. \n"
                                         f"Reason: {str(image_repo[field_content])}"
                                     )
+                            elif isinstance(field_content, str):
+                                image_data = text_chunk_prefix + field_content     # Add prefix to URL if it's to be treated as-is.
                             else:
-                                image_data = field_content
+                                image_data = field_content      # If it's actual image data, just pass it through.
```
farshidz, line 603: I feel like this might be unreachable. But let's leave it for later

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1711,11 +1771,73 @@ def create_empty_query_response(queries: List[BulkSearchQueryEntity]) -> List[Di
...
+        if q.q is None:
+            prefixed_q = q.q
+        elif isinstance(q.q, str):
+            if treat_urls_as_images and _is_image(q.q):
+                # Images get no prefix
+                prefixed_q = q.q
+            else:
+                # Single text query: add prefix
+                prefixed_q = f"{text_query_prefix}{q.q}"
+        else:  # is dict:
```
farshidz, line 1801: Looks like dictionary is a variation of a single string query. You could extract the prefixing logic for one string into another function, and call that for each dict entry (or just once if query is one string) to avoid implementing the same logic twice

src/marqo/tensor_search/tensor_search.py
```diff
@@ -2146,6 +2295,8 @@ def vectorise_multimodal_combination_field(
                                 f"Could not find image found at `{sub_content}`. \n"
                                 f"Reason: {str(image_repo[sub_content])}"
                             )
+                    elif isinstance(sub_content, str):
```
farshidz: The `NsField.treat_urls_and_pointers_as_images` check in line 2289 is redundant (old code but let's fix it here) as `infer_if_image` covers that

src/marqo/tensor_search/tensor_search.py
```diff
@@ -2165,6 +2316,10 @@ def vectorise_multimodal_combination_field(
     try:
         start_time = timer()
         text_vectors = []
+
+        # Add prefix to all text content first
+        text_content_to_vectorise = text_processor.prefix_text_chunks(text_content_to_vectorise, text_chunk_prefix)
```
farshidz: Minor but best to call this something else rather than overwrite/reuse existing variable

tests/processing/test_split_text.py
```diff
@@ -79,3 +79,18 @@ def test_split_text_single_word(self):
...
         assert result == [text]
 
+
+class TestPrefixTextChunks(unittest.TestCase):
+    def test_prefix_test_chunks(self):
+        text_splits = ["a", "b", "c", ""]
+        assert prefix_text_chunks(text_splits, "prefix_") == \
+            ["prefix_a", "prefix_b", "prefix_c", "prefix_"]
+        # Empty string prefix should do nothing
+        assert prefix_text_chunks(text_splits, "") == \
```
farshidz, line 92: This is a separate test case. If you're covering different cases in one function, use subtests so that if the first one fails, the second one still runs. [Like this](https://github.com/marqo-ai/marqo/blob/mainline-poc/tests/core/search/test_search_filter.py#L115)

tests/tensor_search/test_add_documents.py
```diff
@@ -1450,4 +1451,45 @@ def test_determine_document_field_type(self):
...
+                    IndexSettingsField.treat_urls_and_pointers_as_images: True,
+                    IndexSettingsField.text_preprocessing: {
+                        IndexSettingsField.override_text_chunk_prefix: "index-override"
+                    }
+                }
+            }
+        )
+
+        # All prefixes on (request level chosen)
+        assert add_docs.determine_text_chunk_prefix("request-level", index_info_with_override) == "request-level"
```
farshidz: These are separate test cases. I suggest separate tests, each with a short docstring description (docsting for test cases comes up in test results).

tests/tensor_search/test_bulk_search.py
```diff
@@ -184,6 +184,13 @@ def setUp(self):
         ]
         self.selected_device = 'cpu'
 
+        tensor_search.create_vector_index(self.config, "index_name_1")
```
farshidz, line 187: Assuming these tests have been passing, why did you add this code now? There are no other changes

farshidz (changes requested):

src/marqo/s2_inference/model_registry.py
vicilliar: I just made it to be used in tests. But technically yes, it can be selected by a user.

src/marqo/tensor_search/tensor_search.py
vicilliar: I tried this, it's better that it's separate, because we want `text_chunks` to not have prefixes, while `content_chunks` have prefixes.

src/marqo/tensor_search/tensor_search.py
vicilliar: Agreed, I just left it in to not break anything, just in case. It should be cleared up in v2 anyway.

src/marqo/tensor_search/tensor_search.py
vicilliar: Not sure if this would be worth it, as:
1. The `treat_urls_as_images` check would happen for every single dict entry, which is inefficient, as opposed to just doing it once at the start, which I'm doing now.
2. Removing that, the only real logic is checking if `_is_image(q)`, which I don't think is enough to make a function for.

src/marqo/tensor_search/tensor_search.py
vicilliar: Makes sense, removed that check.

src/marqo/tensor_search/tensor_search.py
vicilliar: done

tests/processing/test_split_text.py
vicilliar: Separated the test case.

tests/tensor_search/test_bulk_search.py
vicilliar: Because `tensor_search.run_vectorise_pipeline` now will raise an error if the index does not exist. Because it now calls `get_index_info()` while determining the `text_query_prefix`.

tests/tensor_search/test_add_documents.py
vicilliar: Used subTest to fix this problem

src/marqo/s2_inference/model_registry.py
farshidz: Can you find a way to avoid changing code for test?

farshidz (approved):

PR #647, opened 2023-11-17
Unstructured index

wanliAlex:

* **Other information**:


src/marqo/core/index_management/index_management.py
```diff
@@ -138,7 +138,7 @@ def delete_index_by_name(self, index_name: str) -> None:
         self.vespa_client.deploy_application(app)
         self._delete_index_settings_by_name(index_name)
 
-    def batch_delete_indexes(self, marqo_indexes: List[MarqoIndex]) -> None:
+    def batch_delete_indexes(self, marqo_indexes: List[Union[MarqoIndex, str]]) -> None:
```
farshidz: Generic interfaces are error prone. Create `batch_delete_indexes_by_name`, refactor so that this method simply calls `batch_delete_indexes_by_name`

src/marqo/core/constants.py
```diff
@@ -4,3 +4,6 @@
 MARQO_DOC_CHUNKS = 'chunks'
 MARQO_DOC_EMBEDDINGS = 'embeddings'
 MARQO_DOC_ID = '_id'
+
+VESPA_DOC_ID = "id"
```
farshidz: `core.constants` is probably not the right place for these as they are specific to `VespaIndex` classes. You can create a `vespa_index` package and move these under that, or repeat the values. Either way these constants will go away when we create a pydantic model for Vespa document

src/marqo/tensor_search/models/index_settings.py
```diff
@@ -154,3 +154,8 @@ def from_marqo_index(cls, marqo_index: core.MarqoIndex) -> "IndexSettings":
             )
         else:
             raise errors.InternalError(f"Unknown index type: {type(marqo_index)}")
+
+from pydantic import BaseModel
+class IndexNamesAndSettings(BaseModel):
```
farshidz: My suggestion:
* Name this `BatchIndexSettings` or `IndexSettingsWithName`
* Inherit `IndexSettings` (instead of `BaseModel`) and add `indexName: str`
* API takes `List[IndexSettingsWithName]`

src/marqo/tensor_search/api.py
```diff
@@ -272,6 +274,45 @@ def get_cuda_info():
     return tensor_search.get_cuda_info()
 
 
+@app.post("/batch/indexes/delete-batch")
+@utils.enable_batch_apis()
+def delete_batch_indexes(index_names: List[str], marqo_config: config.Config = Depends(get_config)):
```
farshidz: `batch_delete_indexes` is more in-line with our naming conventions

src/marqo/tensor_search/api.py
```diff
@@ -272,6 +274,45 @@ def get_cuda_info():
...
+    """An internal API used for testing processes. Not to be used by users."""
+    tensor_search.delete_batch_indexes(index_names=index_names, config=marqo_config)
+    return JSONResponse(content={"acknowledged": True}, status_code=200)
+
+
+@app.post("/batch/indexes/create-batch")
+@utils.enable_batch_apis()
+def create_batch_indexes(index_names_and_settings: IndexNamesAndSettings, \
```
farshidz: `batch_create_indexes` is more in-line with our naming conventions

src/marqo/tensor_search/api.py
```diff
@@ -272,6 +274,45 @@ def get_cuda_info():
     return tensor_search.get_cuda_info()
 
 
+@app.post("/batch/indexes/delete-batch")
```
farshidz: How about `POST /indexes/delete-batch` to be consistent with batch delete docs?

src/marqo/tensor_search/api.py
```diff
@@ -272,6 +274,45 @@ def get_cuda_info():
...
+@utils.enable_batch_apis()
+def delete_batch_indexes(index_names: List[str], marqo_config: config.Config = Depends(get_config)):
+    """An internal API used for testing processes. Not to be used by users."""
+    tensor_search.delete_batch_indexes(index_names=index_names, config=marqo_config)
+    return JSONResponse(content={"acknowledged": True}, status_code=200)
+
+
+@app.post("/batch/indexes/create-batch")
```
farshidz: `POST /indexes/create-batch`?

src/marqo/tensor_search/api.py
```diff
@@ -272,6 +274,45 @@ def get_cuda_info():
...
+            "acknowledged": True,
+            "index_names": [index_name for index_name in index_names_and_settings.index_names]
+        },
+        status_code=200
+    )
+
+
+@app.post("/batch/indexes/documents/delete-batch")
+@utils.enable_batch_apis()
+def clear_batch_indexes(index_names: List[str], marqo_config: config.Config = Depends(get_config)):
```
farshidz: I don't think we need this API. You can call `delete_docs` for each index at almost no extra cost (unlike create index, which is much slower if called multiple times due to Vespa deployment)

src/marqo/tensor_search/constants.py
```diff
@@ -4,7 +4,7 @@
 
 ILLEGAL_CUSTOMER_FIELD_NAME_CHARS = {'.', '/', '\n'}
 
-ALLOWED_CUSTOMER_FIELD_TYPES = [str, int, float, bool, list, dict]
+ALLOWED_UNSTRUCTURED_FIELD_TYPES = [str, int, float, bool, list]
```
farshidz, line 7: I can't find this used anywhere?

src/marqo/tensor_search/utils.py
```diff
@@ -339,6 +340,64 @@ def get_best_available_device() -> str:
     return device
 
 
+def is_tensor_field(field: str,
+                    tensor_fields: Optional[List[str]] = None
```
farshidz: Isn't `tensor_fields` compulsory?

src/marqo/tensor_search/utils.py
```diff
@@ -339,6 +340,64 @@ def get_best_available_device() -> str:
...
+                    ) -> bool:
+    """Determine whether a field is a tensor field or not for add_documents calls."""
+    if not tensor_fields:
+        return False
+    else:
+        return field in tensor_fields
+
+
+def calculate_health_status(marqo_os_health_check_response: Optional[Dict]) -> dict:
```
farshidz: This looks like something I've deleted in mainline and you've added back by mistake?

src/marqo/tensor_search/validation.py
```diff
@@ -255,6 +256,50 @@ def validate_field_name(field_name) -> str:
...
+    if not isinstance(field_name, str):
+        raise InvalidFieldNameError("field name must be str!")
+    if field_name.startswith(enums.TensorField.vector_prefix):
+        raise InvalidFieldNameError(F"can't start field name with protected prefix {enums.TensorField.vector_prefix}."
+                                    F" Error raised for field name: {field_name}")
+    if field_name.startswith(enums.TensorField.chunks):
+        raise InvalidFieldNameError(F"can't name field with protected field name {enums.TensorField.chunks}."
+                                    F" Error raised for field name: {field_name}")
+
+    if any(unsupported_string in field_name for unsupported_string in UNSUPPORTED_FIELD_NAME_LIST):
```
farshidz: You should do this validation `UnstructuredVespaIndex` class instead

src/marqo/tensor_search/validation.py
```diff
@@ -138,18 +140,15 @@ def validate_field_content(field_content: Any, is_non_tensor_field: bool) -> Any
     Raises:
         InvalidArgError if field_content is not acceptable
     """
-    if type(field_content) in constants.ALLOWED_CUSTOMER_FIELD_TYPES:
+    if type(field_content) in constants.ALLOWED_UNSTRUCTURED_FIELD_TYPES:
```
farshidz, line 142: Do this validation in `UnstructuredVespaIndex` class. Note this method is also used by structured index add docs, which highlights the perils of these utility-type functions (coupling that is really easy to miss -> regression)

src/marqo/tensor_search/validation.py
```diff
@@ -255,6 +256,50 @@ def validate_field_name(field_name) -> str:
...
+        raise InvalidFieldNameError("field name must be str!")
+    if field_name.startswith(enums.TensorField.vector_prefix):
+        raise InvalidFieldNameError(F"can't start field name with protected prefix {enums.TensorField.vector_prefix}."
+                                    F" Error raised for field name: {field_name}")
+    if field_name.startswith(enums.TensorField.chunks):
+        raise InvalidFieldNameError(F"can't name field with protected field name {enums.TensorField.chunks}."
+                                    F" Error raised for field name: {field_name}")
+
+    if any(unsupported_string in field_name for unsupported_string in UNSUPPORTED_FIELD_NAME_LIST):
+        raise InvalidArgError(
```
farshidz: Since `::` is the only problematic value, we can have a clear and specific error message that says `::` isn't allowed

src/marqo/tensor_search/validation.py
```diff
@@ -255,6 +256,50 @@ def validate_field_name(field_name) -> str:
...
+        field_name:
+
+    returns field_name, if all validations pass
+
+    Raises:
+        InvalidFieldNameError
+    """
+
+    # TODO-Li Do we still have these restrictions for an unstructured index?
+    if not field_name:
```
farshidz, line 272: Unstructured name validation should be consistent with structured name validation (plus the extra restriction of `::`). See `validate_structured_field` in `marqo_index`. I suggest extracting field name validation into a `validate_field_name` function then calling that for both unstructured and structured + plus extra validation of `::` for unstructured. As noted below, do all of this in `UnstructuredVespaIndex` instead of here

src/marqo/tensor_search/validation.py
```diff
@@ -482,3 +527,20 @@ def validate_delete_docs_request(delete_request: MqDeleteDocsRequest, max_delete
...
+
+def validate_result_count(result_count: int):
+    """Validates the result count (limit) for a search operation.
+
+    Args:
+        result_count: the result count to validate
+    Returns:
+        result_count, if nothing is raised
+    """
+    if not isinstance(result_count, int):
```
farshidz: You don't need type validation since these values pass through pydantic models and pydantic takes care of that

src/marqo/tensor_search/validation.py
```diff
@@ -482,3 +527,20 @@ def validate_delete_docs_request(delete_request: MqDeleteDocsRequest, max_delete
...
+        raise IllegalRequestedDocCount(f"result_count must be an int! Received {result_count} of type {type(result_count)}")
+
+    if result_count <= 0:
```
farshidz: I think the code will be more readable if you simply have this in `tensor_search.search` especially after removing the type check above

farshidz (changes requested):
Interim review. More comments will be added

src/marqo/tensor_search/api.py
wanliAlex: The issue is we have an API `POST /indexes/{index_name}` and there is routering problem

src/marqo/tensor_search/api.py
wanliAlex: But that API requires doc_ids, while this one is trying to remove all the documents in an index for API tests.

src/marqo/tensor_search/api.py
wanliAlex: done

src/marqo/tensor_search/api.py
wanliAlex: same, we have an api `POST /indexes/{index_name}`

src/marqo/tensor_search/api.py
wanliAlex: done

src/marqo/core/unstructured_vespa_index/constants.py
```diff
@@ -0,0 +1,7 @@
+# The threshold for a string to be considered short, long in unstructured Vespa.
+# Users will be able to filter on short strings, but not on long strings.
+SHORT_STRING_THRESHOLD = 20
```
farshidz: You could move constants to common to be consistent with structured index package

src/marqo/core/unstructured_vespa_index/constants.py
```diff
@@ -0,0 +1,7 @@
...
+
+UNSUPPORTED_FIELD_NAME_LIST = ["::"]
```
farshidz: Name is a bit misleading. Unsupported field substring? Or you could just hard-code this value since it's the only prohibited subtring (unless it's used in multiple places, then must be a constant).

src/marqo/core/unstructured_vespa_index/constants.py
```diff
@@ -0,0 +1,7 @@
...
+
+FILTER_STRING_BOOL_VALUE = ["true", "false"]
```
farshidz: This is only used in unstructured vespa index. It's best to keep constants local to the module if they're not shared by any other module. Helps with readability and cohesion

src/marqo/core/constants.py
wanliAlex: Removed now. They are in the dedicated `unstructured_document` module

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -0,0 +1,180 @@
...
+                                                    alias=unstructured_common.VESPA_DOC_MULTIMODAL_PARAMS)
+    vector_counts: int = Field(default=0, alias=unstructured_common.FIELD_VECTOR_COUNT)
+
+    match_features: Dict[str, Any] = Field(default_factory=dict, alias=unstructured_common.VESPA_DOC_MATCH_FEATURES)
+
+    # Fields that are excluded when generating vespa documents
+    _EXCLUDED_FIELDS = {"match_features"}
+
+    class Config:
+        allow_population_by_field_name = True
```
farshidz: Inherit `MarqoBaseModel` instead, which covers this feature (added recently).

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -0,0 +1,180 @@
...
+        try:
+            chunk_index: int = int(list(self.match_features[f"closest({unstructured_common.VESPA_DOC_EMBEDDINGS})"] \
+                                            ["cells"].keys())[0])
+        except KeyError:
+            raise ValueError("No match features found in the document")
+        field_name, content = self.vespa_chunks[chunk_index].split("::", 2)
+        return {field_name: content}
+
+
+class UnstructuredIndexDocument(BaseModel):
```
farshidz: Models should ideally be subclasses of `marqo.base_model` classes

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -0,0 +1,180 @@
...
+        marqo_document.update({k: bool(v) for k, v in self.bool_fields.items()})
+        marqo_document[index_constants.MARQO_DOC_ID] = self.marqo__id
+
+        # vespa_embeddings may not return if it comes from a search result
+        embeddings_list = []
+        if self.vespa_embeddings:
+            try:
+                embeddings_list = list(self.vespa_embeddings["blocks"].values())
+            except (KeyError, AttributeError, TypeError):
+                raise VespaDocumentParsingError(f"No embeddings found in the document _id = {self.marqo__id}")
```
farshidz: I wonder if the error message is accurate. Under what circumstances do you expect to reach here?

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -0,0 +1,180 @@
...
+
+        if self.vespa_chunks and embeddings_list:
```
farshidz: What if you have only chunks? e.g. for search we get chunks but don't need embeddings. Chunks are still needed for highlights

src/marqo/core/index_management/index_management.py
wanliAlex: done

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -0,0 +1,180 @@
...
+
+from pydantic import Field, BaseModel
+
+from marqo.core import constants as index_constants
+from marqo.core.exceptions import VespaDocumentParsingError
+from marqo.core.unstructured_vespa_index import common as unstructured_common
+from marqo.core.unstructured_vespa_index import constants as unstructured_constants
+
+
+class UnstructuredVespaDocumentFields(BaseModel):
```
farshidz: All `ValueErrors` that are raised during parsing should be `VespaDocumentParsingError` instead

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -0,0 +1,180 @@
...
+        embeddings_list = []
+        if self.vespa_embeddings:
+            try:
+                embeddings_list = list(self.vespa_embeddings["blocks"].values())
+            except (KeyError, AttributeError, TypeError):
+                raise VespaDocumentParsingError(f"No embeddings found in the document _id = {self.marqo__id}")
+
+        if self.vespa_chunks and embeddings_list:
+            if not len(self.vespa_chunks) == len(embeddings_list):
+                raise ValueError(f"Number of chunks and embeddings do not match for document _id= {self.marqo__id}")
```
farshidz: `VespaDocumentParsingError`

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -0,0 +1,180 @@
...
+            marqo_document[index_constants.MARQO_DOC_TENSORS] = dict()
+            for chunk, embedding in zip(self.vespa_chunks, embeddings_list):
+                if "::" not in chunk:
+                    raise ValueError(f"Chunk {chunk} does not have a field_name::content format")
```
farshidz: `VespaDocumentParsingError`

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -0,0 +1,180 @@
...
+                    json.loads(serialized_multimodal_params)
+
+        if return_highlights and self.match_features:
+            marqo_document[index_constants.MARQO_DOC_HIGHLIGHTS] = self._extract_highlights()
+
+        return marqo_document
+
+    def _extract_highlights(self):
+        if not self.match_features:
+            raise ValueError("No match features found in the document")
```
farshidz: `VespaDocumentParsingError`

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -0,0 +1,180 @@
...
+        try:
+            chunk_index: int = int(list(self.match_features[f"closest({unstructured_common.VESPA_DOC_EMBEDDINGS})"] \
+                                            ["cells"].keys())[0])
+        except KeyError:
+            raise ValueError("No match features found in the document")
```
farshidz: `VespaDocumentParsingError`

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -0,0 +1,180 @@
...
+        return cls(id=document[index_constants.VESPA_DOC_ID],
+                   fields=UnstructuredVespaDocumentFields(**fields))
+
+    @classmethod
+    def from_marqo_document(cls, document: Dict) -> "UnstructuredIndexDocument":
+        """Instantiate an UnstructuredIndexDocument from a valid Marqo document from
+        add_documents"""
+
+        if index_constants.MARQO_DOC_ID not in document:
+            raise ValueError(f"Unstructured Marqo document does not have a {index_constants.MARQO_DOC_ID} field. "
```
farshidz: `MarqoDocumentParsingError`

src/marqo/tensor_search/constants.py
wanliAlex: Validation. But yes, it will be moved to unstructured_index part

src/marqo/core/unstructured_vespa_index/unstructured_document.py
farshidz: I find class names a bti confusing here. How about `UnstructuredVespaDocument` and `UnstructuredVespaDocumentFields` (unchanged)?

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+    id: str = Field(alias=index_constants.VESPA_DOC_ID)
+    fields: UnstructuredVespaDocumentFields = Field(default_factory=UnstructuredVespaDocumentFields,
+                                                    alias=index_constants.VESPA_DOC_FIELDS)
+
+    class Config:
+        allow_population_by_field_name = True
+
+    @classmethod
+    def from_vespa_document(cls, document: Dict) -> "UnstructuredIndexDocument":
```
farshidz: Why do you need this? Shouldn't `UnstructuredIndexDocument(**vespa_doc_dict)` work? i.e. native pydantic deserialization

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -0,0 +1,180 @@
...
+
+    class Config:
+        allow_population_by_field_name = True
+
+    def to_vespa_dictionary(self) -> Dict[str, Any]:
+        # Exclude None and empty fields but keep 0 as we need to have vector_counts in docs
+        return {k: v for k, v in self.dict(exclude_none=True, by_alias=True,
+                                           exclude=self._EXCLUDED_FIELDS).items() if v or v == 0}
+
+    def to_marqo_doc(self, return_highlights: bool = False) -> Dict[str, Any]:
```
farshidz: I think this should be in the other class next to `from_marqo_document` (and named `to_marqo_document`)

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -0,0 +1,180 @@
...
+                    instance.fields.long_string_fields[key] = value
+                instance.fields.strings.append(value)
+            elif isinstance(value, bool):
+                instance.fields.bool_fields[key] = int(value)
+            elif isinstance(value, list) and all(isinstance(elem, str) for elem in value):
+                instance.fields.string_arrays.extend([f"{key}::{element}" for element in value])
+            elif isinstance(value, int):
+                instance.fields.int_fields[key] = value
+                instance.fields.score_modifiers_fields[key] = value
+            elif isinstance(value, float):
```
farshidz, line 93: It's always good to have a final `else:` and error out (parsing error here), in case you get unexpected input. Makes potential bugs in upstream code much easier to catch

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -0,0 +1,180 @@
...
+    @classmethod
+    def from_marqo_document(cls, document: Dict) -> "UnstructuredIndexDocument":
+        """Instantiate an UnstructuredIndexDocument from a valid Marqo document from
+        add_documents"""
+
+        if index_constants.MARQO_DOC_ID not in document:
+            raise ValueError(f"Unstructured Marqo document does not have a {index_constants.MARQO_DOC_ID} field. "
+                             f"This should be assigned for a valid document")
+
+        copied = deepcopy(document)
```
farshidz: I assume you're copying because you want to delete `index_constants.MARQO_DOC_ID`. Instead, you can skip `index_constants.MARQO_DOC_ID` in your loop and avoid the copy operation (which costs a full pass over the data)

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -0,0 +1,180 @@
...
+    # Fields that are excluded when generating vespa documents
+    _EXCLUDED_FIELDS = {"match_features"}
+
+    class Config:
+        allow_population_by_field_name = True
+
+    def to_vespa_dictionary(self) -> Dict[str, Any]:
+        # Exclude None and empty fields but keep 0 as we need to have vector_counts in docs
+        return {k: v for k, v in self.dict(exclude_none=True, by_alias=True,
+                                           exclude=self._EXCLUDED_FIELDS).items() if v or v == 0}
```
farshidz: Do you need this exclusion? You will only have `matchfeatures` in search response

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def to_vespa_document(self) -> Dict[str, Any]:
+        """Convert VespaDocumentObject to a Vespa document.
+        Empty fields are removed from the document."""
+        return {index_constants.VESPA_DOC_ID: self.id,
+                index_constants.VESPA_DOC_FIELDS: self.fields.to_vespa_dictionary()}
+
+    def to_marqo_document(self, return_highlights: bool = False) -> Dict[str, Any]:
+        """Convert VespaDocumentObject back to the original document structure."""
+        return self.fields.to_marqo_doc(return_highlights=return_highlights)
```
farshidz: As noted, I think you should move the logic here and out of the other class

farshidz (changes requested):

src/marqo/tensor_search/validation.py
wanliAlex: done

src/marqo/tensor_search/validation.py
wanliAlex: done

src/marqo/tensor_search/validation.py
wanliAlex: done

src/marqo/core/unstructured_vespa_index/constants.py
wanliAlex: done

src/marqo/core/unstructured_vespa_index/constants.py
wanliAlex: done

src/marqo/core/unstructured_vespa_index/constants.py
wanliAlex: done, removed

src/marqo/core/unstructured_vespa_index/unstructured_document.py
wanliAlex: done

src/marqo/core/unstructured_vespa_index/unstructured_document.py
wanliAlex: done

src/marqo/core/unstructured_vespa_index/unstructured_document.py
wanliAlex: Yes, it is not necessary.

src/marqo/core/unstructured_vespa_index/unstructured_document.py
wanliAlex: done

src/marqo/core/unstructured_vespa_index/unstructured_document.py
wanliAlex: done

src/marqo/core/unstructured_vespa_index/unstructured_document.py
wanliAlex: done

src/marqo/core/unstructured_vespa_index/unstructured_document.py
wanliAlex: done

src/marqo/tensor_search/validation.py
wanliAlex: Done

src/marqo/tensor_search/validation.py
wanliAlex: Done

src/marqo/tensor_search/utils.py
wanliAlex: deleted

src/marqo/tensor_search/utils.py
wanliAlex: _yes_

src/marqo/tensor_search/validation.py
wanliAlex: done

src/marqo/core/unstructured_vespa_index/unstructured_document.py
wanliAlex: I think it should not reach here unless the data from vespa is broken, maybe raise an internal error?

src/marqo/core/unstructured_vespa_index/unstructured_document.py
wanliAlex: It is handled in the function `extract_highlights()`. I agree the logic flow in this part is not clear.

src/marqo/core/unstructured_vespa_index/unstructured_document.py
wanliAlex: done

src/marqo/core/unstructured_vespa_index/unstructured_document.py
wanliAlex: updated

src/marqo/core/unstructured_vespa_index/unstructured_document.py
wanliAlex: done

src/marqo/core/unstructured_vespa_index/unstructured_document.py
wanliAlex: Just like a counterpart to the from_marqo_document so it is easier to read?

src/marqo/tensor_search/api.py
farshidz: Let's do `POST /batch/indexes/delete` then since the previous suggestion clashes with existing API

src/marqo/tensor_search/api.py
farshidz: To avoid clashing with exisiting APIs let's use `POST /batch/indexes`

src/marqo/tensor_search/api.py
farshidz: Then let's have an API that clears a single index. `DELETE /index/{index_name}/clear`. Let's review these new API names with the team later

farshidz (approved):

farshidz:
Comments to address as a follow-up PR:

index_management.py:batch_delete_indexes

No right or wrong here, but we don’t routinely type check function arguments as it’ll be arduous and we have type annotations (even though they’re not enforced at runtime unfortunately). Keeps the code simpler. We type check when it’s user input though (not the case here)

api.py:batch_clear_indexes

As noted, I think an API that clears a single index is more appropriate as it’s more flexible (e.g., if a test needs to clear just one index)
More importantly, 
* API call should go to core.document based on Marqo v2 design. We shouldn’t be expanding tensor_search
*  We don’t use pyvespa except in tests. In this case we need to add the required functionality to Marqo’s Vespa client. Note that pyvespa is currently a dev dependency only so this will fail when we build the image

validation.py:validate_unstructured_index_field_name — unused code. delete

— I think field name validation is still inconsistent across structured and unstructured. Please review this. Looks like v1 validation that may be irrelevant is still applied 

— I think it’s nice that you removed SHORT_STRING_THRESHOLD and made it an index config. Let’s add it to IndexSettings so user can configure it (and move the default value to IndexSettings definition like other defaults)

unstructured_document.py:87 I think this should be bool(value) ?
unstructured_document.py:97 You should be raising MarqoDocumentParsingError when failing to parse a Marqo document

PR #663, opened 2023-12-12
Patch delete documents

wanliAlex:


src/marqo/vespa/vespa_client.py
```diff
@@ -36,7 +36,8 @@ def __init__(self, current_generation: int, wanted_generation: int, converged: b
             self.wanted_generation = wanted_generation
             self.converged = converged
 
-    def __init__(self, config_url: str, document_url: str, query_url: str, pool_size: int = 10):
+    def __init__(self, config_url: str, document_url: str, query_url: str, pool_size: int = 10,
+                 content_cluster_name: str = "content_default"):
```
farshidz: I don't recommend having a default content cluster here. Something like `pool_size` has a sensible default value that always works. That's not the case for content cluster.

src/marqo/vespa/models/delete_document_response.py
```diff
@@ -18,3 +18,8 @@ class DeleteBatchDocumentResponse(BaseModel):
 class DeleteBatchResponse(BaseModel):
     responses: List[DeleteBatchDocumentResponse]
     errors: bool
+
+
+class DeleteAllDocumentResponse(BaseModel):
```
farshidz: `DeleteAllDocumentsResponse` (Documents plural)

farshidz (changes requested):

farshidz (approved):

PR #665, opened 2023-12-13
Fix multimodal field for unstructured index

wanliAlex:


src/marqo/tensor_search/tensor_search.py
```diff
@@ -141,10 +146,15 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
                         f"images for {batch_size} docs using {add_docs_params.image_download_thread_count} threads"
                     )
             ):
+                # TODO - Refactor this part to make it more readable
+                # We need to pass the subfields to the image downloader, so that it can download the images in the
+                # multimodal subfields even if the subfield is not a tensor_field
+                tensor_fields_and_multimodal_subfields = copy.deepcopy(add_docs_params.tensor_fields) \
```
farshidz, line 149: Why are you doing a deep copy?

farshidz, line 149: If you call this `image_fields` it might be easier to understand

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
```diff
@@ -300,6 +300,14 @@ def _validate_list(field_content: list, is_tensor_field: bool) -> None:
 
         return
 
+    @classmethod
+    def validate_multimodal_field_name(cls, multimodal_field_name:str) -> None:
```
farshidz: You shouldn't add public methods outside the abstraction's interface (`VespaIndex` class). This applies to `validate_multimodal_field_name` and `validate_field_name`. Otherwise the abstraction becomes unusable (see Liskov substitution principle). Also have you noticed the field name validation code we already have in `marqo_index.py`?

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
```diff
@@ -300,6 +300,14 @@ def _validate_list(field_content: list, is_tensor_field: bool) -> None:
...
+        cls.validate_field_name(multimodal_field_name)
+        # This error will never be raised because _id can't be a tensor_fields
+        # TODO check if there are other validations needed for multimodal_field_name
+        if multimodal_field_name == "_id":
```
farshidz: What is special about multimodal, since we're not doing this validation for other fields?

farshidz (dismissed):

src/marqo/tensor_search/tensor_search.py
```diff
@@ -345,10 +355,32 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
             # All the plain tensor/non-tensor fields are processed, now we process the multimodal fields
             if document_is_valid and add_docs_params.mappings:
                 multimodal_mappings: Dict[str, Dict] = utils.extract_multimodal_mappings(add_docs_params.mappings)
+
                 for field_name, multimodal_params in multimodal_mappings.items():
                     if not utils.is_tensor_field(field_name, add_docs_params.tensor_fields):
                         raise errors.InvalidArgError(f"Multimodal field {field_name} must be a tensor field")
 
+                    try:
+                        unstructured_vespa_index.validate_multimodal_field_name(field_name)
```
farshidz: Do this validation before any inference as there's no point if the doc is invalid

farshidz (changes requested):

farshidz (approved):

PR #670, opened 2023-12-18
Embedding similarity uses closeness max function

vicilliar:
Bug fix

In structured index schema rank profile, tensor search similarity uses SUM (+) on embedding closeness per field, instead of MAX.

Similarity is now calculated with MAX. Uses binary max tree since max function is only a binary function.
`test_structured_vespa_schema` and `test_structured_vespa_index` tests are now fixed by adding vector count field to test schemas.

no

yes

* **Other information**:

- x] Tests for the changes have been added (for bug fixes/features)
- [] Docs have been added / updated (for bug fixes / features)


src/marqo/core/structured_vespa_index/structured_vespa_schema.py
```diff
@@ -158,6 +158,34 @@ def _generate_document_section(self) -> (List[str], StructuredMarqoIndex):
 
         return document, marqo_index
 
+    def _generate_max_similarity_expression(self, tensor_fields: List[TensorField]) -> str:
+        """
+        Recursively generate max closeness expression for all tensor fields.
+        Because max
```
farshidz: This doesn't read well?

src/marqo/core/structured_vespa_index/structured_vespa_schema.py
```diff
@@ -158,6 +158,34 @@ def _generate_document_section(self) -> (List[str], StructuredMarqoIndex):
...
+        If more than 2, get max of first and (max of the rest).
+        """
+
+        # Base cases
+        # If no tensor fields, return empty string.
+        # If only 1 or 2 tensor fields, get max of one/both.
+        if len(tensor_fields) == 0:
+            return ""
+        elif len(tensor_fields) == 1:
+            return (f'if(query({tensor_fields[0].name}) > 0, '
```
farshidz, line 175: I'm assuming `max` won't accept one argument? e.g., `max(value1)`? Because if it did, you would only need one base case

farshidz: Though this is still better. We don't know if an unnecessary max is going to cause inefficiency, even if it is mathematically inconsequential

tests/core/structured_vespa_index/test_structured_vespa_schema.py
```diff
@@ -149,3 +149,66 @@ def _remove_whitespace_in_schema(self, schema: str) -> str:
         schema = re.sub(r'^\s+', '', schema, flags=re.MULTILINE)
 
         return schema
+
+
+class TestStructuredVespaSchemaMethods(MarqoTestCase):
```
farshidz: I would argue that you don't need these tests.  You can add full schema tests like the ones you fixed above to cover these cases if you find these necessary. As a rule of thumb, we don't write unit tests for private methods -- we only test the public interface. If you find it impossible to get good coverage without unit-testing private methods, then chances are your design isn't cohesive (which is not the case, your design is fine here). Check this out also https://chat.openai.com/share/d6c43184-1ad5-49b5-b209-e659bdb68c33

farshidz (dismissed):

src/marqo/core/structured_vespa_index/structured_vespa_schema.py
vicilliar: Yep according to documentation, max is for 2 arguments: https://docs.vespa.ai/en/reference/ranking-expressions.html#max-x

tests/core/structured_vespa_index/test_structured_vespa_schema.py
vicilliar: Got this, I'll note this for future private methods. That being said, should I just keep them in since I already wrote them? Or would you prefer them removed?

vicilliar: I would argue though that it's cumbersome to make 5 full new schemas just to test the output of 5 different cases of `_generate_max_similarity_expression`. It seems like a lot of wasted space.

farshidz (approved):

PR #685, opened 2024-01-04
Patch image preprocessing

wanliAlex:
Add marqo-yolo, dino-v1, dino-v2 to image preprocessing


src/marqo/tensor_search/tensor_search.py
```diff
@@ -1196,9 +1196,9 @@ def search(config: Config, index_name: str, text: Union[str, dict],
 
     if reranker is not None:
         logger.info("reranking using {}".format(reranker))
-        if searchable_attributes is None:
+        if searchable_attributes is None or searchable_attributes == []:
             raise errors.InvalidArgError(
-                f"searchable_attributes cannot be None when re-ranking. Specify which fields to search and rerank over.")
+                f"searchable_attributes cannot be None or [] when re-ranking. Specify which fields to search and rerank over.")
```
farshidz: This is different from v1 behavior, isn't it?

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -220,7 +220,7 @@ def _to_vespa_tensor_query(self, marqo_query: MarqoTensorQuery) -> Dict[str, Any
         else:
             fields_to_search = self._marqo_index.tensor_field_map.keys()
 
-        tensor_term = self._get_tensor_search_term(marqo_query)
+        tensor_term = self._get_tensor_search_term(marqo_query) if fields_to_search else "False"
```
farshidz, line 223: This is a string not a boolean

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -267,7 +267,7 @@ def _to_vespa_lexical_query(self, marqo_query: MarqoLexicalQuery) -> Dict[str, A
         else:
             fields_to_search = self._marqo_index.lexical_field_map.keys()
 
-        lexical_term = self._get_lexical_search_term(marqo_query)
+        lexical_term = self._get_lexical_search_term(marqo_query) if fields_to_search else "False"
```
farshidz, line 274: This is a string

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -326,7 +326,7 @@ def _get_tensor_search_term(self, marqo_query: MarqoTensorQuery) -> str:
         if terms:
             return f'({" OR ".join(terms)})'
         else:
-            return ''
+            return ""
```
farshidz: What's changing here?

tests/tensor_search/test_searchable_attributes.py
```diff
@@ -0,0 +1,201 @@
...
+from tests.marqo_test import MarqoTestCase
+
+
+class TestSearchStructured(MarqoTestCase):
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        super().setUpClass()
+
+        default_text_index = cls.structured_marqo_index_request(
```
farshidz, line 17: Looks like you only need one of these indexes. I would use a much simpler index as the extra fields aren't relevant for this test. Also I think you should test with an unstructured index too.

farshidz (changes requested):

farshidz (dismissed):

farshidz (approved):

PR #686, opened 2024-01-06
kebab-case index names

farshidz:
Add support for - (dash) in index name


src/marqo/core/index_management/index_management.py
```diff
@@ -115,63 +118,55 @@ def delete_index(self, marqo_index: MarqoIndex) -> None:
...
             marqo_index: Marqo index to delete
         """
-        self.delete_index_by_name(marqo_index.name)
+        app = self.vespa_client.download_application()
+
+        self._remove_schema(app, marqo_index.schema_name)
+        self._remove_schema_from_services(app, marqo_index.schema_name)
+        self._add_schema_removal_override(app)
+        self.vespa_client.deploy_application(app)
+        self._delete_index_settings_by_name(marqo_index.name)
```
pandu-k, line 132: these look like a lot of steps that depend on unreliable hardware (network) - are there adequate safeguards in case any step fails?

tests/tensor_search/integ_tests/test_add_documents_structured.py
pandu-k, line 1: Do we have enough edge cases tested? we can use gpt to spin up some new test cases with more edge cases

src/marqo/core/index_management/index_management.py
farshidz: The process is idempotent. Failure of intermediate steps can create an inconsistent state (e.g., Vespa schema doesn't exist, but Marqo thinks the index exists), but this can be recovered from by calling the API again and getting a 200 response.

tests/tensor_search/integ_tests/test_add_documents_structured.py
farshidz: If you mean for index name encoding, I find the testing sufficient. I can't think of any edge cases not covered.

src/marqo/core/index_management/index_management.py
pandu-k: do the errors the users get descriptive enough for them to take the correct action? 

And, in these inconsistent states, is it ever possible to get into a catastrophic unrecoverable state? Let's say that another user is doing other odd index management requests concurrently?

pandu-k (approved):

pandu-k (approved):

PR #693, opened 2024-01-09
Fix structured bool type bug

farshidz:


tests/tensor_search/integ_tests/test_search_structured.py
```diff
@@ -575,6 +575,10 @@ def test_filtering(self):
             ("int_field_1:[5 TO 30]", 0, None),
             ("int_field_1:[0 TO 30]", 1, "1234"),
             ("bool_field_1:true", 1, "1233"),
+            ("bool_field_1:True", 1, "1233"),
+            ("bool_field_1:tRue", 1, "1233"),
+            ("bool_field_1:false", 0, None),  # no hits for bool_field_1=false
+            ("bool_field_1:some_value", 0, None),  # no hits for bool_field_1 not boolean
```
pandu-k, line 586: maybe create some tests for some docs when bool_field_1=False?

pandu-k: maybe in a separate test case

tests/tensor_search/integ_tests/test_search_unstructured.py
```diff
@@ -689,7 +700,8 @@ def test_lexical_filtering(self):
             ("field1:(other things)", 1, "123456"),
             ("fake_int:234", 1, "other doc"),
             ("fake_float:1.23", 1, "other doc"),
-            ("gapped_field_name:gap", 1, "other doc")
+            ("gapped_field_name:gap", 1, "other doc"),
```
pandu-k, line 703: qq we i don't see any field names with spaces - are we sure this works?

farshidz: It won't. We don't support spaces in field names

pandu-k (approved):

PR #697, opened 2024-01-09
Patch highlights format to List[Dict]

wanliAlex:
Patch highlights format to List[Dict]

### DIFF
--- a/src/marqo/s2_inference/reranking/rerank.py
+++ b/src/marqo/s2_inference/reranking/rerank.py
@@ -5,12 +5,12 @@ from marqo.s2_inference.errors import RerankerError, RerankerNameError
 from marqo.s2_inference.logger import get_logger
 from marqo.s2_inference.reranking.cross_encoders import ReRankerText, ReRankerOwl
 from marqo.s2_inference.reranking.enums import ResultsFields
-from marqo.s2_inference.types import Dict, List
+from marqo.s2_inference.types import Dict, List, Optional
 
 logger = get_logger(__name__)
 
 def rerank_search_results(search_result: Dict, query: str, model_name: str, device: str, 
-                searchable_attributes: List[str], num_highlights: int = 1,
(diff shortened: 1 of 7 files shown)


pandu-k (dismissed):

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1312,7 +1312,7 @@ def _lexical_search(
     # Set the _highlights for each doc as [] to follow Marqo-V1's convention
     if highlights:
         for docs in gathered_docs['hits']:
-            docs['_highlights'] = {}
+            docs['_highlights'] = [{}]
```
pandu-k: this needs to be an empty list - an empty dict inside an empty list means that it is no longer an empty list

src/marqo/s2_inference/reranking/rerank.py
```diff
@@ -1,15 +1,16 @@
...
 from marqo.s2_inference.logger import get_logger
+from marqo.s2_inference.reranking.cross_encoders import ReRankerText, ReRankerOwl
+from marqo.s2_inference.reranking.enums import ResultsFields
+from marqo.s2_inference.types import Dict, List
 
 logger = get_logger(__name__)
 
 def rerank_search_results(search_result: Dict, query: str, model_name: str, device: str, 
-                searchable_attributes: List[str] = None, num_highlights: int = 1, 
+                searchable_attributes: List[str], num_highlights: int = 1,
```
farshidz: Why did you remove the default value here? A bit concerned about changes here as I don't believe we have test coverage?

src/marqo/tensor_search/tensor_search.py
wanliAlex: Done

src/marqo/s2_inference/reranking/rerank.py
wanliAlex: There is a value check before called this function. But I can add it back.

pandu-k (approved):

PR #699, opened 2024-01-09
Update default efSearch and fix pagination bug 

farshidz:


src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -314,17 +314,25 @@ def _get_tensor_search_term(self, marqo_query: MarqoTensorQuery) -> str:
...
-                f'({{targetHits:{marqo_query.limit}, approximate:{str(marqo_query.approximate)}{additional_hits}}}'
-                f'nearestNeighbor({embedding_field_name}, {common.QUERY_INPUT_EMBEDDING}))'
+                f'('
+                f'{{'
+                f'targetHits:{target_hits}, '
+                f'approximate:{str(marqo_query.approximate)}, '
+                f'hnsw.exploreAdditionalHits:{additional_hits}'
+                f'}}'
+                f'nearestNeighbor({embedding_field_name}, {common.QUERY_INPUT_EMBEDDING})'
+                f')'
```
pandu-k, line 337: What about the case where limit = 10, offset = 1000, and efsearch=512
Won't target + additional=512? Don't we need a bigger efsearch to get further into the hit list?

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
pandu-k, line 1: ditto above comment

src/marqo/tensor_search/configs.py
```diff
@@ -29,5 +29,6 @@ def default_env_vars() -> dict:
         EnvVars.MARQO_MAX_DELETE_DOCS_COUNT: 10000,
         EnvVars.MARQO_MAX_SEARCHABLE_TENSOR_ATTRIBUTES: None,
         EnvVars.MARQO_MAX_NUMBER_OF_REPLICAS: 1,
+        EnvVars.MARQO_DEFAULT_EF_SEARCH: 2000,
```
pandu-k, line 32: are we happy with the throughput at this level?

src/marqo/core/structured_vespa_index/structured_vespa_index.py
pandu-k: Ok I see there is search logic ensuring that efSearch is always >= offset + limit

pandu-k: we need to ensure this is clearly specified in documentation

pandu-k (approved):

src/marqo/core/structured_vespa_index/structured_vespa_index.py
farshidz: Yes if efSearch is too low they'll get empty hits

src/marqo/tensor_search/configs.py
farshidz: We are. Performance drop was minor

PR #700, opened 2024-01-09
Patch use existing tensor

wanliAlex:
Patch use existing tensor

### DIFF
--- a/src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
+++ b/src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
@@ -11,6 +11,7 @@ from marqo.core.unstructured_vespa_index import common as unstructured_common
 from marqo.core.unstructured_vespa_index.unstructured_document import UnstructuredVespaDocument
 from marqo.core.vespa_index import VespaIndex
 from marqo.exceptions import InternalError
+import marqo.core.exceptions as core_exceptions
 
 
 class UnstructuredVespaIndex(VespaIndex):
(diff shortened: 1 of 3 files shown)


src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
```diff
@@ -231,6 +232,11 @@ def _get_lexical_search_term(marqo_query: MarqoLexicalQuery) -> str:
 
             return f'{or_terms}{and_terms}'
 
+        if marqo_query.score_modifiers is not None:
```
farshidz: `VespaIndex` classes do support score modifiers for lexical search. We can do this validation somewhere else (closer to API), but also we could just leave it out for the time being.

tests/tensor_search/test_add_documents_use_existing_tensors.py
```diff
@@ -1,584 +1,640 @@
...
+
+                from pprint import pprint
+                pprint(search_res)
+                pprint(get_doc_res)
+                self.assertEqual("content 2", search_res["hits"][0]["text_field_2"])
+                self.assertEqual(1, len(get_doc_res["_tensor_facets"]))
+                self.assertEqual("content 2", get_doc_res["_tensor_facets"][0]["text_field_2"])
+
+
+    # def test_use_existing_tensors_resilience(self):
```
farshidz: Please use skip flags instead of commenting out tests. This way the diff will be much cleaner

farshidz (changes requested):

tests/tensor_search/test_add_documents_use_existing_tensors.py
wanliAlex: done

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
wanliAlex: Done. I have a look at the code again and I think score modifiers is not included for structured index either?

farshidz (approved):

PR #701, opened 2024-01-09
Fix snake case issue in get settings and create index

farshidz:
* Return camelCase in get settings
* Reject snake case in create index

### DIFF
--- a/src/marqo/tensor_search/api.py
+++ b/src/marqo/tensor_search/api.py
@@ -293,7 +293,7 @@ def get_indexes(marqo_config: config.Config = Depends(get_config)):
 @app.get("/indexes/{index_name}/settings")
 def get_settings(index_name: str, marqo_config: config.Config = Depends(get_config)):
     marqo_index = marqo_config.index_management.get_index(index_name)
-    return IndexSettings.from_marqo_index(marqo_index).dict(exclude_none=True)
+    return IndexSettings.from_marqo_index(marqo_index).dict(exclude_none=True, by_alias=True)
 
 
 @app.get("/models")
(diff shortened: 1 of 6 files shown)


src/marqo/tensor_search/api.py
```diff
@@ -293,7 +293,7 @@ def get_indexes(marqo_config: config.Config = Depends(get_config)):
 @app.get("/indexes/{index_name}/settings")
 def get_settings(index_name: str, marqo_config: config.Config = Depends(get_config)):
     marqo_index = marqo_config.index_management.get_index(index_name)
-    return IndexSettings.from_marqo_index(marqo_index).dict(exclude_none=True)
+    return IndexSettings.from_marqo_index(marqo_index).dict(exclude_none=True, by_alias=True)
```
pandu-k, line 296: confused by the `by_alias=True` as it looks like the IndexSettings object is already in camel?

pandu-k (commented):
could we please have a test or two that helps ensure the desired behaviour?

src/marqo/tensor_search/models/index_settings.py
```diff
@@ -40,6 +42,20 @@ class IndexSettings(StrictBaseModel):
         )
     )
 
+    @root_validator(pre=True)
+    def validate_field_names(cls, values):
+        # Verify no snake case field names (pydantic won't catch these due to allow_population_by_field_name = True)
+        def validate_dict_keys(d: dict):
+            for key in d.keys():
+                if '_' in key:
+                    raise ValueError(f"Invalid field name '{key}'")
```
pandu-k: A link to relevant documentation would be very helpful

src/marqo/tensor_search/api.py
farshidz: Other models like `TextPreProcessing` are in snake_case since they're not API-level. This is so that they serialise into camelCase

pandu-k (approved):

PR #717, opened 2024-01-16
Fix snake case vaidation bug in allFields

farshidz:
Fix bug where snake_case is accepted (rather than fail validation) for `dependent_fields`. 

`dependent_fields` is accepted.

Validation error for `dependent_fields`. Must be `dependentFields`

No

Yes

None

None

* **Other information**:


src/marqo/tensor_search/models/index_settings.py
```diff
@@ -45,15 +45,19 @@ class IndexSettings(StrictBaseModel):
...
-                                     f"See Create Index API reference here https://docs.marqo.ai/2.0.0/API-Reference/Indexes/create_index/")
-                if isinstance(d[key], dict):
-                    validate_dict_keys(d[key])
-
-        validate_dict_keys(values)
+        def validate_keys(d: Union[dict, list]):
+            if isinstance(d, dict):
+                for key in d.keys():
+                    if '_' in key:
+                        raise ValueError(f"Invalid field name '{key}'. "
```
wanliAlex, line 52: Should we explicitly specify that the error is related to `snake_case` or `camelCase` in IndexSettings field name here to avoid confusion? E.g.,there is also a `name` in `allFields`

wanliAlex (changes requested):

src/marqo/tensor_search/models/index_settings.py
farshidz: I thought that might be too much information and confusing. Ideally our pydantic model should error out for these fields like any other invalid field name. The reason we need this hack is our use of aliases. I think this difference in implementation shouldn't affect the user experience and error message.

wanliAlex (approved):

PR #722, opened 2024-01-18
Add support for long double fields

wanliAlex:
Feature:
Unstructured index now default use long, double to replace int, float
Structured index now support create fields with type `long`, `double`, `array<long>`, `array<double>`

We use int, float for unstructured index
we only support int, float, array<int>, array<float> in structured index

Unstructured index now default use long, double to replace int, float
Structured index now support create fields with type `long`, `double`, `array<long>`, `array<double>`

no
no
no

no

no

* **Other information**:
no


tests/tensor_search/integ_tests/test_add_documents_structured.py
```diff
@@ -995,3 +1015,32 @@ def test_download_images_thread_count(self):
...
+            ({"long_field_1": 1}, False),  # small positive integer
+            ({"long_field_1": -1}, False),  # small negative integer
+            ({"long_field_1": 100232142864}, False),  # large positive integer that can't be handled by int
+            ({"long_field_1": -923217213}, False),  # large negative integer that can't be handled by int
+            ({'long_field_1': int("1" * 50)}, True),  # overlarge positive integer, should raise error in long field
+            # overlarge negative integer, should raise error in long field
+            ({'long_field_1': -1 * int("1" * 50)}, True),
+            ({"double_field_1": 1e10}, False),  # large positive integer mathematical expression
+            ({"double_field_1": -1e12}, False),  # large negative integer mathematical expression
+            ({"double_field_1": 1e10 + 0.123249357987123}, False),  # large positive float
```
farshidz: Are these the min/max acceptable values?

farshidz (changes requested):
* Add input validation so out of range values don't cause 500. Input validation must take index version into account for unstructured (i.e. older indexes don't support double/long so lower limits)
* Ensure values at limit boundaries are tested, and also test past boudaries we get 400

wanliAlex:
> * Add input validation so out of range values don't cause 500. Input validation must take index version into account for unstructured (i.e. older indexes don't support double/long so lower limits)
> * Ensure values at limit boundaries are tested, and also test past boudaries we get 400

I will add test cases at the boundaries.

As for the index version, I will leave it to manual, integration tests.

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -500,6 +508,52 @@ def _verify_marqo_field_type(self, field_name: str, value: Any):
...
+        MAX_FLOAT = 3.4028235e38
+        MIN_FLOAT = -3.4028235e38
+        SMALLEST_POS_FLOAT = 1.4e-45
+        LARGEST_NEG_FLOAT = -1.4e-45
+        if (value > MAX_FLOAT) or (0 < value < SMALLEST_POS_FLOAT) or (value < MIN_FLOAT) or \
+                (0 > value > LARGEST_NEG_FLOAT):
+            raise InvalidDataRangeError(f'Invalid value {value} for float field. Expected a value in the range '
+                                        f'[{MIN_FLOAT}, {MAX_FLOAT}] or [{LARGEST_NEG_FLOAT}, {SMALLEST_POS_FLOAT}], but '
+                                        f'found {value} '
+                                        f'If you wish to store a value outside of this range, creating a field with type '
```
farshidz: "create a field with type"

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -500,6 +508,52 @@ def _verify_marqo_field_type(self, field_name: str, value: Any):
...
+            self._verify_long_field_range(value)
+        elif index_field.type in (FieldType.Double, FieldType.ArrayDouble):
+            pass
+        else:
+            raise InternalError(f'Invalid field type {index_field.type} for field {index_field.name} called by'
+                                f'_verify_numerical_field_value. Expected one of {FieldType.Float}, {FieldType.Int}, '
+                                f'{FieldType.Long}, {FieldType.Double}.')
+
+    def _verify_float_field_range(self, value: float):
+        MAX_FLOAT = 3.4028235e38
```
farshidz: Best to define these at class level similar to `_VESPA_DOC_ID` so that they're stored only once not per function call, and also for clarity

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -500,6 +508,52 @@ def _verify_marqo_field_type(self, field_name: str, value: Any):
...
+        MIN_FLOAT = -3.4028235e38
+        SMALLEST_POS_FLOAT = 1.4e-45
+        LARGEST_NEG_FLOAT = -1.4e-45
+        if (value > MAX_FLOAT) or (0 < value < SMALLEST_POS_FLOAT) or (value < MIN_FLOAT) or \
```
farshidz: What happens if `< SMALLEST_POS_FLOAT` or `> LARGEST_NEG_FLOAT`? If truncation/rounding, I think that is acceptable and we shouldn't error out for that

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -500,6 +508,52 @@ def _verify_marqo_field_type(self, field_name: str, value: Any):
...
+            raise InvalidDataRangeError(f"Invalid value {value} for int field. Expected a value in the range "
+                                        f"[{MIN_INT}, {MAX_INT}], but found {value}. "
+                                        f"If you wish to store a value outside of this range, creating a field with type "
+                                        f"'{FieldType.Long} or '{FieldType.Double}'. ")
+
+    def _verify_long_field_range(self, value: int):
+        MAX_LONG = 9223372036854775807
+        MIN_LONG = -9223372036854775808
+        if value > MAX_LONG or value < MIN_LONG:
+            raise InvalidDataRangeError(f"Invalid value {value} for long field. Expected a value in the range "
```
farshidz, line 557: Does this actually work? Is Vespa's long range smaller than Python's?

tests/tensor_search/test_api.py
```diff
@@ -115,11 +116,8 @@ def test_invalid_data_type(self):
                 }
             ]
         })
-
-        self.assertEqual(response.status_code, 400)
-        self.assertEqual(response.json()["code"], "invalid_argument")
-        self.assertEqual(response.json()["type"], "invalid_request")
-        assert "Expected a value of" in response.json()["message"] and "but found" in response.json()["message"]
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json()["errors"], True)
```
farshidz, line 120: Why remove the message validation?

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -62,6 +63,13 @@ def to_vespa_document(self, marqo_document: Dict[str, Any]) -> Dict[str, Any]:
 
             index_field = self._marqo_index.field_map[marqo_field]
 
+            if isinstance(marqo_value, (int, float)) and not index_field.type == FieldType.Bool:
```
farshidz: Shouldn't need `not index_field.type == FieldType.Bool` because boolean conversion to int is done further down in line 73

src/marqo/core/structured_vespa_index/structured_vespa_index.py
wanliAlex: done

src/marqo/core/structured_vespa_index/structured_vespa_index.py
wanliAlex: Yes, Vespa will error out for ` int("1" * 30)`. We have a test case for this.

tests/tensor_search/test_api.py
wanliAlex: Add back a different message validation

src/marqo/core/structured_vespa_index/structured_vespa_index.py
wanliAlex: Make sense. Done

src/marqo/core/structured_vespa_index/structured_vespa_index.py
wanliAlex: We need to exclude the `boolean` type here. I will use `if type(marqo_value) in (float, int)` to check instead of `isinstance`

src/marqo/core/structured_vespa_index/structured_vespa_index.py
wanliAlex: They will be round to 0. We think this is a reasonable behavior and we skip the check for that.

farshidz (approved):

PR #730, opened 2024-01-23
Create Marqo settings schema at startup

farshidz:
Bug fix/improvement

  * Marqo settings schema is created when the first index is created. This leads to a number of issues while Marqo settings schema does not exist, including 500 errors on operations such as delete_index, index_exists and cache refresh (prints a verbose warning, not an error).
  * Marqo welcome message is delayed and not printed out right away
  * SIGINT processing is slow and the container takes a while to terminate

  * Create Marqo settings schema in on-start script. If on-start script fails to create Marqo settings schema (e.g., because Vespa is not ready yet), Marqo will behave as before and encounter issues until an index is created, at which point the settings schema will also be created. In other words, if this fails at start-up, we will print a warning and will not retry until an index is created.
  * Flush welcome message right away
  * Terminate API process for SIGINT and SIGTERM

No

Yes

None

None

* **Other information**:

### DIFF
--- a/src/marqo/documentation.py
+++ b/src/marqo/documentation.py
@@ -1,0 +1,16 @@
+import marqo.version
+
+version = marqo.version.__version__
+base_url = 'https://docs.marqo.ai'
+
+
+def _build_url(path):
+    return f'{base_url}/{version}/{path}'
+
+
+def configuring_marqo():
+    return _build_url('Guides/Advanced-Usage/configuration/')
+
+
+def create_index():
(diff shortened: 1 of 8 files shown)


src/marqo/documentation.py
```diff
@@ -0,0 +1,16 @@
+import marqo.version
```
wanliAlex, line 1: Thanks for the work!

wanliAlex (approved):

PR #735, opened 2024-01-24
Increase hits and offset caps

farshidz:
Feature/improvement

Search limit and offset capped at 400 and 1000 respectively. Higher values lead to 500 errors 

Max limit 1000, max offset 10,000. Higher values cause a 400 error

No

Yes

No

No

* **Other information**:


src/marqo/core/index_management/index_management.py
```diff
@@ -34,23 +42,33 @@ class IndexManagement:
...
+    )
 
     def __init__(self, vespa_client: VespaClient):
         self.vespa_client = vespa_client
 
-    def create_settings_schema(self) -> bool:
+    def bootstrap_vespa(self) -> bool:
         """
-        Create the Marqo settings schema if it does not exist.
+        Add Marqo configuration to Vespa application package if an existing Marqo configuration is not detected.
```
pandu-k, line 59: Not a blocker for this PR, but there is an implied 3rd option - an error in configuring vespa. We can detail such a case here, including expected error

src/marqo/upgrades/v20_v21.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class v20v21(Upgrade):
+    """
+    Upgrade from Marqo v2.0 to v2.1.
+
+    The migration comprises two steps:
+    1. Create the default query profile
+    2. Add Marqo version to Marqo settings schema
+    """
```
pandu-k: does this need a validation step? checking if the current version _is_ v2.0?

pandu-k: Do we also need validation step that, after the pushes have been completed, ensures that the query profile exists and the version is as expected? (I realise this extra check is heading in the direction of over-engineering, and purely reducing the chances of inconsistencies rather than guaranteeing them, but it could still be useful for such a signifcant change). We don't necessarily need to include this in this release

pandu-k (commented):
Left a few comments. In terms of migration, these questions need to be answered:
1. Has Marqo 2.1 been tested for backwards compatibility with 2.0? We need a manual test plan/matrix (including whether the upgrade has been performed, or not). 
2. Has Marqo 2.0 been tested for forwards compatibility? This is important for a safe rollout. If we run into issues with 2.1 and we need to rollback we need to guarantee an OK state. Please add the details of this test to the above test plan.

src/marqo/upgrades/v20_v21.py
farshidz: `UpgradeRunner` does this validation and won't call this class if versions aren't right.

farshidz: Added

src/marqo/core/index_management/index_management.py
farshidz: Under normal circumstances there won't be an error. We don't normally document unexpected errors. I have a ticket to add custom errors to this in case of a network issue and at that point I'll document that custom error.

src/marqo/upgrades/v20_v21.py
```diff
@@ -0,0 +1,90 @@
...
+from marqo.api import exceptions as api_exceptions
+from marqo.core.index_management.index_management import IndexManagement, _MarqoConfig
+from marqo.upgrades.upgrade import Upgrade
+from marqo.vespa.models import VespaDocument
+from marqo.vespa.vespa_client import VespaClient
+
+logger = logging.get_logger(__name__)
+
+
+class v20v21(Upgrade):
```
Mehul-Porwal: Can the naming be better like `class v2_0_0_v2_1_0`? This naming seems like it could conflict once we get to double digit major and minor versions.

farshidz: How about now? Python class names should be PascalCase so changed to v2_v0_v2_v1 (V2V0V2V1 class name)

src/marqo/core/index_management/index_management.py
pandu-k: getting sidetracked from the this PR., but I'd say that network errors are almost "normal circumstances"

pandu-k (dismissed):

src/marqo/upgrades/upgrade.py
```diff
@@ -49,7 +53,36 @@ def _for_version(self, from_version) -> Optional[Upgrade]:
         to_version = f'{parsed_version.major}.{parsed_version.minor}'
 
         if from_version == "2.0" and to_version == '2.1':
-            from marqo.upgrades.v2_v0_v2_v1 import v2_0_v2_1
-            return v2_0_v2_1(self.vespa_client)
+            from marqo.upgrades.v2_v0_v2_v1 import V2V0V2V1
+            return V2V0V2V1(self.vespa_client, self.index_management)
```
pandu-k, line 57: nit: these version numbers are pretty unreadable. doesn't matter too much as this code shouldn't be used much

pandu-k (approved):

src/marqo/upgrades/upgrade.py
farshidz: Yes I couldn't come up with something better. But for future versions we can just delete these modules

PR #761, opened 2024-02-11
Vespa app package cleanup

farshidz:
Clean up Vespa application packages and update README


run_marqo.sh
```diff
@@ -73,7 +73,7 @@ elif [ -z "$VESPA_QUERY_URL" ] && [ -z "$VESPA_DOCUMENT_URL" ] && [ -z "$VESPA_C
     if echo "$RESPONSE" | grep -q '"error-code":"NOT_FOUND"'; then
       echo "Marqo did not find an existing vector store. Setting up vector store..."
       # Deploy a dummy application package
-      vespa deploy /app/scripts/vespa_dummy_app --wait 300 >/dev/null 2>&1
```
vicilliar, line 76: Shouldn't we also delete the `vespa_dummy_app` directly along with the zip file?

vicilliar (changes requested):

run_marqo.sh
farshidz: Done

vicilliar (approved):

PR #773, opened 2024-02-28
Document Partial Update

wanliAlex:
feature

We can only update a document with `use_existing_tensor=True`. However, this API needs to fetch and upload the documents. The throughput of this is slow.

We add an update_document API that can update the non_tensor_fields inside the data engine. This can update the documents with high throughput.

no

yes

working on

working on

* **Other information**:
no


src/marqo/core/document/document.py
```diff
@@ -13,12 +19,84 @@ def __init__(self, vespa_client: VespaClient, index_management: IndexManagement)
...
+        marqo_index = self.index_management.get_index(index_name)
+        if marqo_index.type == IndexType.Structured:
+            return self.structured_partial_update_documents(update_documents_params.documents, marqo_index)
+        elif marqo_index.type == IndexType.Unstructured:
+            raise UnsupportedFeatureError("'Partial document update' is not supported for unstructured indexes. "
+                                          "Please use 'add_documents' with 'use_existing_tensor=True' instead.")
+        else:
+            ValueError(f"No known implementation for index type {marqo_index.type}")
+
+    def structured_partial_update_documents(self, documents, marqo_index):
```
farshidz: Ideally we don't want to have an index type distinction here. The Vespa index abstraction should handle this similar to how search is implemented

src/marqo/core/models/update_documents.py
```diff
@@ -0,0 +1,28 @@
+from typing import Dict, Sequence, Any
+
+from marqo.base_model import ImmutableStrictBaseModel
+from pydantic import BaseModel, validator
+from marqo.api.exceptions import BadRequestError
+from marqo.tensor_search.utils import get_best_available_device, read_env_vars_and_defaults_ints
+from marqo.tensor_search.enums import EnvVars
+
+
+class UpdateDocumentsBodyParams(ImmutableStrictBaseModel):
```
farshidz: Isn't this an API model?

src/marqo/tensor_search/api.py
```diff
@@ -207,6 +208,17 @@ def add_or_replace_documents(
         )
 
 
+@app.post("/indexes/{index_name}/documents/update")
+@throttle(RequestType.UPDATE)
```
farshidz: Partial update might be a better type in case we want to throttle standard update separately later

src/marqo/core/models/update_documents.py
wanliAlex: I have decoupled the API model and Core model

src/marqo/tensor_search/api.py
wanliAlex: Make sense

wanliAlex: revised

src/marqo/core/document/document.py
wanliAlex: Done.

src/marqo/api/models/update_documents.py
```diff
@@ -0,0 +1,7 @@
+from typing import Dict, Sequence, Any
+
+from marqo.base_model import ImmutableStrictBaseModel
+
+
+class UpdateDocumentsBodyParams(ImmutableStrictBaseModel):
+    documents: Sequence[Dict[str, Any]]
```
farshidz: I assume this dictionary has some expected schema. e.g., must have `_id`. If so, it should have a model e.g., `MarqoPartialDocument`. We also need a `MarqoDocument` model to replace dicts elsewhere but that's not for this PR

src/marqo/core/document/document.py
```diff
@@ -13,12 +19,85 @@ def __init__(self, vespa_client: VespaClient, index_management: IndexManagement)
...
         marqo_index = self.index_management.get_index(index_name)
-        res: DeleteAllDocumentsResponse = self.vespa_client.delete_all_docs(marqo_index.schema_name)
+        return self.delete_all_docs(marqo_index)
 
+    def delete_all_docs(self, marqo_index):
+        """Deletes all documents in the given index by marqo_index object"""
+        res: DeleteAllDocumentsResponse = self.vespa_client.delete_all_docs(marqo_index.schema_name)
         return res.document_count
+
+    def partial_update_documents_by_index_name(self, index_name,
```
farshidz, line 42: Public methods must have documentation, including args and exceptions

src/marqo/core/document/document.py
```diff
@@ -13,12 +19,85 @@ def __init__(self, vespa_client: VespaClient, index_management: IndexManagement)
...
+                                               update_documents_params: UpdateDocumentsParams) -> Dict:
+        if not self.index_management.index_exists(index_name):
```
farshidz: I think this is redundant because `get_index` below raises this error if index isn't found

src/marqo/core/document/document.py
```diff
@@ -13,12 +19,85 @@ def __init__(self, vespa_client: VespaClient, index_management: IndexManagement)
...
         self.index_management = index_management
 
-    def delete_all_docs(self, index_name: str) -> int:
-        """Deletes all documents in the given index"""
+    def delete_all_docs_by_index_name(self, index_name: str) -> int:
+        """Deletes all documents in the given index by index name"""
         if not self.index_management.index_exists(index_name):
             raise IndexNotFoundError(f"Index {index_name} does not exist")
 
         marqo_index = self.index_management.get_index(index_name)
```
farshidz, line 31: Args must be documented

src/marqo/core/document/document.py
```diff
@@ -13,12 +19,85 @@ def __init__(self, vespa_client: VespaClient, index_management: IndexManagement)
         self.vespa_client = vespa_client
         self.index_management = index_management
 
-    def delete_all_docs(self, index_name: str) -> int:
-        """Deletes all documents in the given index"""
+    def delete_all_docs_by_index_name(self, index_name: str) -> int:
+        """Deletes all documents in the given index by index name"""
         if not self.index_management.index_exists(index_name):
```
farshidz: Redundant as `index_management.get_index` raises this

src/marqo/core/document/document.py
```diff
@@ -13,12 +19,85 @@ def __init__(self, vespa_client: VespaClient, index_management: IndexManagement)
         self.vespa_client = vespa_client
         self.index_management = index_management
 
-    def delete_all_docs(self, index_name: str) -> int:
-        """Deletes all documents in the given index"""
+    def delete_all_docs_by_index_name(self, index_name: str) -> int:
+        """Deletes all documents in the given index by index name"""
```
farshidz: Imperative tone is customary for Python docstrings i.e. "Delete all documents" as opposed to "Deletes". Applies to all docstrings in this PR

src/marqo/core/document/document.py
```diff
@@ -13,12 +19,85 @@ def __init__(self, vespa_client: VespaClient, index_management: IndexManagement)
...
+
+        marqo_index = self.index_management.get_index(index_name)
+        return self.partial_update_documents(update_documents_params, marqo_index)
+
+    def partial_update_documents(self, update_documents_params: UpdateDocumentsParams, marqo_index):
+        """Partially updates documents in the given index by marqo_index object.
+
+        This is not supported for unstructured indexes."""
+        if marqo_index.type == IndexType.Unstructured:
+            raise UnsupportedFeatureError("'Partial document update' is not supported for unstructured indexes. "
```
farshidz: Single quotes around Parital document update seem unnecessary. Also for consistency and brevity, we usually skip the final period in exception messages

src/marqo/core/document/document.py
```diff
@@ -13,12 +19,85 @@ def __init__(self, vespa_client: VespaClient, index_management: IndexManagement)
...
+            raise IndexNotFoundError(f"Index {index_name} does not exist")
+
+        marqo_index = self.index_management.get_index(index_name)
+        return self.partial_update_documents(update_documents_params, marqo_index)
+
+    def partial_update_documents(self, update_documents_params: UpdateDocumentsParams, marqo_index):
+        """Partially updates documents in the given index by marqo_index object.
+
+        This is not supported for unstructured indexes."""
+        if marqo_index.type == IndexType.Unstructured:
```
farshidz, line 83: It's usually safer to cover all possibilities with an else

src/marqo/core/document/document.py
```diff
@@ -13,12 +19,85 @@ def __init__(self, vespa_client: VespaClient, index_management: IndexManagement)
...
+
+    def partial_update_documents_by_index_name(self, index_name,
+                                               update_documents_params: UpdateDocumentsParams) -> Dict:
+        if not self.index_management.index_exists(index_name):
+            raise IndexNotFoundError(f"Index {index_name} does not exist")
+
+        marqo_index = self.index_management.get_index(index_name)
+        return self.partial_update_documents(update_documents_params, marqo_index)
+
+    def partial_update_documents(self, update_documents_params: UpdateDocumentsParams, marqo_index):
```
farshidz: Return type annotation missing

src/marqo/core/document/document.py
```diff
@@ -13,12 +19,85 @@ def __init__(self, vespa_client: VespaClient, index_management: IndexManagement)
...
+                vespa_document = VespaDocument(**vespa_index.to_vespa_partial_document(doc))
+                vespa_documents.append(vespa_document)
+            except ParsingError as e:
+                unsuccessful_docs.append(
+                    (index, {'_id': doc.get('_id', ''), 'error': e.message,
+                             "status": int(api_exceptions.InvalidArgError.status_code),
+                             'code': api_exceptions.InvalidArgError.code})
+                )
+
+        vespa_res: UpdateBatchResponse = self.vespa_client.update_batch(vespa_documents, marqo_index.schema_name)
```
farshidz: Is this client method specific to partial updates? If so the name should indicate this

src/marqo/core/document/document.py
```diff
@@ -13,12 +19,85 @@ def __init__(self, vespa_client: VespaClient, index_management: IndexManagement)
...
+
+        res = self._translate_update_document_response(vespa_res, unsuccessful_docs,
+                                                       marqo_index.name, start_time)
+
+        return res
+
+    def _translate_update_document_response(self, responses: UpdateBatchResponse, unsuccessful_docs: List,
```
farshidz: Respond type should be a model

src/marqo/core/models/marqo_index.py
```diff
@@ -397,6 +397,13 @@ def field_map_by_type(self) -> Dict[FieldType, List[Field]]:
                                            for field_type in FieldType}
                                   )
 
+    @property
+    def dependent_fields_names(self) -> Set[str]:
+        return self._cache_or_get('dependent_fields_names',
```
farshidz, line 403: To be clear, this is all dependent fields from all multimodal fields in one set. Is this what you want?

farshidz: Also a docstring would help

src/marqo/core/models/update_documents.py
```diff
@@ -0,0 +1,29 @@
+from typing import Dict, Sequence, Any
+
+from pydantic import validator
+
+from marqo.api.exceptions import BadRequestError
```
farshidz: Core shouldn't be raising API exceptions

src/marqo/core/models/update_documents.py
```diff
@@ -0,0 +1,29 @@
...
+    documents: Sequence[Dict[str, Any]]
+
+    @validator('documents')
+    def validate_docs(cls, documents):
+        doc_count = len(documents)
+
+        max_doc = read_env_vars_and_defaults_ints(EnvVars.MARQO_MAX_DOCUMENTS_BATCH_SIZE)
+
+        if doc_count == 0:
+            raise BadRequestError(message="Received empty add documents request")
```
farshidz: update documents request?

src/marqo/core/models/update_documents.py
```diff
@@ -0,0 +1,29 @@
...
+
+from pydantic import validator
+
+from marqo.api.exceptions import BadRequestError
+from marqo.base_model import ImmutableStrictBaseModel
+from marqo.tensor_search.enums import EnvVars
+from marqo.tensor_search.utils import read_env_vars_and_defaults_ints
+
+
+class UpdateDocumentsParams(ImmutableStrictBaseModel):
```
farshidz: I wonder if something like `UpdateDocumentsRequest` is more descriptive? Also module name should usually match class name

src/marqo/core/models/update_documents.py
```diff
@@ -0,0 +1,29 @@
...
+    documents: Sequence[Dict[str, Any]]
+
+    @validator('documents')
+    def validate_docs(cls, documents):
+        doc_count = len(documents)
+
+        max_doc = read_env_vars_and_defaults_ints(EnvVars.MARQO_MAX_DOCUMENTS_BATCH_SIZE)
```
farshidz: This appears to be an API level condition. One option is to move this model to API, and just get a `Sequence[Dict[str, Any]]` in Core without using a model

farshidz (changes requested):

src/marqo/core/document/document.py
wanliAlex: done

src/marqo/vespa/vespa_client.py
```diff
@@ -425,6 +425,20 @@ def delete_batch(self,
 
         return batch_response
 
+    def update_batch(self, batch: List[VespaDocument],
```
farshidz: This needs to be tested

src/marqo/vespa/vespa_client.py
```diff
@@ -425,6 +425,20 @@ def delete_batch(self,
...
+                     schema: str,
+                     concurrency: int = 10,
```
farshidz: Note this concurrency level could restrict throughput. Ideally this should be configurable through an env var

src/marqo/core/models/update_documents.py
wanliAlex: done

src/marqo/vespa/vespa_client.py
```diff
@@ -425,6 +425,20 @@ def delete_batch(self,
...
+                     timeout: int = 60) -> UpdateBatchResponse:
+
+        if not batch:
```
farshidz, line 469: Is this just partial updates? Or will this work for normal updates to? If partial, name should reflect that

src/marqo/vespa/vespa_client.py
```diff
@@ -563,6 +577,54 @@ async def _feed_batch_async(self, batch: List[VespaDocument],
...
+
+    async def _update_document_async(self, semaphore: asyncio.Semaphore, async_client: httpx.AsyncClient,
+                                     document: VespaDocument, schema: str,
+                                     timeout: int) -> UpdateBatchDocumentResponse:
+        doc_id = document.id
+        data = {'fields': document.fields}
+
+        async with semaphore:
+            end_point = f'{self.document_url}/document/v1/{schema}/{schema}/docid/{doc_id}?create=false'
+            data["condition"] = f'{schema}.marqo__id==\"{doc_id}\"'
```
farshidz: pydocs of the public method should indicate that will fail if doc ID doesn't exist (and error code)

src/marqo/vespa/vespa_client.py
```diff
@@ -563,6 +577,54 @@ async def _feed_batch_async(self, batch: List[VespaDocument],
...
+            except httpx.HTTPError as e:
+                raise VespaError(e) from e
+
+        try:
+            # This will cover 200 and document-specific errors. Other unexpected errors will be raised.
+            return UpdateBatchDocumentResponse(**resp.json(), status=resp.status_code)
+        except JSONDecodeError:
+            if resp.status_code == 200:
+                # A 200 response shouldn't reach here
+                raise VespaError(f'Unexpected response: {resp.text}')
```
farshidz, line 664: `from e` so we don't lose the stack trace

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -52,6 +52,78 @@ class StructuredVespaIndex(VespaIndex):
     def __init__(self, marqo_index: StructuredMarqoIndex):
         self._marqo_index = marqo_index
 
+    def to_vespa_partial_document(self, marqo_document: Dict[str,Any]) -> Dict[str, Any]:
+        vespa_id: Optional[str] = None
+        vespa_fields: Dict[str, Any] = dict()
+        score_modifiers: Dict[str, float] = dict()
+
+        if constants.MARQO_DOC_ID not in marqo_document:
+            raise MarqoDocumentParsingError(f"Marqo encountered an error when parsing the partial update document "
```
farshidz: I think the message is too verbose. Just " '{constants.MARQO_DOC_ID}' is a required field "
                                            f"but it does not exist." should suffice. Also printing our the whole dict seems like a bad idea

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -52,6 +52,78 @@ class StructuredVespaIndex(VespaIndex):
...
+            if marqo_field == constants.MARQO_DOC_ID:
+                continue
+            if marqo_field == constants.MARQO_DOC_TENSORS:
+                raise MarqoDocumentParsingError(f"Marqo encountered an error when parsing the partial update document "
+                                                f"'{marqo_document}'. You can not modify '{marqo_field}' "
+                                                f"field. ")
+
+            tensor_fields_names = [tensor_field.name for tensor_field in self._marqo_index.tensor_fields]
+            if marqo_field in tensor_fields_names:
+                raise MarqoDocumentParsingError(f"Marqo encountered an error when parsing the partial update document "
```
farshidz: Message too verbos. Applies to all exceptions below

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -52,6 +52,78 @@ class StructuredVespaIndex(VespaIndex):
...
+                                                f"'{marqo_document}'. You CAN NOT modify '{marqo_field}' "
+                                                f"field as this is a tensor field.")
```
farshidz: You cannot (single word). No need for all caps

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -52,6 +52,78 @@ class StructuredVespaIndex(VespaIndex):
...
+            if isinstance(marqo_value, list) and len(marqo_value) > 0 and type(marqo_value[0]) in (float, int):
+                for v in marqo_value:
+                    self._verify_numerical_field_value(v, index_field)
+
+            if index_field.type == FieldType.Bool:
+                # Booleans are stored as bytes in Vespa
+                marqo_value = int(marqo_value)
+
+            if index_field.lexical_field_name:
+                vespa_fields[index_field.lexical_field_name] = dict()
```
farshidz: Optioanlly you could inline these like ` = {'assign': ...}`

farshidz (changes requested):

src/marqo/api/models/update_documents.py
wanliAlex: We decided not to use pydantic model for partial documents after a discussion because we don't want to reject a whole batch of documents if one document is missing a `_id` field

src/marqo/core/models/update_documents.py
wanliAlex: Done. Move this validation to API level.

src/marqo/core/structured_vespa_index/structured_vespa_index.py
wanliAlex: yes that's better

src/marqo/vespa/vespa_client.py
wanliAlex: done

src/marqo/vespa/vespa_client.py
wanliAlex: Good catch!

src/marqo/core/models/marqo_index.py
wanliAlex: done

src/marqo/vespa/vespa_client.py
wanliAlex: done

farshidz (approved):

PR #780, opened 2024-03-06
Patch the _id not return issue

wanliAlex:
- **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
bug fix
- **What is the current behavior?** (You can also link to an open issue here)
we don't return `_id` in the partial update response due to a private attribute pydantic model issue
- **What is the new behavior (if this is a feature change)?**
we can return `_id` now
- **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
no
- **Have unit tests been run against this PR?** (Has there also been any additional testing?)
yes
- **Related Python client changes** (link commit/PR here)
no
- **Related documentation changes** (link commit/PR here)
no
- **Other information**:
no
- **Please check if the PR fulfills these requirements**
- [x]  The commit message follows our guidelines
- [x]  Tests for the changes have been added (for bug fixes/features)
- [x]  Docs have been added / updated (for bug fixes / features)


src/marqo/core/document/document.py
```diff
@@ -57,7 +57,7 @@ def partial_update_documents_by_index_name(self, index_name,
         marqo_index = self.index_management.get_index(index_name)
 
         res: MarqoUpdateDocumentsResponse = self.partial_update_documents(partial_documents, marqo_index)
-        return res.dict(exclude_none=True)
+        return res.dict(exclude_none=True, by_alias=True)
```
farshidz: Wouldn't it be nicer to return the model rather than a dict?

wanliAlex: Done. We now return the model directly and format the response in the API function.

farshidz (approved):

PR #800, opened 2024-04-10
Fix telemetry cleanup bug

farshidz:
* Fix bug where telemetry does not clean up objects when telemetry=false
* This PR also adds a memory profiler endpoint `/memory` disabled by default and enable through an environment variable

When telemetry=false, Telemetry leaks objects by adding a class variable and not cleaning up when the request is complete

Clean up the dict when done

No

Yes

No

No

* **Other information**:


src/marqo/core/monitoring/memory_profiler.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def get_memory_profile() -> MemoryProfile:
+    tracemalloc.start()
+
+    snapshot = tracemalloc.take_snapshot()
+    stats = snapshot.statistics('lineno')
+
+    # GC
+    gc.collect()  # Force garbage collection
```
pandu-k: does it makes sense for the the get_profile function to also do a GC? doesn't that mean the profiled memory is accurate? 
Also, I am wary of a read call (like get_mem_profile) with a side effect like this. 
Perhaps a dedicated gc() function is better

tests/tensor_search/test_api.py
```diff
@@ -34,6 +35,30 @@ def test_add_or_replace_documents_tensor_fields(self):
             self.assertEqual(response.status_code, 200)
             mock_add_documents.assert_called_once()
```
pandu-k, line 37: nice. really appreciate adding unit tests even for these debug endpoints

pandu-k (dismissed):

pandu-k (approved):

PR #802, opened 2024-04-12
Inference Cache

wanliAlex:
Feature

No inference cache

We now support inference cache in Marqo for frequent search requests

no

running

no

no

* **Other information**:
no


src/marqo/s2_inference/inference_cache/abstract_cache.py
```diff
@@ -0,0 +1,83 @@
+from abc import ABC, abstractmethod
```
vicilliar: If I recall correctly, we want to move away from the `s2_inference` module and start using `inference`. Is this correct @farshidz ? In this case, maybe we should move this into that directory.

src/marqo/s2_inference/inference_cache/abstract_cache.py
```diff
@@ -0,0 +1,83 @@
...
+from typing import List
+
+
+class MarqoAbstractCache(ABC):
```
vicilliar: Do we need an abstract cache? If the key is model + content and the value is the embedding, what other caches could we have aside from InferenceCache?

src/marqo/s2_inference/inference_cache/inference_cache.py
```diff
@@ -0,0 +1,77 @@
...
+        with self.lock.gen_wlock():
+            self.cache[key] = value
+
+    def __getitem__(self, key: str) -> List[float]:
+        with self.lock.gen_rlock():
+            return self.cache[key]
+
+    def __setitem__(self, key: str, value: List[float]) -> None:
+        with self.lock.gen_wlock():
+            self.cache[key] = value
```
vicilliar: This currently is just the same logic as `set()`

src/marqo/s2_inference/inference_cache/inference_cache.py
```diff
@@ -0,0 +1,77 @@
...
+
+from readerwriterlock import rwlock
+
+from marqo.api.exceptions import EnvVarError
+from marqo.s2_inference.inference_cache.abstract_cache import MarqoAbstractCache
+from marqo.s2_inference.inference_cache.marqo_lfu_cache import MarqoLFUCache
+from marqo.s2_inference.inference_cache.marqo_lru_cache import MarqoLRUCache
+
+
+class InferenceCache:
```
vicilliar: Should we name this `MarqoInferenceCache` for uniformity?

src/marqo/s2_inference/inference_cache/marqo_lfu_cache.py
```diff
@@ -0,0 +1,35 @@
...
+        return self.cache.get(key, default)
+
+    def set(self, key: str, value: List[float]):
+        self.cache[key] = value
+
+    def __contains__(self, key: str) -> bool:
+        return key in self.cache
+
+    def __setitem__(self, key: str, value: List[float]):
+        self.cache[key] = value
```
vicilliar: Where is the logic to pop items until the cache has free space?

src/marqo/s2_inference/inference_cache/marqo_lru_cache.py
```diff
@@ -0,0 +1,34 @@
...
+
+from marqo.s2_inference.inference_cache.abstract_cache import MarqoAbstractCache
+
+
+class MarqoLRUCache(MarqoAbstractCache):
+    """Least Recently Used (LRU) cache implementation.
+
+    This class is currently implemented using cachetools.LRUCache, but it can be replaced with any other LRU cache."""
+    def __init__(self, maxsize: int):
+        self.cache = LRUCache(maxsize=maxsize)
```
vicilliar: Is this line the only difference between `MarqoLRUCache` and `MarqoLFUCache`? Meaning that we don't do much separate logic in Marqo? If so, it might be excessive to make them separate modules like this. Unless we project the code to diverge even more.

src/marqo/s2_inference/s2_inference.py
```diff
@@ -494,6 +563,22 @@ def _convert_tensor_to_numpy(output:Union[FloatTensor, Tensor]) -> ndarray:
...
+    A function that converts cached embeddings List[float] to the expected output List[List[float]]
+
+    Args:
+        cached_embeddings (List[float]): The cached embeddings to convert
+    Return:
+        List[List[float]]: The converted embeddings in the expected output format
+    """
+    if not isinstance(cached_embeddings, list):
+        raise TypeError(f"expected a list of floats but received {type(cached_embeddings)}")
+    if not isinstance(cached_embeddings[0], float):
```
vicilliar, line 579: What if one of the other elements isn't a float?

src/marqo/s2_inference/s2_inference.py
```diff
@@ -564,11 +649,11 @@ def _get_model_loader(model_name: str, model_properties: dict) -> Any:
 
 
 def get_available_models():
-    return available_models
+    return _available_models
```
vicilliar: These 2 functions were also declared above (line 151)

vicilliar (changes requested):

src/marqo/s2_inference/inference_cache/inference_cache.py
```diff
@@ -0,0 +1,77 @@
...
+
+    The key is a string consisting of model_cache_key and content to identify the cache.
+    The value is a list of floats representing the embeddings.
+    """
+    _CACHE_TYPE_MAPPINGS = {
+        "LFU": MarqoLFUCache,
+        "LRU": MarqoLRUCache
+    }
+
+    def __init__(self, cache_size: int, cache_type: str = "LRU"):
```
farshidz: Rather than cache type it may make more sense to get an instance of `MarqoAbstractCache` here. Not sure though. An enum at least though, as noted above

farshidz (changes requested):
Both cache classes need to be unit tested

src/marqo/s2_inference/inference_cache/abstract_cache.py
```diff
@@ -0,0 +1,83 @@
...
+        """Set the value for key in the cache.
+
+        Args:
+            key: __description__
+            value: __description__
+        """
+        pass
+
+    @abstractmethod
+    def popitem(self) -> None:
```
farshidz: `pop_item` would be more pythonic

farshidz: Maybe just `pop`?

src/marqo/s2_inference/inference_cache/inference_cache.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class InferenceCache:
+    """Inference cache class that wraps MarqoAbstractCache and LRUCache with a read-write lock.
+
+    The key is a string consisting of model_cache_key and content to identify the cache.
+    The value is a list of floats representing the embeddings.
+    """
+    _CACHE_TYPE_MAPPINGS = {
+        "LFU": MarqoLFUCache,
```
farshidz: Why not use an enum?

src/marqo/s2_inference/inference_cache/inference_cache.py
```diff
@@ -0,0 +1,77 @@
...
+        "LRU": MarqoLRUCache
+    }
+
+    def __init__(self, cache_size: int, cache_type: str = "LRU"):
+        self.cache_size = cache_size
+        self.cache_type = cache_type
+        self.cache = self._build_cache()
+
+        self.lock = rwlock.RWLockFair()
```
farshidz: I think the locking should happen inside your cache class so the cache is thread safe and consumers don't have to worry about this

farshidz: You could even have locking functionality in the base abstract class for implementations to use

src/marqo/s2_inference/inference_cache/marqo_lru_cache.py
farshidz: I think it's possible we'll want to use a different library for one of these types later on. The separation makes sense IMO

src/marqo/s2_inference/s2_inference.py
```diff
@@ -15,39 +16,48 @@
...
+_available_models = dict()
 # A lock to protect the model loading process
 lock = threading.Lock()
 MODEL_PROPERTIES = load_model_properties()
+_marqo_inference_cache = InferenceCache(cache_size=read_env_vars_and_defaults_ints(EnvVars.MARQO_INFERENCE_CACHE_SIZE),
+                                        cache_type=read_env_vars_and_defaults(EnvVars.MARQO_INFERENCE_CACHE_TYPE))
 
 
-def vectorise(model_name: str, content: Union[str, List[str]], model_properties: dict = None,
+def vectorise(model_name: str, content: Union[str, List[Union[str,Image]], List[Image]],
```
farshidz: Can content be a list of str and image mixed? I.e. a list having two types in it

src/marqo/s2_inference/s2_inference.py
```diff
@@ -67,14 +77,64 @@ def vectorise(model_name: str, content: Union[str, List[str]], model_properties:
         model_auth=model_auth
     )
 
+    if _marqo_inference_cache.is_enabled() and enable_cache:
+        if isinstance(content, str):
+            inference_cache_key = f"{model_cache_key}||{content}"
```
farshidz: I think making sure that cached values for different models don't clash is InferenceCache class's responsibility. You should move this logic there. InferenceCache will then require the string, but also model etc to access the cache.

src/marqo/s2_inference/s2_inference.py
```diff
@@ -67,14 +77,64 @@ def vectorise(model_name: str, content: Union[str, List[str]], model_properties:
...
+                    if vectorised is None:
+                        contents_to_vectorise.append(content_item)
+                    else:
+                        cached_output.append((loc, vectorised))
+
+            if contents_to_vectorise:
+                vectorised_outputs: List[List[float]] = _encode_without_cache(model_cache_key, contents_to_vectorise,
+                                                                              normalize_embeddings, **kwargs)
+                for contents_to_vectorise_loc, vectorised_output in enumerate(vectorised_outputs):
+                    _marqo_inference_cache[f"{model_cache_key}||{contents_to_vectorise[contents_to_vectorise_loc]}"] = (
```
farshidz: Is this caching images here?

src/marqo/s2_inference/s2_inference.py
```diff
@@ -67,14 +77,64 @@ def vectorise(model_name: str, content: Union[str, List[str]], model_properties:
...
+                        vectorised_output)
+
+                for loc, cached_vector in cached_output:
+                    vectorised_outputs.insert(loc, cached_vector)
```
farshidz, line 115: Are we breaking the order here when some items in the list were cached and some were not?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -349,7 +349,8 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
                                     model_properties=marqo_index.model.get_properties(), content=content_chunks,
                                     device=add_docs_params.device, normalize_embeddings=normalize_embeddings,
                                     infer=marqo_index.treat_urls_and_pointers_as_images,
-                                    model_auth=add_docs_params.model_auth
+                                    model_auth=add_docs_params.model_auth,
+                                    enable_cache=False
```
farshidz: Why is this hard-coded false?

farshidz (changes requested):

src/marqo/s2_inference/s2_inference.py
wanliAlex: Actually no, I should use `List[str], List[Image]`

src/marqo/s2_inference/inference_cache/inference_cache.py
wanliAlex: Agree

src/marqo/s2_inference/inference_cache/inference_cache.py
wanliAlex: Done

src/marqo/s2_inference/inference_cache/inference_cache.py
wanliAlex: I will use enum but keep the existing `_build_cache` function. Because if I don't build the cache here, I need to build it in the s2_inference module, which I think it's not a good idea.

src/marqo/s2_inference/inference_cache/inference_cache.py
wanliAlex: Done

src/marqo/s2_inference/s2_inference.py
wanliAlex: done

src/marqo/s2_inference/s2_inference.py
wanliAlex: Yes, Image represented by a `url` string

src/marqo/s2_inference/s2_inference.py
wanliAlex: yes, remove the duplicated one

src/marqo/s2_inference/inference_cache/abstract_cache.py
wanliAlex: As discussed, we add a layer of abstraction here in case we want to change the cache implementation in the future.

src/marqo/s2_inference/inference_cache/inference_cache.py
wanliAlex: I added this method as it is a convention to have a set method in a cache, although I am not using it.

src/marqo/s2_inference/inference_cache/marqo_lfu_cache.py
wanliAlex: The logic is handled by the cachetools.LFUCache itself.

src/marqo/s2_inference/s2_inference.py
wanliAlex: Good question. There is a bit of trade-off here. It's safer to check the type of all the elements, but it is slower. Therefore I decided to only check the type of the first element.

src/marqo/s2_inference/inference_cache/abstract_cache.py
wanliAlex: Good catch. Moved all the inference cache modules to the `marqo.inference` folder

src/marqo/tensor_search/tensor_search.py
wanliAlex: I will remove these values in add_documments calls. They will use the default `False` now

src/marqo/s2_inference/s2_inference.py
wanliAlex: No, the orders are preserved. I will add extra tests to confirm this

farshidz (approved):

PR #803, opened 2024-04-12
Embed API endpoint

vicilliar:
Feature

No embed endpoint

Embed endpoint created to directly retrieve vectorised content (API format similar to search).
`content` can be a single item (str or dict of weights) or a list of items.

No

In progress

In progress

In progress

* **Other information**:


src/marqo/tensor_search/models/api_models.py
```diff
@@ -85,6 +85,42 @@ class BulkSearchQuery(BaseMarqoModel):
     queries: List[BulkSearchQueryEntity]
 
 
+class EmbeddingRequestParams(BaseMarqoModel):
```
farshidz: We're moving these to /api/models. See `RollbackRequest` for example. I suggest following the same convention (e.g., `EmbeddingRequest`)

tests/tensor_search/integ_tests/test_embed.py
```diff
@@ -0,0 +1,405 @@
...
+import uuid
+
+
+class TestEmbed(MarqoTestCase):
+    @classmethod
+    def setUpClass(cls) -> None:
+        super().setUpClass()
+
+        # Mock query result
+        cls.mock_query_result = QueryResult(
```
farshidz, line 36: Can you separate unit tests and integ tests? Unit tests don't need a Marqo index and can be much more thorough (because they're faster)

farshidz (changes requested):

src/marqo/tensor_search/models/api_models.py
vicilliar: done

tests/tensor_search/integ_tests/test_embed.py
vicilliar: Hi Farshid, at least for the `embed` feature, there's not much point in separating integ and unit tests, as there aren't helper functions to test (aside from validators). The integ tests test all relevant functionality.

src/marqo/api/models/embed_request.py
```diff
@@ -0,0 +1,53 @@
...
+from marqo.tensor_search.enums import SearchMethod
+from marqo.tensor_search.models.private_models import ModelAuth
+from marqo.tensor_search.models.score_modifiers_object import ScoreModifier
+from marqo.tensor_search.models.search import SearchContext, SearchContextTensor
+from marqo.tensor_search.models.api_models import BaseMarqoModel
+from marqo.api.exceptions import InvalidArgError
+from marqo.core.models.marqo_index import MarqoIndex
+
+
+class EmbeddingRequest(BaseMarqoModel):
```
farshidz: File name and class name should ideally match (unless the file has multiple classes)

src/marqo/api/models/embed_request.py
```diff
@@ -0,0 +1,53 @@
...
+    # content can be a single query or list of queries. Queries can be a string or a dictionary.
+    content: Union[Union[str, Dict[str, float]], List[Union[str, Dict[str, float]]]]
```
farshidz: Can a list have both strings and dict? I don't think that would be valid. I think what you mean is `Union[str, Dict[str, float], List[str], List[Dict[str, float]]]`

src/marqo/api/models/embed_request.py
```diff
@@ -0,0 +1,53 @@
...
+        if (isinstance(value, list) or isinstance(value, dict)) and len(value) == 0:
+            raise ValueError("Embed content list should not be empty")
+
+        # Convert all types of content into a list
+        if isinstance(value, str) or isinstance(value, dict):
+            list_to_validate = [value]
+        elif isinstance(value, List):
+            list_to_validate = value
+        else:
+            raise ValueError("Embed content should be a string, dictionary, or a list of strings or dictionaries")
```
farshidz: a string, a dictionary, or a ... (missing a before dictionary)

src/marqo/api/models/embed_request.py
```diff
@@ -0,0 +1,53 @@
...
+
+        for item in list_to_validate:
+            if isinstance(item, str):
+                pass
```
farshidz: `continue` is what you would normally use in a for loop. Though I think pass has the same effect

src/marqo/api/models/embed_request.py
```diff
@@ -0,0 +1,53 @@
...
+            elif isinstance(item, dict):
+                if len(item) == 0:
+                    raise ValueError("Dictionary content should not be empty")
+                if not all(isinstance(k, str) for k in item.keys()):
```
farshidz: You could probably do it faster in one pass over key value pairs. Though the difference would likely be negligible here considering the small size

src/marqo/api/models/embed_request.py
```diff
@@ -0,0 +1,53 @@
...
+                    raise ValueError("Keys in dictionary content should all be strings")
+                if not all(isinstance(v, float) for v in item.values()):
+                    raise ValueError("Values in dictionary content should all be floats")
+            else:
+                raise ValueError("Embed content should be a string, dictionary, or a list of strings or dictionaries")
```
farshidz: a dictionary

src/marqo/core/embed/embed.py
```diff
@@ -0,0 +1,123 @@
+from timeit import default_timer as timer
+from typing import List, Optional, Union, Iterable, Sequence, Dict, Any, Tuple
+import numpy as np
+
+import marqo.core.unstructured_vespa_index.common as unstructured_common
+from marqo import marqo_docs
+from marqo import exceptions as base_exceptions
+from marqo.api import exceptions as api_exceptions
+from marqo.core import exceptions as core_exceptions
+# We depend on _httprequests.py for now, but this may be replaced in the future, as
```
farshidz: ?

src/marqo/core/embed/embed.py
```diff
@@ -0,0 +1,123 @@
...
+from marqo.tensor_search.tensor_search_logging import get_logger
+from marqo.vespa.exceptions import VespaStatusError
+from marqo.vespa.models import VespaDocument, FeedBatchResponse, QueryResult
+
+logger = get_logger(__name__)
+
+
+class Embed:
+    def __init__(self, config):
+        self.config = config
```
farshidz: The config object is problematic. It has extra things you don't really depend on and tends to cause dependency cycles when core objects depend on each other. Replace with specific dependencies

src/marqo/core/embed/embed.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def embed_content(
+                    self, content: Union[Union[str, Dict[str, float]], List[Union[str, Dict[str, float]]]],
```
farshidz: See comment on the other type annotation above

src/marqo/core/embed/embed.py
```diff
@@ -0,0 +1,123 @@
...
+                List of embeddings corresponding to the content. If content is a list, the return list will be in the same order.
+                If content is a string, the return list will only have 1 item.
+        """
+
+        # Content validation is done in API model layer
+        t0 = timer()
+
+        # Determine device
+        if device is None:
+            selected_device = utils.read_env_vars_and_defaults("MARQO_BEST_AVAILABLE_DEVICE")
```
farshidz: Can you pass this in the constructor? We don't want to access env vars outside of the API. It reduces your code's reusability and it's also less explicit (since you're reading configuration outside of class initialization and method signature)

src/marqo/core/embed/embed.py
```diff
@@ -0,0 +1,123 @@
...
+            if selected_device is None:
+                raise base_exceptions.InternalError("Best available device was not properly determined on Marqo startup.")
+            logger.debug(f"No device given for search. Defaulting to best available device: {selected_device}")
+        else:
+            selected_device = device
+
+        # Generate input for the vectorise pipeline (Preprocessing)
+        RequestMetricsStore.for_request().start("embed.query_preprocessing")
```
farshidz, line 83: A general note. I wonder if we can do better for Telemetry keys. e.g., have them all in one file, or build them from module name, or something like that (something better than hard-coding them every time). Let's discuss later, no need to change here

src/marqo/core/embed/embed.py
```diff
@@ -0,0 +1,123 @@
...
+        marqo_index = index_meta_cache.get_index(config=self.config, index_name=index_name)
```
farshidz: Also a general note. index meta cache is problematic because it can't be passed as a dependency at class initialization, and requires the config object (in reality it just needs IndexManagement). We need to refactor it

src/marqo/tensor_search/api.py
```diff
@@ -54,6 +56,7 @@ def generate_config() -> config.Config:
 
 
 _config = generate_config()
+_embed = embed_module.Embed(config=_config)
```
farshidz: The standard pattern is to add `Embed` to config, rather than pass config to it. Similar to how `IndexManagement` etc are in config

src/marqo/tensor_search/enums.py
```diff
@@ -78,6 +78,7 @@ class RequestType:
     DELETE = "DELETE"
     CREATE = "CREATE"
     PARTIAL_UPDATE = "PARTIAL_UPDATE"
+    EMBED = "EMBED"     # Request for embeddings of a given input content
```
farshidz: Do we need a separate throttling category? Note this is increasing concurrency. Maybe combine with search?

tests/tensor_search/integ_tests/test_embed.py
```diff
@@ -0,0 +1,438 @@
...
+        # Any tests that call add_documents, search, bulk_search need this env var
+        self.device_patcher = mock.patch.dict(os.environ, {"MARQO_BEST_AVAILABLE_DEVICE": "cpu"})
+        self.device_patcher.start()
+
+    def tearDown(self) -> None:
+        super().tearDown()
+        self.device_patcher.stop()
+
+    def test_embed_equivalent_to_add_docs(self):
+        for index in [self.unstructured_default_text_index, self.structured_default_text_index]:
```
farshidz, line 137: A single line description for each test would be great. Consider following the naming convention discussed before

tests/tensor_search/integ_tests/test_embed.py
```diff
@@ -0,0 +1,438 @@
...
+                mock_vespa_client_query.side_effect = pass_through_query
+
+                @unittest.mock.patch("marqo.vespa.vespa_client.VespaClient.query", mock_vespa_client_query)
+                def run():
+                    tensor_search.search(
+                        config=self.config, index_name=index.name, text="I am the GOAT.",
+                        search_method=enums.SearchMethod.TENSOR
+                    )
+                    return True
+                assert run()
```
farshidz: I think using `self.assert(` creates better error messages in case of failure as it calls the test framework. `assert` is a native python instruction. Especially true for equality assertions so applies to all tests here

farshidz (changes requested):

src/marqo/api/models/embed_request.py
vicilliar: A list can have both strings and dict, and it should work fine. So I think the original type signature is correct.

src/marqo/api/models/embed_request.py
vicilliar: done

src/marqo/api/models/embed_request.py
vicilliar: done

src/marqo/api/models/embed_request.py
vicilliar: fixed

src/marqo/api/models/embed_request.py
vicilliar: done!

src/marqo/core/embed/embed.py
vicilliar: I'll remove this. It was a leftover from a different module.

src/marqo/api/models/embed_request.py
vicilliar: done

src/marqo/core/embed/embed.py
vicilliar: done

src/marqo/core/embed/embed.py
vicilliar: Agreed, let's discuss this another time.

src/marqo/core/embed/embed.py
vicilliar: Noted on this!

src/marqo/tensor_search/api.py
vicilliar: done

src/marqo/tensor_search/enums.py
vicilliar: Done.

tests/tensor_search/integ_tests/test_embed.py
vicilliar: Yep, sorry for missing these! Changing the assert statements.

tests/tensor_search/integ_tests/test_embed.py
vicilliar: done

src/marqo/api/models/embed_request.py
farshidz: Then `Union[str, Dict[str, float], List[Union[str, Dict[str, float]]]]`?

farshidz (approved):

PR #809, opened 2024-04-17
add new validation endpoint

ali60:

feature

Introduce a new endpoint to validate index setting

Introduce a new endpoint to validate index setting

No

yes

N/A
N/A

* **Other information**:


src/marqo/tensor_search/api.py
```diff
@@ -187,6 +188,12 @@ def memory():
     return memory_profiler.get_memory_profile()
 
 
+@app.get('/validation')
```
wanliAlex: Would it be better if we use `POST` and put the settings_object in the request body?

src/marqo/core/schema_validation.py
```diff
@@ -0,0 +1,74 @@
+from enum import Enum
```
wanliAlex: Can you create a module validation in core and put your functions in a class `Validation`? Check the `core->document->document` as an example.

wanliAlex (changes requested):

src/marqo/tensor_search/api.py
ali60: yes that makes sense

src/marqo/tensor_search/api.py
```diff
@@ -191,6 +192,12 @@ def memory():
     return memory_profiler.get_memory_profile()
 
 
+@app.post('/validation')
+@utils.enable_schema_validation()
```
farshidz: `enable_ops_api`? Just we can reuse this for future APIs of the same nature

src/marqo/tensor_search/enums.py
```diff
@@ -70,6 +70,7 @@ class EnvVars:
     MARQO_ENABLE_BATCH_APIS = "MARQO_ENABLE_BATCH_APIS"
     MARQO_ENABLE_UPGRADE_API = "MARQO_ENABLE_UPGRADE_API"
     MARQO_ENABLE_DEBUG_API = "MARQO_ENABLE_DEBUG_API"
+    MARQO_SCHEMA_VALIDATION = "MARQO_SCHEMA_VALIDATION"
```
farshidz: `MARQO_ENABLE_OPS_API`

src/marqo/tensor_search/api.py
```diff
@@ -191,6 +192,12 @@ def memory():
     return memory_profiler.get_memory_profile()
 
 
+@app.post('/validation')
```
farshidz: `validate/index` is better. Can add other validation endpoints with this format later

src/marqo/core/validation/validation.py
```diff
@@ -0,0 +1,74 @@
...
+        On success, it returns 200 and a dictionary representing the settings.
+        On failure due to an ValidationError, it returns 400 and an error message.
+        For any other exception, it returns 500 and an error message.
+    """
+    try:
+        settings_object = json.loads(settings_json)
+        index_settings = IndexSettings(**settings_object)
+        marqo_request = index_settings.to_marqo_index_request(index_name)
+        # Convert the successful marqo_request (if needed) to a dictionary representation
+        convert_marqo_request_to_dict(marqo_request)
```
farshidz: I don't think we need this?

src/marqo/core/validation/validation.py
```diff
@@ -0,0 +1,74 @@
...
+        return JSONResponse(
```
farshidz: We don't return JSON response from Core because it's specific to HTTP APIs. We could raise an exception here (with void return type)

src/marqo/core/validation/validation.py
```diff
@@ -0,0 +1,74 @@
...
+    except ValidationError as e:
+        return JSONResponse(
+            content={
+                "validated": False,
+                "validation_error": str(e),
+                "index": index_name
+            },
+            status_code=400
+        )
+    except Exception as e:
```
farshidz: When do we expect an Exception? If it's unexpected (i.e. indicates a bug) then let's leave it unhandled. We don't return unhandled exception messages in the API since 2.2

src/marqo/core/validation/validation.py
```diff
@@ -0,0 +1,74 @@
+from enum import Enum
+from decimal import Decimal
+from typing import Optional, Tuple
+from marqo.tensor_search.models.index_settings import IndexSettings
+from pydantic import ValidationError
+from fastapi.responses import JSONResponse
+import json
+
+
+def validate_settings_object(index_name, settings_json) -> Tuple[int, Optional[str]]:
```
farshidz: `IndexManagement` is the right place for this code

farshidz (changes requested):

src/marqo/core/validation/validation.py
ali60: yes I can remove it

src/marqo/core/validation/validation.py
ali60: I think earlier Li asked me to move to new validation folder, but sure I can move it to index_management

src/marqo/core/validation/validation.py
ali60: If unhandled exception like a bug happens, then should we return 5xx ? If you don't return anything, how should the client know what happened?

farshidz: If you don't handle it, it'll be a 500 automatically. The API takes care of that

src/marqo/tensor_search/api.py
```diff
@@ -191,6 +193,29 @@ def memory():
     return memory_profiler.get_memory_profile()
 
 
+@app.post('/validate/{index_name}')
```
wanliAlex: We want to have an endpoint as `/validate/index` if `index_name` is not necessary, so we can have `/validate/model` in the future

src/marqo/core/index_management/validation.py
```diff
@@ -0,0 +1,29 @@
+from marqo.tensor_search.models.index_settings import IndexSettings
```
wanliAlex: @farshidz Would it be better to leave this module as separated python file, or should it be included in the index_management class?

src/marqo/core/index_management/validation.py
```diff
@@ -0,0 +1,29 @@
...
+from pydantic import ValidationError
+import json
+import marqo.logging
+
+logger = marqo.logging.get_logger(__name__)
+
+
+def validate_settings_object(index_name, settings_json) -> bool:
```
wanliAlex: nit: type hints

src/marqo/tensor_search/api.py
```diff
@@ -191,6 +193,29 @@ def memory():
...
+@utils.enable_ops_api()
+def schema_validation(index_name: str, settings_object: str):
```
wanliAlex: Why are we reading the object as an `str` and `json.loads()` it later? Can we read it as a dictionary here?

src/marqo/tensor_search/api.py
```diff
@@ -191,6 +193,29 @@ def memory():
...
+                "validated": True,
+                "index": index_name
+            },
+            status_code=200
+        )
+    except ValidationError as e:
+        return JSONResponse(
+            content={
+                "validated": False,
+                "validation_error": str(e),
```
wanliAlex: Should it be camelCase in the API?

src/marqo/tensor_search/api.py
```diff
@@ -191,6 +193,29 @@ def memory():
     return memory_profiler.get_memory_profile()
 
 
+@app.post('/validate/{index_name}')
+@utils.enable_ops_api()
+def schema_validation(index_name: str, settings_object: str):
+    try:
+        validate_settings_object(index_name, settings_object)
```
wanliAlex: @farshidz Is it necessary to create a pydantic class for this function, instead of formating the response in the API?

wanliAlex (changes requested):

src/marqo/core/index_management/validation.py
farshidz: Sorry @ali60 I meant having this as a function within `IndexManagement` class. I agree with @wanliAlex it doesn't need its own module

src/marqo/tensor_search/api.py
farshidz: Yes API response and request JSON must be camel case

src/marqo/tensor_search/api.py
farshidz: I don't mind either way. A pydantic class would be nice, but this is fine too

src/marqo/core/index_management/validation.py
ali60: ok I'll move it

src/marqo/tensor_search/api.py
ali60: index name is not needed, I changed it to this I thought that what @farshidz asked me earlier

ali60: actually index_name is needed because `to_marqo_index_request` method needs it

src/marqo/core/index_management/validation.py
ali60: btw `IndexManagement` looks like a big class with many responsibilities

src/marqo/tensor_search/api.py
ali60: you mean to move `json.loads` here?

src/marqo/tensor_search/api.py
ali60: @wanliAlex changed the signature to @app.post('/validate/index/{index_name}')

src/marqo/core/index_management/index_management.py
```diff
@@ -112,6 +115,32 @@ def create_index(self, marqo_index_request: MarqoIndexRequest) -> MarqoIndex:
...
+            None: If the validation is successful, nothing is returned.
+        """
+        try:
+            index_settings = IndexSettings(**settings_dict)
+            index_settings.to_marqo_index_request(index_name)
+        except ValidationError as e:
+            logger.debug(f'Validation error for index {index_name}: {e}')
+            raise
+        except Exception as e:
+            logger.error(f'Exception while validating index {index_name}: {e}')
```
wanliAlex: We don't need this because the error will be logged in the error handler

wanliAlex (changes requested):

farshidz (dismissed):

farshidz (approved):

wanliAlex (approved):

PR #811, opened 2024-04-19
Add cuda utilization

wanliAlex:
feature

we don't return cuda utilization

we now return cuda utilization

no

yes

no

no

* **Other information**:
no

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ setup(
         "fastapi_utils",
         "opencv-python-headless",
         "psutil",
-        "multilingual_clip,"
+        "multilingual_clip",
         "pynvml==11.5.0"
     ],
     name="marqo-engine",
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,8 @@ setup(
         "fastapi_utils",
         "opencv-python-headless",
         "psutil",
-        "multilingual_clip"
+        "multilingual_clip,"
+        "pynvml==11.5.0"
     ],
     name="marqo-engine",
     version="0.1.10",
--- a/src/marqo/core/models/marqo_cuda_info.py
+++ b/src/marqo/core/models/marqo_cuda_info.py
@@ -8,6 +8,7 @@ class MarqoCudaDeviceInfo(StrictBaseModel):
     memory_used: str
     total_memory: str
     utilization: str
+    memory_used_percent: str
 
 
 class MarqoCudaInfoResponse(StrictBaseModel):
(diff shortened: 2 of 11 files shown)


setup.py
```diff
@@ -32,7 +32,8 @@
         "fastapi_utils",
         "opencv-python-headless",
         "psutil",
-        "multilingual_clip"
+        "multilingual_clip",
```
vicilliar, line 35: Maybe we should stop updating setup.py. It's obsolete now, since Marqo is no longer a PyPI package right? cc @farshidz @pandu-k

src/marqo/core/models/marqo_cuda_info.py
```diff
@@ -0,0 +1,15 @@
+from marqo.base_model import StrictBaseModel
+from typing import List
+
+
+class MarqoCudaDeviceInfo(StrictBaseModel):
```
vicilliar: Let's make the class name match the filename

vicilliar: MarqoCudaInfo could be good, it would also be closer to MarqoCudoInfoResponse

src/marqo/tensor_search/api.py
```diff
@@ -347,8 +347,8 @@ def get_cpu_info():
 
 
 @app.get("/device/cuda")
-def get_cuda_info():
-    return tensor_search.get_cuda_info()
+def get_cuda_info(marqo_config: config.Config = Depends(get_config)):
+    return marqo_config.monitoring.get_cuda_info()
```
vicilliar, line 351: Does this make 1tensor_search.get_cuda_info` obsolete then? Maybe it should be deleted if so.

vicilliar (changes requested):

src/marqo/core/models/marqo_cuda_info.py
wanliAlex: done

src/marqo/tensor_search/api.py
wanliAlex: done

setup.py
wanliAlex: Yeah that makes sense. Maybe we should still update it now, and remove the whole thing in the future?

vicilliar: That's fine with me

vicilliar (dismissed):
Please check if this broke any py-marqo or api tests, since we changed the output of the `get_cuda_info` endpoint

vicilliar (approved):

PR #821, opened 2024-04-29
Adding the prefix feature to Marqo V2

RaynorChavez:
Feature: Prefixes

Currently we do not prepend prefixes by default to e-5 models, nor do we allow prepending custom prefixes to any other model.

e-5 models will have prefixes for `queries` and `chunks` by default, as defined in the model registry. Users will also be able to define their own custom prefixes for `chunks` to be added as documents and their `queries`. 

  In the `embed` endpoint, we have a new field: `content_type` with 3 Options: ‘query’, ‘document’, None. Defaults to ‘query’.
  1. If the user wants to use the default text_query_prefix, leave it as ‘query’.
  2. If the user wants to use the default text_chunk_prefix, leave it as ‘document’.
  3. If the user wants a custom prefix, they must put it in the content itself.

None Identified. the new fields are optional

* **Have unit tests been run against this PR?** (Has there also been any additional testing?
Tests have been created and passed

* **Other information**:


src/marqo/s2_inference/processing/text.py
```diff
@@ -150,5 +150,26 @@ def split_text(text: str, split_by: str = 'sentence', split_length: int = 2, spl
     # assume a uniform seperator when reconstructing the sentences
     text_splits = _reconstruct_multi_list(segments, seperator)
 
+    # Add prefix to each text split if provided
+    if prefix is not None:
+        text_splits = [f"{prefix} {split}" for split in text_splits]
```
vicilliar: There might not be a point in adding prefixes here if we already have a `prefix_text_chunks` function. For context, we want the vectorised content to have prefixes, but we don't want the actual text chunks saved to have prefixes. In essence the prefixes are 
silent". We want to be able to generate a chunk list with and without prefixes.

src/marqo/tensor_search/add_docs.py
```diff
@@ -192,4 +196,51 @@ def determine_document_dict_field_type(field_name: str, field_content, mappings:
...
+
+    # Use the appropriate prefix from index_settings based on prefix_type
+    if prefix_type == "text_query_prefix" and index_settings.text_query_prefix is not None:
+        return index_settings.text_query_prefix
+    elif prefix_type == "text_chunk_prefix" and index_settings.text_chunk_prefix is not None:
+        return index_settings.text_chunk_prefix
+
+
+    # Fallback to model_properties defined prefix
+    model_properties = get_model_properties_from_registry(index_settings.model)
```
vicilliar: 2 things to note:

1. It looks like `MarqoIndex` doesn't have an `IndexSettings` property. Instead, the `Model`, `TextPreProcessing`, etc, are direct properties of it: https://github.com/marqo-ai/marqo/blob/6e40f6ccb5896d14229b86b54f59383bf1ce3c8b/src/marqo/core/models/marqo_index.py#L193
2. The `Model` object has its own attribute `properties` so you probably want to check that first: https://github.com/marqo-ai/marqo/blob/6e40f6ccb5896d14229b86b54f59383bf1ce3c8b/src/marqo/core/models/marqo_index.py#L121

src/marqo/tensor_search/add_docs.py
```diff
@@ -192,4 +196,51 @@ def determine_document_dict_field_type(field_name: str, field_content, mappings:
         else:
             raise base_exceptions.InternalError(f"Invalid dict field type: '{mappings[field_name]['type']}' for field: '{field_name}' in mappings. Must be one of {[t.value for t in enums.MappingsObjectType]}")
     else:
-        return None
\ No newline at end of file
+        return None
+    
+
+
+def determine_text_prefix(request_level_prefix: str, index_settings: IndexSettings, prefix_type: str) -> str:
```
vicilliar: determine_text_chunk_prefix could be a better name, to specify that it's for chunks.

vicilliar: Ah I see, if this is both for query and chunk prefixes, it's best we move this to a different module. Something in the `core` dir would be good.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1287,12 +1293,15 @@ def search(config: Config, index_name: str, text: Optional[Union[str, dict]],
         if approximate is None:
             approximate = True
 
+        marqo_index = index_meta_cache.get_index(config=config, index_name=index_name)
+        text_query_prefix = determine_text_prefix(text_query_prefix, marqo_index.index_settings, "text_query_prefix")
```
vicilliar: there's no point in determining the prefix in both `search` and `vector_text_search`. We can just do this in `vector_text_search`

vicilliar (dismissed):
Also, make sure this is implemented in the `embed` endpoint.

src/marqo/tensor_search/tensor_search.py
RaynorChavez: Oh yeah, thanks for catching that. I forgot to delete the one in `search`

src/marqo/core/embed/embed.py
```diff
@@ -30,7 +31,8 @@ def validate_default_device(cls, value):
     def embed_content(
                     self, content: Union[str, Dict[str, float], List[Union[str, Dict[str, float]]]],
                     index_name: str, device: str = None, image_download_headers: Optional[Dict] = None,
-                    model_auth: Optional[ModelAuth] = None
+                    model_auth: Optional[ModelAuth] = None,
+                    content_type: str = "query"
```
vicilliar: This could be better as an Enum/Constant that has "query" and "document"

src/marqo/core/embed/embed.py
```diff
@@ -67,6 +76,14 @@ def embed_content(
         else:
             raise base_exceptions.InternalError(f"Content type {type(content)} is not supported for embed endpoint.")
 
+        # Decide on the prefix
+        if content_type == "query":
+            prefix = determine_text_prefix(None, marqo_index, "text_query_prefix")
+        elif content_type == "document":
+            prefix = determine_text_prefix(None, marqo_index, "text_chunk_prefix")
+        else:
+            prefix = determine_text_prefix(content_type, marqo_index, "text_query_prefix")
```
vicilliar: There should be an option to have no prefix at all. Maybe when `content_type` is `None`

src/marqo/core/structured_vespa_index/structured_vespa_schema.py
```diff
@@ -164,7 +164,9 @@ def _generate_document_section(self, schema_name: str) -> (List[str], Structured
             created_at=self._index_request.created_at,
             updated_at=self._index_request.updated_at,
             fields=fields,
-            tensor_fields=tensor_fields
+            tensor_fields=tensor_fields,
+            override_text_query_prefix=self._index_request.override_text_query_prefix,
```
vicilliar, line 168: This is good, so the prefix no longer has to be an attribute of schema.

src/marqo/tensor_search/add_docs.py
```diff
@@ -192,4 +197,54 @@ def determine_document_dict_field_type(field_name: str, field_content, mappings:
         else:
             raise base_exceptions.InternalError(f"Invalid dict field type: '{mappings[field_name]['type']}' for field: '{field_name}' in mappings. Must be one of {[t.value for t in enums.MappingsObjectType]}")
     else:
-        return None
\ No newline at end of file
+        return None
+    
+
+def determine_text_prefix(request_level_prefix: str, marqo_index: MarqoIndex, prefix_type: str) -> str:
```
vicilliar: This shouldn't be in add_docs anymore, as it's used for both chunk prefixes and query prefixes

src/marqo/s2_inference/processing/text.py
```diff
@@ -150,5 +150,26 @@ def split_text(text: str, split_by: str = 'sentence', split_length: int = 2, spl
     # assume a uniform seperator when reconstructing the sentences
     text_splits = _reconstruct_multi_list(segments, seperator)
 
+    # Add prefix to each text split if provided
+    if prefix is not None:
+        text_splits = [f"{prefix}{split}" for split in text_splits]
```
vicilliar: Why do we have this if `prefix_text_chunks` already adds prefixes?

src/marqo/tensor_search/add_docs.py
```diff
@@ -192,4 +197,54 @@ def determine_document_dict_field_type(field_name: str, field_content, mappings:
...
+
+    # Check for the presence of the textChunkPrefix or textQueryPrefix in the MarqoIndex object.
+    if prefix_type == "text_query_prefix" and marqo_index.override_text_query_prefix is not None:
+        return marqo_index.override_text_query_prefix
+    elif prefix_type == "text_chunk_prefix" and marqo_index.override_text_chunk_prefix is not None:
+        return marqo_index.override_text_chunk_prefix
+
+    # Fallback to model_properties defined prefix
+    try:
+        model_properties = get_model_properties_from_registry(marqo_index.model.name)
```
vicilliar: Use the `properties` from `model`, since the prefix might be there, before checking the registry

src/marqo/tensor_search/models/api_models.py
```diff
@@ -80,6 +81,8 @@ class BulkSearchQueryEntity(SearchQuery):
 
     context: Optional[SearchContext] = None
     scoreModifiers: Optional[ScoreModifier] = None
+    text_query_prefix: Optional[str] = None
+    #text_chunk_prefix: Optional[str] = None
```
vicilliar: remove this comment

src/marqo/tensor_search/tensor_search.py
```diff
@@ -297,7 +299,8 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
                             split_overlap = marqo_index.text_preprocessing.split_overlap
                             content_chunks: List[str] = text_processor.split_text(field_content, split_by=split_by,
                                                                                   split_length=split_length,
-                                                                                  split_overlap=split_overlap)
+                                                                                  split_overlap=split_overlap,
+                                                                                  prefix=text_chunk_prefix)
```
vicilliar: we want to be able to have content_chunks (with prefix) and text_chunks (without prefix). This way we can store the text in vespa with no prefix, but the embeddings will be affected by it. That's why it's better to have the prefixing not be in the `split_text` function.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -558,6 +562,7 @@ def _add_documents_structured(config: Config, add_docs_params: AddDocsParams, ma
     total_vectorise_time = 0
     batch_size = len(add_docs_params.docs)  # use length before deduplication
     image_repo = {}
+    text_chunk_prefix = determine_text_prefix(add_docs_params.text_chunk_prefix, marqo_index, "text_chunk_prefix")
```
vicilliar: It's redundant for the string to be "text_chunk_prefix", what if it's just "text_chunk" or "text_query".

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1695,14 +1705,61 @@ def get_content_vector(possible_jobs: List[VectorisedJobPointer], job_to_vectors
...
+    for q in queries:
+        text_query_prefix = determine_text_prefix(q.text_query_prefix, q.index, "text_query_prefix")
 
-def run_vectorise_pipeline(config: Config, queries: List[BulkSearchQueryEntity], device: Union[Device, str]) -> Dict[
-    Qidx, List[float]]:
+        if q.q is None:
+            prefixed_q = q.q
+        elif isinstance(q.q, str):
+            if _is_image(q.q) and ((isinstance(q.index, UnstructuredMarqoIndex) and q.index.treat_urls_and_pointers_as_images) or
+                (isinstance(q.index, StructuredMarqoIndex))):
```
vicilliar: We would have to check if the field is an `image_pointer` or not, if it's a structured index. If it's an image pointer, then it should have no prefix.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -2105,10 +2168,12 @@ def vectorise_multimodal_combination_field_structured(
         start_time = timer()
         text_vectors = []
         if len(text_content_to_vectorise) > 0:
+            
             with RequestMetricsStore.for_request().time(f"create_vectors"):
+                prefixed_text_content_to_vectorise = text_processor.prefix_text_chunks(text_content_to_vectorise, text_chunk_prefix)
```
vicilliar, line 2172: can we verify through tests that this does not add double prefix? Like text_content_to_vectorise has prefixes, then this adds another prefix?

tests/tensor_search/integ_tests/test_embed.py
```diff
@@ -173,6 +173,53 @@ def test_embed_equivalent_to_add_docs(self):
...
+                self.assertEqual(len(get_docs_res[enums.TensorField.tensor_facets]), 1)
+
+                get_docs_embedding = get_docs_res[enums.TensorField.tensor_facets][0][enums.TensorField.embedding]
+
+                # Embed request the same text
+                embed_res = embed(
+                    marqo_config=self.config, index_name=index.name,
+                    embedding_request=EmbedRequest(
+                        content=["I am the GOAT."],
+                        content_type="PREFIX: "
```
vicilliar: I'm confused about this. Shouldn't content_type be either document, query, or None?

tests/tensor_search/integ_tests/test_embed.py
```diff
@@ -218,6 +265,53 @@ def run():
...
+                elif isinstance(index, StructuredMarqoIndex):
+                    embedding_key = "marqo__query_embedding"
+                search_query_embedding = vespa_query_kwargs["query_features"][embedding_key]
+
+                # Embed request the same text
+                embed_res = embed(
+                    marqo_config=self.config, index_name=index.name,
+                    embedding_request=EmbedRequest(
+                        content=["I am the GOAT."],
+                        content_type="PREFIX: "
```
vicilliar: Ideally, to get the same behavior, use the test model that has a prefix, and check that both embed and search use the default query prefix set for this model

tests/tensor_search/test_add_documents_use_existing_tensors.py
```diff
@@ -350,6 +354,77 @@ def test_use_existing_tensor_multimodal_changed(self):
                     self.assertEqual('{"text_field_1": "content 1", "text_field_2": "content 2-updated"}',
                                      get_doc_res["_tensor_facets"][0]["multimodal_field"])
 
+    def test_determine_text_chunk_prefix(self):
```
vicilliar: I think this is in the wrong file, this is for add documents with use existing tensors feature.

Also, please combine this with the test_determine_text_query_prefix tests, as they're essentially both unit tests testing the same function `determine_text_prefix`.

vicilliar (changes requested):

src/marqo/tensor_search/tensor_search.py
RaynorChavez: We're only getting the prefix here, which we then pass on to various functions in different parts of the function. so setting it to `text_chunk` or `text_query` would make its use unclear. Perhaps `text_prefix` would be better?

RaynorChavez: Actually `text_chunk_prefix` would be good too since we don't expect `text_query_prefix` here in the `_add_documents_structured` function

src/marqo/tensor_search/tensor_search.py
RaynorChavez: I initially did, but I since this is for queries, I'm not sure that we need to - is there a way for users to define whether a query is an image pointer or not?

src/marqo/tensor_search/tensor_search.py
RaynorChavez: Afaik, we don't prepend prefixes to text_content_to_vectorise, whose content is taken from `field_content`, which is not modified in both `_add_documents_unstructured` and `_add_documents_structured` in the code block below `# All the plain tensor/non-tensor fields are processed, now we process the multimodal fields`, where  `vectorise_multimodal_combination_field_unstructured` is called from. 

We only added prefixing explicitly in `_add_documents_unstructured` and `_add_documents_structured` in the code block `# C) field type is standard`. 

I did some tests on this yesterday and seem to have confirmed that no double prefixing happens. 

We can write a more visible/specific test for this though

tests/tensor_search/test_add_documents_use_existing_tensors.py
RaynorChavez: Thanks, I'll move it to the `test_prefix.py` file. 

As for the `test_determine_text_query_prefix` from `test_search.py`, I initially put this exact test there (hence the duplication), but found that we skip `test_search.py` making it irrelevant, so I can't really combine them. I'll just remove `test_determine_text_query_prefix` from there for clarity.

src/marqo/tensor_search/tensor_search.py
RaynorChavez: I think with what we did before, the text chunks are stored normally already with only the prefix going to the embeddings directly. 

But you are right that it's better to separate the prefixing logic from `split_text`. I have since just wrapped the output of `split_text` with `prefix_text_chunks` to get the prefixed version according to your suggestion. Thanks!

src/marqo/tensor_search/tensor_search.py
RaynorChavez: Actually I think `test_prefix_multimodal` shows this already

src/marqo/tensor_search/tensor_search.py
RaynorChavez: I consulted Li on this, and he explained that, in queries, all URLs pointing to images are treated as images and not text, so we can just simplify the condition to `if _is_image(q.q)`

scripts/vespa_local/vespa_local.py
```diff
@@ -24,7 +24,7 @@ def restart(args):
 def deploy_config(args):
     os.system('vespa config set target local')
     here = os.path.dirname(os.path.abspath(__file__))
-    os.system(f'vespa deploy {here}')
+    os.system(f'vespa deploy "{here}"')
```
vicilliar: Why was this added? Was it for path names with a space?

src/marqo/core/embed/embed.py
```diff
@@ -67,6 +82,16 @@ def embed_content(
         else:
             raise base_exceptions.InternalError(f"Content type {type(content)} is not supported for embed endpoint.")
 
+        # Decide on the prefix 
+        if content_type == EmbedContentType.Query:
+            prefix = determine_text_prefix(None, marqo_index, "text_query_prefix")
```
vicilliar: Remove "prefix", it seems redundant given the function. Maybe something like `prefix = determine_text_prefix(None, marqo_index, "text_query")`

src/marqo/core/embed/embed.py
```diff
@@ -67,6 +82,16 @@ def embed_content(
...
+        elif content_type == EmbedContentType.Document:
+            prefix = determine_text_prefix(None, marqo_index, "text_chunk_prefix")
```
vicilliar: same comment as above

src/marqo/core/embed/embed.py
```diff
@@ -67,6 +82,16 @@ def embed_content(
...
+        elif content_type is None:
+            prefix = ""
+        else:
+            raise ValueError(f"Invalid content_type: {content_type}. Must be EmbedContentType.Query, EmbedContentType.Document, or None.")
```
vicilliar: Must be "query", "document", or None (if no prefix is to be added). Don't expose the `EmbedContentType` enum to users.

src/marqo/core/utils/prefix.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Returns:
+        str: The determined prefix, or None if no prefix is found
+    """
+    if request_level_prefix is not None:
+        return request_level_prefix
+    
+
+    # Check for the presence of the textChunkPrefix or textQueryPrefix in the MarqoIndex object.
+    if prefix_type == "text_query_prefix" and marqo_index.override_text_query_prefix is not None:
```
vicilliar: same comment, just make it "text_query" and "text_chunk". Consider using an enum for these instead. Since there are only 2 distinct options.

src/marqo/core/utils/prefix.py
```diff
@@ -0,0 +1,62 @@
...
+        if marqo_index.model is not None:
+            model_properties = marqo_index.model.properties
+            if model_properties is not None:
+                default_prefix = model_properties.get(prefix_type)
+                if default_prefix is not None:
+                    return default_prefix
+        
+        model_properties = get_model_properties_from_registry(marqo_index.model.name)
+
+        if model_properties is not None:
```
vicilliar, line 45: You're duplicating the code:

```
if model_properties is not None:
            default_prefix = model_properties.get(prefix_type)
            if default_prefix is not None:
                return default_prefix
```
Maybe we can move the `model_properties = get_model_properties_from_registry(marqo_index.model.name)` line up and make it only execute when the `marqo_index.model.properties` is `None`

src/marqo/core/utils/prefix.py
```diff
@@ -0,0 +1,62 @@
...
+            default_prefix = model_properties.get(prefix_type)
+            if default_prefix is not None:
+                return default_prefix
+            else:
+                return ""
+        else:
+            raise ValueError(f"Model properties not found for model: {marqo_index.model}")
+    except InvalidModelPropertiesError as e:
+        pass
```
vicilliar: Not sure if we should just ignore this. If there's a `InvalidModelPropertiesError`, it should be raised.

src/marqo/tensor_search/tensor_search.py
vicilliar: Ah that's not what I meant, the variable name is good as `text_chunk_prefix`. I meant the 3rd argument for the `determine_text_prefix` function.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -2024,7 +2085,7 @@ def vectorise_multimodal_combination_field_unstructured(field: str,
 def vectorise_multimodal_combination_field_structured(
         field: str, multimodal_object: Dict[str, dict], doc: dict, doc_index: int,
         doc_id: str, device: str, marqo_index: StructuredMarqoIndex, image_repo, field_map: dict,
-        model_auth: Optional[ModelAuth] = None
+        model_auth: Optional[ModelAuth] = None, text_chunk_prefix: str = ""
```
vicilliar: To keep the defaults consistent, the default for `text_chunk_prefix` should also be `None`

src/marqo/tensor_search/tensor_search.py
vicilliar: Yep alright, just confirming! Sounds good then.

tests/tensor_search/integ_tests/test_embed.py
```diff
@@ -102,22 +108,38 @@ def setUpClass(cls) -> None:
             tensor_fields=["text_field_1", "text_field_2", "image_field_1"]
         )
 
+        structured_image_index_with_e5_small = cls.structured_marqo_index_request(
+            model=Model(name='hf/e5-small'),
```
vicilliar: It's probably better to use the test prefix model for these tests, since if we decide to change the `hf/e5-small` default prefixes, these tests will break.

tests/tensor_search/test_search.py
```diff
@@ -1372,3 +1376,74 @@ def run():
...
 
         assert run()
+
+    def test_determine_text_query_prefix(self):
+        """
+        Ensures proper priority order is followed when determining the query prefix.
+        search request-level > index override-level > model default level
+        """
+
+        marqo_index_with_model_default = UnstructuredMarqoIndex(
```
vicilliar: I don't think we should be creating indexes directly like this, as it makes cleanup unpredictable as well.
1. Instead, use `cls.create_indexes` and create all your indexes in `setUpClass`. 
2. Also, `tests/tensor_search/test_search.py` is skipped, afaik. This should be in `test_search_combined` under integ tests.

vicilliar (changes requested):

scripts/vespa_local/vespa_local.py
RaynorChavez: Yes, for some reason it did not get included when I merged mainline. We just wrapped the pathname with " "

tests/tensor_search/test_search.py
RaynorChavez: Apologies, this should have been deleted in the last round. It's updated counterpart is test_determine_text_chunk_prefix in test_prefix.py

src/marqo/core/utils/prefix.py
RaynorChavez: revised!

src/marqo/core/utils/prefix.py
RaynorChavez: made a new class

```
class DeterminePrefixContentType(Enum):
    TextChunk = "text_chunk_prefix"
    TextQuery = "text_query_prefix"
```

for these!

src/marqo/core/embed/embed.py
RaynorChavez: revised!

src/marqo/core/embed/embed.py
RaynorChavez: Regarding this, I am removing `EmbedContentType` so we don't have to expose the `EmbedContentType` to users. We now take string, as before, but added a pydantic validator to limit the responses

```
@pydantic.validator('content_type', pre=True, always=True)
    def validate_content_type(cls, v):
        if v is not None and v not in ["query", "document"]:
            raise ValueError("content_type must be 'query', 'document', or None")
        return v
```

this can be found int `embed_request.py`

src/marqo/tensor_search/tensor_search.py
RaynorChavez: revised, as above, we are now using the enum `DeterminePrefixContentType`

src/marqo/tensor_search/tensor_search.py
RaynorChavez: revised!

tests/tensor_search/integ_tests/test_embed.py
RaynorChavez: revised!

src/marqo/api/models/embed_request.py
```diff
@@ -4,26 +4,19 @@
...
+
 
 
 class EmbedRequest(BaseMarqoModel):
     # content can be a single query or list of queries. Queries can be a string or a dictionary.
     content: Union[str, Dict[str, float], List[Union[str, Dict[str, float]]]]
     image_download_headers: Optional[Dict] = None
     modelAuth: Optional[ModelAuth] = None
-    content_type: Optional[EmbedContentType] = EmbedContentType.Query
+    content_type: Optional[str] = "query"
```
vicilliar: Why was the `EmbedContentType` Enum removed? I think it was correct to have that. We just shouldn't expose the enum name to users in error messages.

vicilliar (changes requested):

src/marqo/api/models/embed_request.py
RaynorChavez: Reimplemented, thanks!

src/marqo/s2_inference/processing/text.py
RaynorChavez: removed and wrapped the output of split_text with prefix_text_chunks

vicilliar (approved):

PR #832, opened 2024-05-08
prefix refactor and backwards compatibility

RaynorChavez:
Bug fix

As previously implemented, users who have "embed", "add_documents", and "query" requests on old indexes with e5 models will no longer receive the same output as before, since they will not have known/implemented about the “content_type” field in embed, and will not have set their own prefixes on the (1) request level, and (2) index settings. Therefore, marqo prefixes their documents with the model default prefixes. 

Cases:
1. Old indexes, when read by marqo 2.6 will have the new fields `text_query_prefix` and `text_chunk_prefix` populated by `None`, as is the default in `MarqoIndex` and `IndexSettings`
2. New indexes user populated with values in `text_query_prefix` and `text_chunk_prefix` will use these as the prefixes for the model, unless defined otherwise at the request level. This includes empty strings `""`
3. New indexes without user-populated values  in `text_query_prefix` and `text_chunk_prefix` 

This PR fixes the breaking change from the original prefix feature implementation

Yes. Also ran against `marqo-api-tests`

* **Other information**:


src/marqo/core/models/marqo_index.py
```diff
@@ -188,6 +190,26 @@ def _update_model_properties_from_registry(self) -> None:
                 raise InvalidArgumentError(
                     f'Invalid model properties for model={model_name}. Reason: {e}.'
                 )
+            
+    def _populate_prefix_fields(self) -> None:
+        if self.text_query_prefix is None or self.text_query_prefix == "":
```
RaynorChavez: I found a bug (?) wherein marqo transforms text_query_prefix and text_query_prefix from an explicit None from the user into an empty string "". In the future, I would like to explore the possibility of preserving the NoneType in the index creation process, so that the population can look like this:

1. if the user doesn't set anything for these fields or explicitly set as None, the prefix will be taken from model defaults
2. if the user explicitly sets prefix to be any string (even empty ""), then we can set the prefix to that

Currently, the user is not able to set an empty prefix directly due to the bug where NoneType assignments don't seem to get preserved, so the only prefixes available are the (1) model defaults, or (2) any user-defined string with len > 0.

But I think this is minor and we could patch it in the future.

src/marqo/core/embed/embed.py
```diff
@@ -83,10 +83,14 @@ def embed_content(
...
+
+        # For backwards compatibility
+        if marqo_index.model.text_query_prefix is None or marqo_index.model.text_chunk_prefix is None:
+            prefix = ""
+        elif content_type == EmbedContentType.Query:
+            prefix = marqo_index.get_text_query_prefix(None)
         elif content_type == EmbedContentType.Document:
-            prefix = determine_text_prefix(None, marqo_index, DeterminePrefixContentType.TextChunk)
+            prefix = marqo_index.get_text_chunk_prefix(None)
         elif content_type is None:
```
farshidz, line 94: Is this valid input `content_type=None`?

src/marqo/core/models/marqo_index.py
```diff
@@ -188,6 +190,26 @@ def _update_model_properties_from_registry(self) -> None:
...
+            self.text_query_prefix = self._get_default_prefix("text_query_prefix")
+        else:
+            self.text_query_prefix = self.text_query_prefix
```
farshidz: This doesn't do anything

src/marqo/core/models/marqo_index.py
```diff
@@ -188,6 +190,26 @@ def _update_model_properties_from_registry(self) -> None:
...
+       
+        if self.text_chunk_prefix is None or self.text_chunk_prefix == "":
+            self.text_chunk_prefix = self._get_default_prefix("text_chunk_prefix")
+        else:
+            self.text_chunk_prefix = self.text_chunk_prefix
```
farshidz: Doesn't do anything

src/marqo/core/models/marqo_index.py
```diff
@@ -188,6 +190,26 @@ def _update_model_properties_from_registry(self) -> None:
...
+        
+
+    def _get_default_prefix(self, prefix_type: str) -> Optional[str]:
+        try:
+            model_properties = self.get_properties()
+            default_prefix = model_properties.get(prefix_type)
+            return default_prefix
+        except InvalidModelPropertiesError:
```
farshidz: When does this happen? I don't think the code in try will ever raise this

src/marqo/core/models/marqo_index.py
```diff
@@ -188,6 +190,26 @@ def _update_model_properties_from_registry(self) -> None:
...
+            return None
```
farshidz: Not sure when we expect to return None?

src/marqo/core/models/marqo_index.py
```diff
@@ -270,6 +291,34 @@ def _cache_or_get(self, key: str, func):
         if key not in self._cache:
             self._cache[key] = func()
         return self._cache[key]
+    
+    def get_text_query_prefix(self, request_level_prefix: Optional[str] = None) -> str:
```
farshidz: Looks like these two should also be under `Model`. Also the call to `self.model._populate_prefix_fields()` should be in `Model.init`

farshidz (changes requested):

src/marqo/core/embed/embed.py
RaynorChavez: Yes, returns the MarqoIndex stored prefixes. The method has only one parameter: the request level prefix.

src/marqo/core/models/marqo_index.py
RaynorChavez: Function deleted

src/marqo/core/embed/embed.py
RaynorChavez: Though we can just delete the None and use marqo_index.get_text_chunk_prefix() instead :D

src/marqo/core/models/marqo_index.py
RaynorChavez: Removed the try block since validation already happens in get_properties() and .get() will return None if it doesn't exist.

src/marqo/core/models/marqo_index.py
RaynorChavez: Removed the try block since validation already happens in get_properties() and .get() will return None if it doesn't exist.

src/marqo/core/models/marqo_index.py
RaynorChavez: Moved them to Model.

Also, regarding why i put request_level_prefix here instead of just directly using them in tensor_search is for easier code readability. I did not want to have to repeat the conditional to check whether the request level prefix is none or a string, so that logic was moved here. Please let me know if you have better ideas!

src/marqo/core/index_management/index_management.py
```diff
@@ -161,6 +167,14 @@ def batch_create_indexes(self, marqo_index_requests: List[MarqoIndexRequest]) ->
                 if self.index_exists(index.name):
                     raise IndexExistsError(f"Index {index.name} already exists")
 
+
+        # Populate the prefix fields if they are None
+        for index in marqo_index_requests:
+            if index.model.text_query_prefix is None:
+                index.model.text_query_prefix = index.model._get_default_prefix("text_query_prefix")
+            if index.model.text_chunk_prefix is None:
+                index.model.text_chunk_prefix = index.model._get_default_prefix("text_chunk_prefix")
```
farshidz: Two issues here. 1. you're not covering batch create index. 2. you're calling a private method of another class here (your IDE will be complaining too)

farshidz: Maybe just add a public `get_default_text_query_prefix` and `get_default_text_chunk_prefix`. Add an index management test to make sure this works as expected

src/marqo/core/models/marqo_index.py
```diff
@@ -188,6 +190,39 @@ def _update_model_properties_from_registry(self) -> None:
...
+
+        # For backwards compatibility. Since older versions of Marqo did not have a text_query_prefix field,
+        # we need to return an empty string if the model does not have a text_query_prefix. 
+        # We know that the value of text_query_prefix is None in old indexes since the model was not populated
+        # from the registry.
+        if self.text_query_prefix is None:
+            return ""
+
+        # Else return the model default as populated during initialization
+        return self.text_query_prefix or ""
```
farshidz: You've already covered None so just `return self.text_query_prefix` will suffice

src/marqo/core/models/marqo_index.py
```diff
@@ -188,6 +190,39 @@ def _update_model_properties_from_registry(self) -> None:
...
+
+        # For backwards compatibility. Since older versions of Marqo did not have a text_chunk_prefix field,
+        # we need to return an empty string if the model does not have a text_chunk_prefix. 
+        # We know that the value of text_chunk_prefix is None in old indexes since the model was not populated
+        # from the registry.
+        if self.text_chunk_prefix is None:
+            return ""
+
+        # Else return the model default as populated during initialization
+        return self.text_chunk_prefix or ""
```
farshidz: Just `return self.text_chunk_prefix`

farshidz (changes requested):

src/marqo/core/models/marqo_index.py
RaynorChavez: Removed

src/marqo/core/index_management/index_management.py
RaynorChavez: Added public methods! though `batch_create_indexes` was also covered previously

src/marqo/core/models/marqo_index.py
RaynorChavez: Removed

farshidz (approved):

PR #838, opened 2024-05-15
Fix Concurrent Deployment Issue via Zookeeper

wanliAlex:
Bug fix, feature

We don't have any concurrency protection on create/delete index.
If multiple threads are creating/deleting indexes at the same time, Marqo will face a race condition

We use the Zookeeper inside Vespa to implement a distributed lock. The lock will solve the concurrent issue.

no

in progress

no
no

* **Other information**:
no


src/marqo/api/exceptions.py
```diff
@@ -195,6 +195,11 @@ class ModelNotInCacheError(__InvalidRequestError):
     status_code = HTTPStatus.NOT_FOUND
 
 
+class ConflictError(__InvalidRequestError):
```
farshidz: I wonder if we could be more explicit that "conflict error"? Do you foresee this being used for other conflicts?

src/marqo/core/distributed_lock/deployment_lock.py
```diff
@@ -0,0 +1,116 @@
...
+
+from marqo.core.distributed_lock.abstract_distributed_lock import AbstractExpiringDistributedLock
+from marqo.core.distributed_lock.marqo_kazoo_client import MarqoKazooClient
+from marqo.core.exceptions import ConflictError, BackendCommunicationError
+from marqo.logging import get_logger
+
+logger = get_logger(__name__)
+
+
+class DeploymentLock(AbstractExpiringDistributedLock):
```
farshidz: What's deployment specific about this? I think this could be a generic lock implementation without deployment references

farshidz: What is specific about this is that it uses zookeeper. So I would have Zookeeper in the name instead. e.g., `ZookeeperExpiringDistributedLock` or `ZookeeperDistributedLock` (you could make the expiring property optional?)

src/marqo/core/distributed_lock/abstract_distributed_lock.py
```diff
@@ -0,0 +1,61 @@
...
+    @abstractmethod
+    def __init__(self, zookeeper_client: MarqoKazooClient, path: str, max_lock_period: float, watchdog_interval: float):
+        """Initialize the distributed lock.
+        Args:
+            zookeeper_client: The Zookeeper client.
+            path: The lock path.
+            max_lock_period: The maximum period of time the lock can be held.
+            watchdog_interval: The interval at which the watchdog checks the lock status.
+        """
+        self.zookeeper_client = zookeeper_client
```
farshidz: I think this abstraction is a bit questionable. Lots of Zookeeper code is leaking into the abstraction so it's not independent of Zookeeper. Also see my comments on the implementation class below

src/marqo/core/distributed_lock/deployment_lock.py
```diff
@@ -0,0 +1,116 @@
...
+        conflict_error_message: The error message to display when the lock is not acquired. If not provided, we use the
+            class object's error message.
+
+    Raises:
+        RuntimeError: If the lock is not instantiated and called.
+        BackendCommunicationError: If the Zookeeper client cannot connect to the server.
+        ConflictError: If the lock is not acquired.
+    """
+    if lock is None:
+        raise RuntimeError("Deployment lock is not instantiated and cannot be used for index creation/deletion")
```
farshidz: InternalError might be more consistent with the codebase

farshidz: Why do you permit None if it's an error?

src/marqo/core/distributed_lock/abstract_distributed_lock.py
```diff
@@ -0,0 +1,61 @@
...
+        self.lock = Lock(zookeeper_client, path)
```
farshidz: How does this constructor take a custom Marqo class for Zookeeper?

src/marqo/core/distributed_lock/deployment_lock.py
```diff
@@ -0,0 +1,116 @@
...
+    elif lock.zookeeper_client.state != KazooState.CONNECTED:
+        try:
+            lock.zookeeper_client.start(5)
```
farshidz: Is there a reason not to have the logic to handle Zookeeper connection and errors in your `acquire` method? Accessing `lock.zooekeepr_client` seems to violate the lock class's encapsulation

src/marqo/core/distributed_lock/abstract_distributed_lock.py
```diff
@@ -0,0 +1,61 @@
...
+        self.max_lock_period = max_lock_period
```
farshidz: All these variable should start with `_` to mark them as internal/private. We don't seem to be following this pattern in our classes (mostly my own code) and that's something to address as well later

src/marqo/core/index_management/index_management.py
```diff
@@ -55,8 +55,27 @@ class IndexManagement:
         '''
     )
 
-    def __init__(self, vespa_client: VespaClient):
+    _DEPLOYMENT_LOCK_PATH = '/marqo__deployment_lock'
```
farshidz: Having all lock paths in one place makes it easier to ensure to clashes. So perhaps you could have a `lock_deployment` method somewhere in a class that will have one method per distributed lock we're using for all of Marqo

src/marqo/core/index_management/index_management.py
```diff
@@ -55,8 +55,27 @@ class IndexManagement:
...
+    _DEPLOYMENT_LOCK_WATCHDOG_INTERVAL = 5
+
+    def __init__(self, vespa_client: VespaClient, zookeeper_client: Optional[MarqoKazooClient] = None):
         self.vespa_client = vespa_client
+        self.zookeeper_client = zookeeper_client
+        self.deployment_lock: Optional[DeploymentLock] = self._instantiate_deployment_lock()
+
+    def _instantiate_deployment_lock(self) -> Optional[DeploymentLock]:
+        """Instantiate a DeploymentLock"""
+        if self.zookeeper_client is None:
```
farshidz: I believe we discussed making zookeeper required and not allowing deployments without it?

src/marqo/core/distributed_lock/deployment_lock.py
```diff
@@ -0,0 +1,116 @@
...
+        """
+        super().__init__(zookeeper_client, path, max_lock_period, watchdog_interval)
+        self.acquire_timeout = acquire_timeout
+        self.error_message = error_message
+
+    def _watchdog(self):
+        """Watchdog thread to monitor lock status and release the lock if it is held for too long."""
+        while True:
+            if self.lock_acquired_time and (time.time() - self.lock_acquired_time) > self.max_lock_period:
+                self.release()
```
farshidz: How is the running process notified of this release?

src/marqo/core/distributed_lock/marqo_kazoo_client.py
```diff
@@ -0,0 +1,12 @@
+from kazoo.client import KazooClient
+
+
+class MarqoKazooClient(KazooClient):
```
farshidz: I wonder if `marqo.vespa` is a better place for this. Also you could call is `ZookeeperClient`, similar to `VespaClient`. Kazoo is a library to access Zookeepr, this is not a Kazoo client.

farshidz (changes requested):

src/marqo/core/distributed_lock/abstract_distributed_lock.py
wanliAlex: I have updated the abstraction and it is decoupled with the zookeeper

src/marqo/core/distributed_lock/abstract_distributed_lock.py
wanliAlex: I have updated the abstraction and it is decoupled with the zookeeper

src/marqo/core/distributed_lock/abstract_distributed_lock.py
wanliAlex: I have updated this in the concrete implementation.

src/marqo/core/distributed_lock/deployment_lock.py
wanliAlex: I use `ZookeeperDistributedLock` and remove the watch dog as discussed

src/marqo/core/distributed_lock/deployment_lock.py
wanliAlex: removed this watch dog

src/marqo/core/distributed_lock/deployment_lock.py
wanliAlex: There isn't an internal in `core`. Is it ok to raise this error and let it be caught as an internal error?

wanliAlex: None is for an instantiate issue in `embed` class

src/marqo/core/distributed_lock/deployment_lock.py
wanliAlex: I have moved this into the acquire function

src/marqo/core/index_management/index_management.py
wanliAlex: I create a python module to store the function

src/marqo/core/index_management/index_management.py
wanliAlex: Due to instantiation issue in `embed` class

src/marqo/api/exceptions.py
wanliAlex: Updated to IndexCreationAndDeletionConflickError

src/marqo/core/distributed_lock/marqo_kazoo_client.py
wanliAlex: I don't think it's the perfect place as the directory is named as `vespa`. But I think this is better than `core` as this client should parallel with `vespa_client`

src/marqo/core/distributed_lock/distributed_lock_manager.py
```diff
@@ -0,0 +1,17 @@
+from marqo.vespa.marqo_zookeeper_client import MarqoZookeeperClient
+from marqo.core.distributed_lock.zookeeper_distributed_lock import ZookeeperDistributedLock
+
+_DEPLOYMENT_LOCK_PATH = "/marqo__deployment_lock"
+
+
+def get_deployment_lock(zookeeper_client: MarqoZookeeperClient, acquire_timeout=1) -> ZookeeperDistributedLock:
```
farshidz: Minor, but simply `ZookeeperClient` is more consistent with `VespaClient`

farshidz: I wonder if you could simplify this and add it to `zookeeper_distributed_lock.py`. It makes even more sense when you realize this is just a factory method for the lock class. Really good pattern otherwise

src/marqo/core/distributed_lock/zookeeper_distributed_lock.py
```diff
@@ -0,0 +1,79 @@
...
+            try:
+                self._zookeeper_client.start()
+            except KazooTimeoutError as e:
+                raise BackendCommunicationError("Marqo cannot connect to backend concurrent manager. "
+                                                "Your request cannot be processed at this time. "
+                                                "Please check your network settings and try again later.") from e
+        try:
+            acquired = self._lock.acquire(timeout=self._acquire_timeout)
+            if not acquired:
+                raise ZooKeeperLockNotAcquiredError("Failed to acquire the lock.")
```
farshidz: Why/when does this happen?

farshidz: `Zookeeper` not `ZooKeeper`

src/marqo/core/distributed_lock/zookeeper_distributed_lock.py
```diff
@@ -0,0 +1,79 @@
...
+            else:
+                return True
+        except ConnectionClosedError:
+            raise BackendCommunicationError("Marqo cannot connect to backend concurrent manager "
+                                            "when acquiring the lock. "
+                                            "Your request cannot be processed at this time. "
+                                            "Please check your network settings and try again later.")
```
farshidz: Will the user see this message? I think these should all be 500 generic failures because there's not much the user can do

src/marqo/core/distributed_lock/zookeeper_distributed_lock.py
```diff
@@ -0,0 +1,79 @@
...
+        except LockTimeout:
+            raise ZooKeeperLockNotAcquiredError("Failed to acquire the lock.")
```
farshidz: The convention is no period at the end of exception messages

src/marqo/core/exceptions.py
```diff
@@ -12,6 +12,18 @@ class IndexNotFoundError(MarqoError):
     pass
 
 
+class IndexCreationAndDeletionConflictError(MarqoError):
```
farshidz: Maybe something more generic like `ConcurrencyConflictError`. Not the best name but you get the idea

src/marqo/core/index_management/index_management.py
```diff
@@ -90,37 +103,48 @@ def create_index(self, marqo_index_request: MarqoIndexRequest) -> MarqoIndex:
...
             InvalidVespaApplicationError: If Vespa application is invalid after applying the index
+            RuntimeError: If deployment lock is not instantiated
+            IndexCreationAndDeletionConflictError: If another index creation/deletion operation is
+                in progress and the lock cannot be acquired
         """
-        app = self.vespa_client.download_application()
-        configured = self._marqo_config_exists(app)
+        
+        if self._zookeeper_deployment_lock is None:
+            raise RuntimeError("Deployment lock is not instantiated and cannot be used for index creation/deletion")
```
farshidz: Could be `InternalError`? The conflict error might be applicable here too

src/marqo/core/index_management/index_management.py
```diff
@@ -158,51 +182,62 @@ def batch_create_indexes(self, marqo_index_requests: List[MarqoIndexRequest]) ->
...
-        Delete a Marqo index.
+        Delete a Marqo index. To make this operation thread-safe, use delete_index_by_name instead.
 
         This method is idempotent and does not raise an error if the index does not exist.
 
         Args:
             marqo_index: Marqo index to delete
         """
-        app = self.vespa_client.download_application()
+        app = self.vespa_client.download_application(wait_for_application_convergence=True)
```
farshidz: This should be `check_for_application_convergence` and error out with concurrency exception if hasn't converged. I wouldn't wait for convergence here

src/marqo/core/index_management/index_management.py
```diff
@@ -158,51 +182,62 @@ def batch_create_indexes(self, marqo_index_requests: List[MarqoIndexRequest]) ->
...
             IndexExistsError: If an index already exists
             InvalidVespaApplicationError: If Vespa application is invalid after applying the indexes
+            RuntimeError: If deployment lock is not instantiated
+            IndexCreationAndDeletionConflictError: If another index creation/deletion operation is
+                in progress and the lock cannot be acquired
         """
-        app = self.vespa_client.download_application()
-        configured = self._add_marqo_config(app)
+        if self._zookeeper_deployment_lock is None:
+            raise RuntimeError("Deployment lock is not instantiated and cannot be used for index creation/deletion")
```
farshidz: This check and error is repeated a number of times. Better to wrap it in an internal function and call that. You could even have a custom context manager (internal one) that wraps the lock context manager and performs this check as well. Both options fine

src/marqo/vespa/vespa_client.py
```diff
@@ -110,6 +114,9 @@ def download_application(self) -> str:
         Returns:
             Path to the downloaded application
         """
+        if wait_for_application_convergence:
+            self.wait_for_application_convergence()
```
farshidz: As noted I suggest a check and error if hasn't converged, rather than wait

tests/core/distributed_lock/test_distributed_lock.py
```diff
@@ -0,0 +1,109 @@
...
+        lock2 = ZookeeperDistributedLock(self.zookeeper_client, "/lock2")
+        self.assertTrue(lock1.acquire())
+        self.assertTrue(lock2.acquire())
+        lock1.release()
+        lock2.release()
+
+    def test_distributed_lock_zeroTimeoutAcquisition(self):
+        """Test lock acquisition with zero timeout."""
+        lock = get_deployment_lock(self.zookeeper_client, acquire_timeout=0)
+        self.assertTrue(lock.acquire())
```
farshidz, line 90: Shouldn't this fail due to timeout?

tests/core/index_management/test_index_management.py
```diff
@@ -276,3 +279,85 @@ def test_get_marqo_version_v20_successful(self):
             )
 
             self.assertEqual(self.index_management.get_marqo_version(), '2.0')
+
+    def test_createAndDeleteIndexCannotBeConcurrent(self):
+        """Test to ensure create_index and delete_index is not concurrent"""
```
farshidz: Description doesn't make  sense. What does it mean that they're not concurrent?

tests/core/index_management/test_index_management.py
```diff
@@ -276,3 +279,85 @@ def test_get_marqo_version_v20_successful(self):
...
+        t_1.join()
+        self.index_management.delete_index_by_name(index_name_1)
+
+    def test_createIndexFailIfNoZookeeperProvided(self):
+        self.index_management = IndexManagement(self.vespa_client, zookeeper_client=None)
+        index_name = 'a' + str(uuid.uuid4()).replace('-', '')
+        marqo_index_request = self.structured_marqo_index_request(
+            name=index_name,
+            model=Model(name='ViT-B/32'),
+            distance_metric=DistanceMetric.PrenormalizedAngular,
```
farshidz: Why do you need all this custom index config for this test? Seems irrelevant to the test case

tests/core/index_management/test_index_management.py
```diff
@@ -276,3 +279,85 @@ def test_get_marqo_version_v20_successful(self):
...
+        """Test to ensure if no Zookeeper client is provided, deployment lock is None
+        """
+        self.index_management = IndexManagement(self.vespa_client, zookeeper_client=None)
+        self.assertIsNone(self.index_management._zookeeper_deployment_lock)
+
+    def test_deploymentLock(self):
+        """Test to ensure if Zookeeper client is provided, deployment lock is not None"""
+        self.assertIsInstance(self.index_management._zookeeper_deployment_lock, ZookeeperDistributedLock)
+        lock = self.index_management._zookeeper_deployment_lock
+        self.assertEqual(lock._lock.path, '/marqo__deployment_lock')
```
farshidz: Lock path is implementation details. No need to test that. Just test that the locking works as expected and leave those details to the impelementation

farshidz (changes requested):

src/marqo/core/distributed_lock/distributed_lock_manager.py
wanliAlex: done

wanliAlex: done

src/marqo/core/distributed_lock/zookeeper_distributed_lock.py
wanliAlex: It happens when the same lock is trying to acquire twice. It will not raise a LockTimeout error but just return a False

wanliAlex: done

src/marqo/core/distributed_lock/zookeeper_distributed_lock.py
wanliAlex: done

src/marqo/core/exceptions.py
wanliAlex: I make it `OperationConflictError`

src/marqo/core/index_management/index_management.py
wanliAlex: Add an `InternalError` in `core.exceptions`

wanliAlex: Add an `InternalError` in `core.exceptions`

src/marqo/core/index_management/index_management.py
wanliAlex: Done. I add a 1-second wait though

src/marqo/core/index_management/index_management.py
wanliAlex: Done

src/marqo/vespa/vespa_client.py
wanliAlex: wait for 1 second

tests/core/distributed_lock/test_distributed_lock.py
wanliAlex: No, it will still try to acquire without waiting

tests/core/index_management/test_index_management.py
wanliAlex: Update the doc string

tests/core/index_management/test_index_management.py
wanliAlex: use an unstructured index with defaults to make it easier

tests/core/index_management/test_index_management.py
wanliAlex: done

src/marqo/core/distributed_lock/zookeeper_distributed_lock.py
wanliAlex: Users will be able to see this message and the error will be a 500. I make it more generic by saying `Marqo cannot connect to backend concurrent manager`

farshidz (dismissed):

farshidz (approved):

PR #813, opened 2024-04-20
Configurable search timeout

vicilliar:
* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Feature

* **What is the current behavior?** (You can also link to an open issue here)
No configurable search timeout

* **What is the new behavior (if this is a feature change)?**
Configurable search timeout. New env var introduced: `VESPA_SEARCH_TIMEOUT_MS`. Defaults to 1000ms.

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
No

* **Have unit tests been run against this PR?** (Has there also been any additional testing?)
In progress

* **Related Python client changes** (link commit/PR here)

* **Related documentation changes** (link commit/PR here)

* **Other information**:

* **Please check if the PR fulfills these requirements**
- [ ] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes/features)
- [ ] Docs have been added / updated (for bug fixes / features)

PR #843, opened 2024-05-22
Configurable search timeout (#813)

farshidz:
Applying configurable search timeout feature to 2.2. See [original PR](https://github.com/marqo-ai/marqo/pull/813)


src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -295,7 +295,7 @@ def _to_vespa_tensor_query(self, marqo_query: MarqoTensorQuery) -> Dict[str, Any
 
         if not marqo_query.approximate:
             query['ranking.softtimeout.enable'] = False
-            query['timeout'] = '300s'
+            query['timeout'] = 300 * 1000  # 5 minutes
```
wanliAlex, line 298: FYI you change the value type from `str` to `int`. If the test can pass, then it's fine.

src/marqo/vespa/vespa_client.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, current_generation: int, wanted_generation: int, converged: b
             self.converged = converged
 
     def __init__(self, config_url: str, document_url: str, query_url: str,
-                 content_cluster_name: str, pool_size: int = 10):
+                 content_cluster_name: str, default_search_timeout_ms: int = 1000, pool_size: int = 10):
```
wanliAlex, line 40: doc string args is missing

src/marqo/vespa/vespa_client.py
```diff
@@ -142,7 +143,7 @@ def wait_for_application_convergence(self, timeout: int = 120) -> None:
         raise VespaError(f"Vespa application did not converge within {timeout} seconds")
 
     def query(self, yql: str, hits: int = 10, ranking: str = None, model_restrict: str = None,
-              query_features: Dict[str, Any] = None, **kwargs) -> QueryResult:
+              query_features: Dict[str, Any] = None, timeout: float = None, **kwargs) -> QueryResult:
```
wanliAlex: this timeout is `float` while others are `int`

wanliAlex (changes requested):

pandu-k (approved):

src/marqo/core/structured_vespa_index/structured_vespa_index.py
farshidz: The client is making a string with `ms` suffix, that's why

src/marqo/vespa/vespa_client.py
farshidz: Need to address this in mainline

wanliAlex (approved):

src/marqo/vespa/vespa_client.py
farshidz: Addressed

PR #884, opened 2024-07-02
Enable Lexical score modifiers and change lexical from weakAND to OR

RaynorChavez:
Feature

- score modifiers are not supported in lexical search
- `weakAND` is used for constructing in `or_phrases` in constructing lexical search terms. This is for accellerated OR like search https://docs.vespa.ai/en/using-wand-with-vespa.html 

- Enables support for score modifiers when conducting lexical search
- Use `OR` instead of `weakAND` in `or_phrases` in constructing lexical search terms. This is so that all documents are matched against the query during search.

Yes

None

None

* **Other information**:

### DIFF
--- a/src/marqo/core/structured_vespa_index/structured_vespa_index.py
+++ b/src/marqo/core/structured_vespa_index/structured_vespa_index.py
@@ -462,9 +462,9 @@ class StructuredVespaIndex(VespaIndex):
         lexical_term = self._get_lexical_search_term(marqo_query) if fields_to_search else "False"
         filter_term = self._get_filter_term(marqo_query)
         if filter_term:
-            filter_term = f' AND {filter_term}'
+            search_term = f'({lexical_term}) AND ({filter_term})'
         else:
-            filter_term = ''
+            search_term = f'({lexical_term})'
 
         select_attributes = self._get_select_attributes(marqo_query)
         summary = common.SUMMARY_ALL_VECTOR if marqo_query.expose_facets else common.SUMMARY_ALL_NON_VECTOR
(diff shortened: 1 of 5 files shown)


src/marqo/version.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "2.9.0"
+__version__ = "2.10.0"
```
papa99do: Should this be changed during the release process?

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -480,7 +480,7 @@ def _to_vespa_lexical_query(self, marqo_query: MarqoLexicalQuery) -> Dict[str, A
             query_inputs.update(score_modifiers)
 
         query = {
-            'yql': f'select {select_attributes} from {self._marqo_index.schema_name} where {lexical_term}{filter_term}',
+            'yql': f'select {select_attributes} from {self._marqo_index.schema_name} where {search_term}',
```
papa99do, line 483: The change here is to wrap lexical_term and filter_term in parentheses. Is this a bug fix? Is there test cases to cover this?

farshidz: I think no parentheses wouldn't have been an issue with weakAnd because you would get weakAnd(... OR ..., ... OR ...) AND filter

farshidz (approved):

src/marqo/version.py
RaynorChavez: I think it would be better if we changed the marqo version during development to the feature's corresponding marqo version. 

Is there a better way to do this?

papa99do: Do we use this info in any of the tests?

papa99do (approved):

src/marqo/version.py
RaynorChavez: for 2.9.0 we used the version for tests, for 2.10.0 @vicilliar might be

PR #905, opened 2024-07-15
fix invalid image error message and return False if not image

RaynorChavez:
Error message change

When an invalid image is processed, marqo gives an error: `input str of `{image_path}` is not a local file or a valid url`
which might confuse users.

The error message is changed to:
```
input str of `{image_path}` is not a local file or a valid url. 
If you are using Marqo Cloud, please note that images can only be downloaded "
from a URL and local files are not supported. "
If you are running Marqo in a Docker container, you will need to use a Docker "
volume so that your container can access host files."
```
for clarity

No

* **Other information**:

### DIFF
--- a/src/marqo/s2_inference/clip_utils.py
+++ b/src/marqo/s2_inference/clip_utils.py
@@ -116,11 +116,13 @@ def load_image_from_path(image_path: str, image_download_headers: dict, timeout_
             if metrics_obj is not None:
                 metrics_obj.stop(f"image_download.{image_path}")
     else:
-        raise UnidentifiedImageError(f"input str of `{image_path}` is not a local file or a valid url. "
+        raise UnidentifiedImageError(f"Input str of `{image_path}` is not a local file or a valid url. "
                                      f"If you are using Marqo Cloud, please note that images can only be downloaded "
                                      f"from a URL and local files are not supported. "
                                      f"If you are running Marqo in a Docker container, you will need to use a Docker "
-                                     f"volume so that your container can access host files.")
+                                     f"volume so that your container can access host files. "
+                                     f"For more information, please refer to: "
+                                     f"https://docs.marqo.ai/latest/Guides/Advanced-Usage/images/")
 
     return img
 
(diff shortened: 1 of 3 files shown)


wanliAlex (changes requested):
Can you add a test for this (ensure we return bool instead of None)? Maybe also add this link https://docs.marqo.ai/2.10/Guides/Advanced-Usage/images/ in the error message

wanliAlex (dismissed):

src/marqo/s2_inference/clip_utils.py
```diff
@@ -116,7 +116,13 @@ def load_image_from_path(image_path: str, image_download_headers: dict, timeout_
...
                 metrics_obj.stop(f"image_download.{image_path}")
     else:
-        raise UnidentifiedImageError(f"input str of `{image_path}` is not a local file or a valid url")
+        raise UnidentifiedImageError(f"Input str of {image_path} is not a local file or a valid url. "
+                                     f"If you are using Marqo Cloud, please note that images can only be downloaded "
+                                     f"from a URL and local files are not supported. "
+                                     f"If you are running Marqo in a Docker container, you will need to use a Docker "
+                                     f"volume so that your container can access host files. "
+                                     f"For more information, please refer to: "
+                                     f"https://docs.marqo.ai/latest/Guides/Advanced-Usage/images/")
```
farshidz: Use `marqo_docs` module so link it tested and has the right version

tests/s2_inference/test_clip_utils.py
```diff
@@ -56,6 +56,43 @@ def test_downloadImageFromRrlCloseCalled(self, MockCurl):
         # Check if c.close() was called
         mock_curl_instance.close.assert_called_once()
 
+    def test_is_image(self):
+        # Test with valid image file extensions
+        self.assertTrue(clip_utils._is_image('image.jpg'))
```
farshidz: Use subtests so in one run we know which ones passed and which ones failed

farshidz (changes requested):

src/marqo/s2_inference/clip_utils.py
RaynorChavez: addressed

tests/s2_inference/test_clip_utils.py
RaynorChavez: addressed

farshidz (approved):

PR #908, opened 2024-07-16
Patch Pycurl, add url encoding

wanliAlex:
Bug fix

When an image URL consists of non-ASCII characters, a 500 error is raised. We replaced the `requests` package with `pycurl` in the 2.8 release. `pycurl` does not encode the URL and raises an error. This is a regression and we need to fix it.

The URL will be encoded the same as how the `requests` package does. No 500 error will be returned.

* **Other information**:


src/marqo/s2_inference/clip_utils.py
```diff
@@ -139,10 +139,15 @@ def download_image_from_url(image_path: str, image_download_headers: dict, timeo
     if not isinstance(timeout_ms, int):
         raise InternalError(f"timeout must be an integer but received {timeout_ms} of type {type(timeout_ms)}")
 
+    try:
+        encoded_url = requote_uri(image_path)
```
papa99do: Is it possible that image_path is already encoded? How do we tell?

wanliAlex: This function passes the given URL through an unquote/quote cycle to ensure that it is fully and consistently quoted. There are unit tests on these cases (URL with encoded characters, URL with mixed encoded and unencoded characters)

src/marqo/s2_inference/clip_utils.py
```diff
@@ -160,6 +165,26 @@ def download_image_from_url(image_path: str, image_download_headers: dict, timeo
     return buffer
 
 
+def encode_url(url: str) -> str:
```
farshidz, line 175: This isn't used except in tests?

wanliAlex: solved

farshidz (approved):

PR #910, opened 2024-07-18
Remove unused config

farshidz:
Remove config variables from Marqo 1 that aren't used or referenced anywhere.


src/marqo/tensor_search/enums.py
```diff
@@ -63,11 +62,9 @@ class EnvVars:
     MARQO_ROOT_PATH = "MARQO_ROOT_PATH"
     MARQO_MAX_CPU_MODEL_MEMORY = "MARQO_MAX_CPU_MODEL_MEMORY"
     MARQO_MAX_CUDA_MODEL_MEMORY = "MARQO_MAX_CUDA_MODEL_MEMORY"
-    MARQO_EF_CONSTRUCTION_MAX_VALUE = "MARQO_EF_CONSTRUCTION_MAX_VALUE"
```
papa99do, line 73: Should the corresponding env var in config.py also deleted? same to `MARQO_MAX_NUMBER_OF_REPLICAS`

farshidz: Addressed

papa99do (approved):

PR #911, opened 2024-07-19
Update Marqo Document API Response Format

wanliAlex:
Bug fix

When we feed a batch of documents to Vespa, if there is an http error (e.g., time out error) for one document, the request will raise a 500 error without the status of each document. This could be a transient network issue that only happens to one document while other documents are indexed into Vespa. We modify the error message logic to better demonstrate the indexing results.

In this PR, we update the Marqo document API format. Each document will have its individual response status for batch requests (update documents, add documents, get_documents_by_ids). For example, if a single document fails due to a transient network issue, we still return 200 for the whole batch response, but 500 for the specific document.

We also include the failure count, error count, and success count in the returned response's header for SLA purpose.

no

* **Other information**:


src/marqo/vespa/models/__init__.py
```diff
@@ -1,5 +1,5 @@
 from .delete_document_response import DeleteDocumentResponse
-from .feed_response import FeedBatchDocumentResponse, FeedBatchResponse, FeedDocumentResponse
+from .feed_response import FeedDocumentResponse, FeedBatchResponse, FeedDocumentResponse
```
farshidz: Repeated import, one import removed

src/marqo/vespa/models/feed_response.py
```diff
@@ -4,17 +4,14 @@
 
 
 class FeedDocumentResponse(BaseModel):
-    path_id: str = Field(alias='pathId')
-    id: str
-
-
-class FeedBatchDocumentResponse(BaseModel):
```
farshidz, line 11: Why are we merging these two?

src/marqo/vespa/vespa_client.py
```diff
@@ -237,7 +237,7 @@ def feed_document(self, document: VespaDocument, schema: str, timeout: int = 60)
 
         self._raise_for_status(resp)
```
farshidz: You're still going to fail here so if it's not 200 it won't reach the next line?

src/marqo/vespa/vespa_client.py
```diff
@@ -675,57 +675,88 @@ async def _update_document_async(self, semaphore: asyncio.Semaphore, async_clien
...
-        try:
-            # This will cover 200 and document-specific errors. Other unexpected errors will be raised.
-            return UpdateDocumentResponse(**resp.json(), status=resp.status_code)
-        except JSONDecodeError:
-            if resp.status_code == 200:
-                # A 200 response shouldn't reach here
-                raise VespaError(f'Unexpected response: {resp.text}')
-
-            self._raise_for_status(resp)
+                error_message = (f"Marqo has encountered an error while updating this document to Vespa and "
```
farshidz: We don't name Vespa in error messages. We refer to vector store

src/marqo/vespa/vespa_client.py
```diff
@@ -675,57 +675,88 @@ async def _update_document_async(self, semaphore: asyncio.Semaphore, async_clien
...
+                                 f"the document is not updated. Original error: {e}. "
+                                 f"This might be a transient network issue and you can "
+                                 f"retry the operation. If the issue persists, please contact Marqo support")
```
farshidz: Contact Marqo support probably shouldn't be here because they could be open source users

farshidz: I would opt for a short message here and let the caller have a detailed message. The client is very low-level and doesn't have to have detailed messages suitable for the user

src/marqo/vespa/vespa_client.py
```diff
@@ -675,57 +675,88 @@ async def _update_document_async(self, semaphore: asyncio.Semaphore, async_clien
...
+                return UpdateDocumentResponse(status=500, message=error_message, pathId=error_doc_path_id, id=doc_id)
+            except JSONDecodeError as e:
+                if resp.status_code == 200:
+                    # A 200 response shouldn't reach here
+                    raise ValueError(f"Marqo received an unexpected response from Vespa: {resp.text} with 200 "
```
farshidz: I feel like it's okay to fail the whole batch and raise a 500 in this case because we don't really know what we're dealing with when there's a `JSONDecodeError` and it's not a network issue. Keen to get @papa99do 's review here

src/marqo/vespa/vespa_client.py
```diff
@@ -675,57 +675,88 @@ async def _update_document_async(self, semaphore: asyncio.Semaphore, async_clien
...
+                                     f"status_code. The response is not in the JSON format and Marqo can not decode "
+                                     f"it") from e
+                else:
+                    error_message = (f"Marqo has encountered an error while update this document to Vespa and the "
```
farshidz: Ditto

farshidz (changes requested):

src/marqo/vespa/vespa_client.py
wanliAlex: Raising a VespaError now. The VespaError will fail the whole batch.

src/marqo/vespa/vespa_client.py
wanliAlex: done

src/marqo/vespa/vespa_client.py
wanliAlex: done

src/marqo/vespa/vespa_client.py
wanliAlex: done

src/marqo/vespa/vespa_client.py
wanliAlex: Yes because we are feeding a single document here so we don't have to catch the error. A HTTP error in this feeding should fail the whole request

src/marqo/vespa/models/feed_response.py
wanliAlex: I think they are duplicated because even for `FeedBatchDocumentResponse`, we feed one single document in one semaphore. So the response format is the same. The only different between batch_feed and feed is how we handle the return errors: 1) raise the error directly when feeding one document and 2) catch the error and return the error message on the specific document when a batch of documents

src/marqo/vespa/models/__init__.py
wanliAlex: done

src/marqo/vespa/models/feed_response.py
wanliAlex: Reverted this change

src/marqo/core/models/marqo_add_documents_response.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class MarqoAddDocumentsResponse(MarqoBaseModel):
+    errors: bool
+    processingTimeMs: float
+    index_name: str  # TODO Change this to camelCase in the future (Breaking change!)
+    items: List[MarqoAddDocumentsItem]
+    success_count: int = Field(exclude=True, default=0)
+    error_count: int = Field(exclude=True, default=0)
+    failure_count: int = Field(exclude=True, default=0)
```
papa99do: Since this is used in 3 different places, is it worth extracting them to a new Model class? `BatchResponseCount` or `BatchResponseStats`?  It can be initialised with a list of items, which either contains status field or is a dict (which will use 200 for default). It can also have a to_html_header method that returns a dict we can put to the html header.

src/marqo/vespa/models/feed_response.py
```diff
@@ -4,17 +4,21 @@
...
     path_id: str = Field(alias='pathId')
     id: str
 
 
 class FeedBatchDocumentResponse(BaseModel):
+    """A response from feeding a document to Vespa while the
+    document is part of a batch."""
     status: int
-    path_id: str = Field(alias='pathId')
+    path_id: Optional[str] = Field(alias='pathId', default=None)
```
papa99do, line 16: Why is path_id nullable now? Is this change backward compatible?

tests/core/document/test_partial_document_update.py
```diff
@@ -201,8 +202,9 @@ def test_update_text_field_tensor(self):
             "text_field_tensor": "I can't be updated",
             "_id": "1"
         }
-        r = update_documents(body=UpdateDocumentsBodyParams(documents=[updated_doc]),
-                             index_name=self.structured_index_name, marqo_config=self.config)
+        r = self.config.document.partial_update_documents_by_index_name(
```
papa99do, line 205: I see we changed from testing the api method to 'service' method. I reckon the intention is to test the 'right' thing?
In this case, should we not include `dict(exclude_none=True, by_alias=True)` but to check the pydantic model directly? Then we might need another test to test the `.dict(exclude_none=True, by_alias=True)` part in api.py.

src/marqo/vespa/vespa_client.py
```diff
@@ -675,46 +675,77 @@ async def _update_document_async(self, semaphore: asyncio.Semaphore, async_clien
...
+
+        Note: This method is used by the async feed batch method to feed documents concurrently. Unhandled exceptions
+        will be raised in the main thread and leads a 500 error for the whole batch. Therefore, exceptions should be
+        handled gracefully in this method for the specific document. We should keep the error message as similar as the
+        Vespa error messages since this is a low level method. Overwrite the error message in higher level methods.
+
+        Exceptions that are handled in this method:
+        1. httpx.RequestError: We convert this error to a 500 error for the specific document and put 'Network Error' in
+        the message.
+        2.
```
papa99do: Missing the doc?

src/marqo/vespa/vespa_client.py
```diff
@@ -675,46 +675,77 @@ async def _update_document_async(self, semaphore: asyncio.Semaphore, async_clien
...
-                raise VespaError(e) from e
-
-        try:
-            # This will cover 200 and document-specific errors. Other unexpected errors will be raised.
-            return FeedBatchDocumentResponse(**resp.json(), status=resp.status_code)
-        except JSONDecodeError as e:
-            if resp.status_code == 200:
-                # A 200 response shouldn't reach here
-                raise VespaError(f'Unexpected response: {resp.text}') from e
+            except httpx.RequestError as e:
```
papa99do, line 737: In httpx, `HTTPError` is the base class for `RequestError` and `HTTPStatusError`. We changed to only capture `RequestError` here, `HTTPStatusError` will be raised to the outer layer. Is this intended? 
Had another look at the httpx lib, It seems that HTTPStatusError can only be raised by calling res.raise_for_status. And we have our own logic to raise for status and was handled below. So I think it should be OK to capture either `HTTPError` or `RequestError`. 
I think we should make this consistent across this file.

papa99do: It seems that `HTTPStatusError` can only be raised by calling res.raise_for_status. And we have our own logic to raise for status and was handled below. So I think it should be OK.

tests/core/document/test_partial_document_update.py
wanliAlex: Your understanding is correct. The only reason I am not doing this is that would take much more effort to revise the unit tests code. I am not sure if we have enough time to do this in this release

src/marqo/vespa/models/feed_response.py
wanliAlex: Actually, I should delete this attribute because it is not used anywhere in our code except test.

wanliAlex: An error response would not return a path_id so I make it optional. It shouldn't be an issue because this field is not used anywhere in the code base except tests.

src/marqo/core/models/marqo_add_documents_response.py
wanliAlex: Good idea

papa99do (dismissed):
LGTM

src/marqo/core/document/document.py
```diff
@@ -194,7 +194,7 @@ def translate_add_documents_response(self, responses: Optional[FeedBatchResponse
         """
 
         new_items: List[MarqoAddDocumentsItem] = []
-        # A None response means all documents are errors
+        # A None response no documents are sent to Vespa
```
papa99do: Does this mean it's a validation error?

papa99do (dismissed):

src/marqo/vespa/vespa_client.py
```diff
@@ -675,46 +675,82 @@ async def _update_document_async(self, semaphore: asyncio.Semaphore, async_clien
...
         try:
-            # This will cover 200 and document-specific errors. Other unexpected errors will be raised.
             return UpdateDocumentResponse(**resp.json(), status=resp.status_code)
-        except JSONDecodeError:
+        except JSONDecodeError as e:
             if resp.status_code == 200:
-                # A 200 response shouldn't reach here
-                raise VespaError(f'Unexpected response: {resp.text}')
+                # A 200 response shouldn't reach here, so we error out the whole batch
+                raise VespaError(cause=e, message="Unexpected response from Vespa") from e
```
farshidz: Why don't you include the text like we did before?

farshidz (approved):

PR #912, opened 2024-07-22
Hybrid search for unstructured indexes

vicilliar:
Feature

Hybrid search is not supported for unstructured indexes

Hybrid search now works on unstructured indexes.

- Searchable attributes are not supported, errors out if attempted.
- No more "modifiers" rank profiles in unstructured schemas. This affects all search, not just hybrid.

No. But hybrid unstructured will not work on indexes older than version 2.11.0

In progress

None

* **Other information**:


src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -56,13 +56,6 @@ class StructuredVespaIndex(VespaIndex):
     _MAX_LONG = 9223372036854775807
     _MIN_LONG = -9223372036854775808
 
-    _VERSION_2_9_0 = semver.VersionInfo.parse("2.9.0")
```
papa99do, line 59: Note: Moved to base class VespaIndex to reuse in UnstructuredVespaIndex

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
```diff
@@ -224,45 +229,146 @@ def _get_score_modifiers(marqo_query: MarqoQuery) -> \
                 else:
                     raise InternalError(f'Unknown score modifier type {modifier.type}')
 
-            # Note one of these could be empty, but not both
-            return {
-                unstructured_common.QUERY_INPUT_SCORE_MODIFIERS_MULT_WEIGHTS: mult_tensor,
-                unstructured_common.QUERY_INPUT_SCORE_MODIFIERS_ADD_WEIGHTS: add_tensor
-            }
+            if self._marqo_index_version < self._VERSION_2_10_0:
```
papa99do: What about unstructured index created with version 2.10.x? the new query inputs does not exists yet.

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
```diff
@@ -69,8 +69,11 @@ def _to_vespa_tensor_query(self, marqo_query: MarqoTensorQuery) -> Dict[str, Any
 
         score_modifiers = self._get_score_modifiers(marqo_query)
 
-        ranking = unstructured_common.RANK_PROFILE_EMBEDDING_SIMILARITY_MODIFIERS if score_modifiers \
-            else unstructured_common.RANK_PROFILE_EMBEDDING_SIMILARITY
+        if self._marqo_index_version < self._VERSION_2_10_0:
```
vicilliar: This might need to be 2.11

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
vicilliar: This should be 2.11.0

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
vicilliar: Actually use: MARQO_UNSTRUCTURED_HYBRID_SEARCH_MINIMUM_VERSION

src/marqo/core/search/hybrid_search.py
```diff
@@ -83,24 +83,36 @@ def search(
...
-        if isinstance(marqo_index, UnstructuredMarqoIndex):
-            raise core_exceptions.UnsupportedFeatureError(
-                "Unstructured indexes are not yet supported for hybrid search. "
-                "Please use a structured index.")
-
+        # Version checks (different for structured and unstructured)
         marqo_index_version = marqo_index.parsed_marqo_version()
-        if marqo_index_version < constants.MARQO_HYBRID_SEARCH_MINIMUM_VERSION:
+        if isinstance(marqo_index, StructuredMarqoIndex) and \
+                marqo_index_version < constants.MARQO_STRUCTURED_HYBRID_SEARCH_MINIMUM_VERSION:
```
farshidz: You're comparing a `VersionInfo` with a string. Does this work or do we have a bug in 2.10?

src/marqo/core/search/hybrid_search.py
```diff
@@ -83,24 +83,36 @@ def search(
...
+                f"This index was created with Marqo {marqo_index_version}."
             )
 
         # Use default hybrid settings if not provided
         if hybrid_parameters is None:
             hybrid_parameters = HybridParameters()
 
+        # TODO: Remove when unstructured searchable attributes are supported
+        if isinstance(marqo_index, UnstructuredMarqoIndex) and \
+                (hybrid_parameters.searchableAttributesTensor is not None or \
```
farshidz, line 110: Do we not support specifying attributes to rank by?

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -52,15 +52,27 @@ class UnstructuredVespaDocument(MarqoBaseModel):
...
+
     _VESPA_DOC_FIELDS = "fields"
     _VESPA_DOC_ID = "id"
 
     @classmethod
     def from_vespa_document(cls, document: Dict) -> "UnstructuredVespaDocument":
         """Instantiate an UnstructuredVespaDocument from a Vespa document."""
         fields = document.get(cls._VESPA_DOC_FIELDS, {})
+
+        raw_lexical_score = fields.pop(unstructured_common.VESPA_DOC_HYBRID_RAW_LEXICAL_SCORE) \
```
farshidz, line 76: Does the custom searcher add this as a document field?

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -175,8 +187,18 @@ def to_marqo_document(self, return_highlights: bool = False) -> Dict[str, Any]:
                     json.loads(serialized_multimodal_params)
 
         if return_highlights and self.fields.match_features:
-            if not self.fields.vespa_chunks:
-                raise VespaDocumentParsingError(f"Document {self.fields.marqo__id} does not have any chunks. "
-                                                f"No highlights can be extracted.")
-            marqo_document[index_constants.MARQO_DOC_HIGHLIGHTS] = self.fields.extract_highlights()
+            if self.fields.match_features[f"closest({unstructured_common.VESPA_DOC_EMBEDDINGS})"]["cells"] == {}:
```
farshidz: How did this code handle lexical before this change?

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -175,8 +187,18 @@ def to_marqo_document(self, return_highlights: bool = False) -> Dict[str, Any]:
...
+                if not self.fields.vespa_chunks:
+                    raise VespaDocumentParsingError(f"Document {self.fields.marqo__id} does not have any chunks. "
+                                                    f"No highlights can be extracted.")
+                marqo_document[index_constants.MARQO_DOC_HIGHLIGHTS] = self.fields.extract_highlights()
+
+        # Hybrid search raw scores
+        if self.raw_tensor_score is not None:
+            marqo_document[index_constants.MARQO_DOC_HYBRID_TENSOR_SCORE] = self.raw_tensor_score
+        if self.raw_lexical_score is not None:
+            marqo_document[index_constants.MARQO_DOC_HYBRID_LEXICAL_SCORE] = self.raw_lexical_score
```
farshidz, line 205: Seems a bit odd that we have score as a document field. The score coming from Vespa isn't a document field

src/marqo/core/unstructured_vespa_index/unstructured_vespa_schema.py
```diff
@@ -172,81 +164,172 @@ def _generate_unstructured_schema(cls, marqo_index: UnstructuredMarqoIndex) -> s
...
+                    summary {self._EMBEDDINGS} type tensor<float>(p{{}}, x[{dimension}]) {{}}
                 }}
             }}
             """
         )
+
+        return schema
+
+    def _generate_rank_profiles(self, marqo_index: UnstructuredMarqoIndex):
+        rank_profiles: List[str] = list()
```
farshidz: This implementation seems inconsistent with the rest of this class. Since the schema is the same for all indexes, we have almost hard-coded it (with values interpolated) which is more readable. I think it's best we keep the class consistent.

src/marqo/tensor_search/validation.py
```diff
@@ -415,7 +415,8 @@ def validate_dict(field: str, field_content: Dict, is_non_tensor_field: bool, ma
                     f"The field {field} is a map field and only supported for indexes created with Marqo 2.9.0 or later. "
                     f"See {marqo_docs.map_fields()} and {marqo_docs.mappings()}."
                 )
-
+            # TODO: It's also possible that unstructured custom vector field gets here, and will error out because of
```
farshidz: Not quite sure what this means (why should content be numeric?) but why are we not addressing it since it seems to be going to cause 500s?

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
```diff
@@ -45,12 +44,13 @@ def to_vespa_query(self, marqo_query: MarqoQuery) -> Dict[str, Any]:
                                          'You can create a structured index '
                                          'by `mq.create_index("your_index_name", type="structured")`')
 
-        if isinstance(marqo_query, MarqoTensorQuery):
+        if isinstance(marqo_query, MarqoHybridQuery):       # TODO: Rethink structure so order of checking doesn't matter
```
farshidz, line 50: Why does order of checking matter?

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
farshidz: I agree. This version checks in this class don't seem right. Should be < 2.11

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
```diff
@@ -224,45 +229,146 @@ def _get_score_modifiers(marqo_query: MarqoQuery) -> \
...
+            The dictionaries are split between 'mult' and 'add' weights.
+
+            search_type: 'lexical' or 'tensor'. It will be added to the end of the
+            dictionary key to be parsed out in custom searcher. This allows the 2 score modifiers lists to be used
+            separately in the custom searcher.
+            """
+            mult_tensor = {}
+            add_tensor = {}
+            for modifier in score_modifiers:
+                if modifier.type == ScoreModifierType.Multiply:
```
farshidz: Maybe this logic could be a separate function as it's repeated in the classic score modifiers function as well?

src/marqo/core/search/hybrid_search.py
papa99do: I checked the implementation of semver, it supports comparison with a string or a dict. This works as expected. But it would be safer to parse it to a semver. (so even we swap the side it would still work).

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
papa99do: @farshidz Because of the class hierarchy. MarqoHybridQuery inherits both TensorQuery and LexicalQuery. This is something we discussed when we first implement this class. We should revisit this design at some point to see if it makes sense to have MarqoHybridQuery inherits directly from MarqoQuery.

tests/tensor_search/integ_tests/test_dict_score_modifiers.py
```diff
@@ -21,7 +21,6 @@ def setUpClass(cls) -> None:
         # UNSTRUCTURED indexes
         unstructured_default_text_index = cls.unstructured_marqo_index_request(
             model=Model(name='random/small'),
-            marqo_version="2.9.0"
```
papa99do, line 24: why do we remove the marqo_version here? Is this test case for old score modifiers?

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
```diff
@@ -224,45 +229,146 @@ def _get_score_modifiers(marqo_query: MarqoQuery) -> \
...
-                or_terms = ''
-
-            if marqo_query.and_phrases:
-                and_terms = ' AND '.join([
-                    f'default contains "{phrase}"' for phrase in marqo_query.and_phrases
-                ])
-                if or_terms:
-                    or_terms = f'({or_terms})'
-                    and_terms = f' AND ({and_terms})'
+    def _get_hybrid_score_modifiers(self, hybrid_query: MarqoHybridQuery) -> \
```
papa99do: I found a bit of duplication here. Same as the next method. If the logic is exactly same for structured and unstructured. Is it worth extracting them to the base class?

src/marqo/core/search/hybrid_search.py
vicilliar: We don't yet, it's still a fast follower.

vicilliar: Also, this whole functionality is blocked off until we support searchable attributes with unstructured (2.12)

src/marqo/core/unstructured_vespa_index/unstructured_document.py
vicilliar: yes. and it already does this for the structured implementation.

src/marqo/core/search/hybrid_search.py
vicilliar: done

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
vicilliar: fixed

src/marqo/core/unstructured_vespa_index/unstructured_document.py
vicilliar: There would be no `self.fields.match_features` in that case, because it's only in the embedding similarity rank profile. This issue arises because there's a possibility now for a doc with match features (but it's empty) and empty vespa chunks.

src/marqo/core/unstructured_vespa_index/unstructured_document.py
vicilliar: The alternative is to add it as a field in unstructured_schema, but that doesn't seem right either.

tests/tensor_search/integ_tests/test_dict_score_modifiers.py
vicilliar: I think having that there was an oversight. unstructured_default_text_index should always be the latest version. @RaynorChavez Is there anything wrong with me removing this one line?

src/marqo/tensor_search/validation.py
vicilliar: This is because custom vector field is a dict content type, but if no `mappings` is given, we will get here, which assumes it is a dict score modifiers type, which needs all dict values to be numeric.

tests/tensor_search/integ_tests/test_dict_score_modifiers.py
RaynorChavez: @vicilliar, we can safely remove this. This was an undeleted holdover from a previous version of the test. marqo_version by default is set to be the latest version in `unstructured_default_text_index`.

src/marqo/core/unstructured_vespa_index/unstructured_vespa_schema.py
vicilliar: I'm following the structured_vespa_schema pattern for this one. I found it easier to modularize when we add things line by line. Do you want to follow that approach?

vicilliar: Nevermind, fixed to match the rest of the class.

tests/tensor_search/integ_tests/test_dict_score_modifiers.py
vicilliar: Awesome, thanks!

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
vicilliar: fixed

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
vicilliar: I'll extract these to the base class: `_convert_score_modifiers_to_tensors`, `_get_score_modifiers`, `_get_hybrid_score_modifiers`.

src/marqo/core/unstructured_vespa_index/unstructured_document.py
farshidz, line 76: This class is a mix of class methods and instance methods. Can we get rid of class methods?

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
farshidz: You can do `type(marqo_query) is ...` for a strict type check

src/marqo/core/unstructured_vespa_index/unstructured_document.py
farshidz: As agreed move this into `self.fields.extract_highlights`

src/marqo/core/unstructured_vespa_index/unstructured_document.py
vicilliar: Sounds good, we can move this refactor to another PR? Might take a bit of extra time.

papa99do: These from_xxx factory methods should be class methods.

tests/tensor_search/integ_tests/test_hybrid_search.py
```diff
@@ -1274,49 +1620,77 @@ def test_hybrid_search_none_query_with_context_vectors_passes(self):
...
+                      str(e.exception))
 
+    def test_hybrid_parameters_with_wrong_search_method_fails(self):
+        """
+        Test that hybrid parameters with wrong search method fails.
+        """
+
+        # TODO: Use api.search() instead of tensor_search.search()
+        # Covered in API tests
+        pass
```
papa99do, line 1696: nit. This file is super large now. For better maintenance, consider extracting common methods or break it down to multiple files. Does not need to be done in this PR.

src/marqo/core/structured_vespa_index/structured_vespa_schema.py
```diff
@@ -275,10 +276,10 @@ def _generate_rank_profiles(self, marqo_index: StructuredMarqoIndex) -> List[str
             rank_profiles.append(f'query({field.embeddings_field_name}): 0')
         rank_profiles.append(f'query({common.QUERY_INPUT_EMBEDDING}) tensor<float>(x[{model_dim}])')
         rank_profiles.append(f'query({common.QUERY_INPUT_BM25_AGGREGATOR}): 0')
-        rank_profiles.append(f'query({common.QUERY_INPUT_SCORE_MODIFIERS_MULT_WEIGHTS_LEXICAL}) tensor<double>(p{{}})')
-        rank_profiles.append(f'query({common.QUERY_INPUT_SCORE_MODIFIERS_ADD_WEIGHTS_LEXICAL}) tensor<double>(p{{}})')
-        rank_profiles.append(f'query({common.QUERY_INPUT_SCORE_MODIFIERS_MULT_WEIGHTS_TENSOR}) tensor<double>(p{{}})')
-        rank_profiles.append(f'query({common.QUERY_INPUT_SCORE_MODIFIERS_ADD_WEIGHTS_TENSOR}) tensor<double>(p{{}})')
+        rank_profiles.append(f'query({constants.QUERY_INPUT_SCORE_MODIFIERS_MULT_WEIGHTS_LEXICAL}) tensor<double>(p{{}})')
```
papa99do, line 279: nit. rank profiles are quite similar for two types of index now. Should we consider extract some common logic out to the base class? Or at least using the same `textwrap.dedent` approach for better readability? Can be done in a separate PR

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
```diff
@@ -210,59 +216,46 @@ def tree_to_filter_string(node: search_filter.Node) -> str:
...
-                    add_tensor[modifier.field] = modifier.weight
-                else:
-                    raise InternalError(f'Unknown score modifier type {modifier.type}')
-
-            # Note one of these could be empty, but not both
-            return {
-                unstructured_common.QUERY_INPUT_SCORE_MODIFIERS_MULT_WEIGHTS: mult_tensor,
-                unstructured_common.QUERY_INPUT_SCORE_MODIFIERS_ADD_WEIGHTS: add_tensor
-            }
+    def _get_lexical_search_term(self, marqo_query: MarqoLexicalQuery) -> str:
```
papa99do, line 219: Is this method changed? Is this just the change of order and some comments?

tests/tensor_search/integ_tests/test_hybrid_search.py
vicilliar: Noted, this is probably best done by extracting common logic, maybe in a future refactor.

src/marqo/core/structured_vespa_index/structured_vespa_schema.py
vicilliar: @farshidz Mentioned that it's better to use `textwrap.dedent` for unstructured, since indexes will have the same rank profiles, but for structured, we use single line `append` statements to customize whether lexical/tensor/hybrid rank profiles are added.

src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
vicilliar: I just extracted it, since it was under `_to_vespa_lexical_query` before, but now we need it for hybrid. Also added a check to determine whether to use `marqo_query.hybrid_parameters.scoreModifiersLexical` or `marqo_query.score_modifiers`

papa99do (dismissed):

papa99do (approved):

farshidz (approved):

PR #914, opened 2024-07-23
Fix bug: max limit and offset ignored

RaynorChavez:
Bug fix

Marqo ignores when user sets the environment variables `MARQO_MAX_SEARCH_LIMIT` `MARQO_MAX_SEARCH_OFFSET` and uses the hardcoded values 1,000 and 10,000 respectively

The values from the two variables above, when set by the user are used by marqo in the `<query-profile>`

None

Yes

* **Other information**:


src/marqo/vespa/vespa_client.py
```diff
@@ -40,7 +40,7 @@ def __init__(self, current_generation: int, wanted_generation: int, converged: b
     def __init__(self, config_url: str, document_url: str, query_url: str,
                  content_cluster_name: str, default_search_timeout_ms: int = 1000, pool_size: int = 10,
                  feed_pool_size: int = 10, get_pool_size: int = 10, delete_pool_size: int = 10,
-                 partial_update_pool_size: int = 10):
+                 partial_update_pool_size: int = 10, max_search_offset: int = 10000, max_search_limit: int = 1000):
```
vicilliar: Where did we get 10000 and 1000 as the limits?

tests/tensor_search/test_api.py
```diff
@@ -199,6 +200,75 @@ def test_search_timeout_short_timer_fails(self):
...
+
+                        res = self.client.post("/indexes/" + index.name + "/search?device=cpu", json={
+                            "q": "irrelevant",
+                            "searchMethod": "TENSOR",
+                            "limit": custom_limit + 1,
+                            "offset": custom_offset + 1,
+                        })
+                        self.assertEqual(res.status_code, 400)
+
+            with self.subTest(search_method="HYBRID"):
```
vicilliar: you can execute this in a loop `for search_method in ["TENSOR", "HYBRID"]:` so you don't repeat the code

src/marqo/vespa/vespa_client.py
RaynorChavez: I got them from here: https://github.com/marqo-ai/marqo/blob/mainline/src/marqo/api/configs.py

Do you think we should refer to them or just make these two required?

tests/tensor_search/test_api.py
RaynorChavez: Good point! Will refactor in the morning

RaynorChavez: Does Hybrid support Unstructured now though?

RaynorChavez: test removed

src/marqo/core/index_management/index_management.py
```diff
@@ -450,14 +457,35 @@ def _add_default_query_profile(self, app: str) -> bool:
...
+        new_query_profile = self._DEFAULT_QUERY_PROFILE_TEMPLATE.format(
+            max_hits=self._max_search_limit,
+            max_offset=self._max_search_offset
+        )
+
+        if os.path.exists(profile_path):
+            with open(profile_path, 'r') as f:
+                existing_content = f.read()
+            if existing_content == new_query_profile:
+                return False
```
vicilliar: shouldn't we have an elif branch here to actually write to the profile path file if it does not match the new query profile?

src/marqo/core/index_management/index_management.py
```diff
@@ -96,12 +98,17 @@ def bootstrap_vespa(self) -> bool:
         app = self.vespa_client.download_application()
         configured = self._marqo_config_exists(app)
 
+        query_profile_updated = self._update_query_profile(app)
```
vicilliar: Just to clarify, when we deploy the vespa application, we're always reading from `search/query-profiles/default.xml`? Do we not also want to call `self._save_marqo_version(version.get_version())`

src/marqo/core/index_management/index_management.py
```diff
@@ -96,12 +98,17 @@ def bootstrap_vespa(self) -> bool:
...
+
         if not configured:
             self._add_marqo_config(app)
             self.vespa_client.deploy_application(app)
             self.vespa_client.wait_for_application_convergence()
             self._save_marqo_version(version.get_version())
             return True
+        elif query_profile_updated:
+            self.vespa_client.deploy_application(app)
+            self.vespa_client.wait_for_application_convergence()
```
vicilliar: We might want to return True here as well, since we redeployed the application. What are the implications of this?

vicilliar (dismissed):

src/marqo/core/index_management/index_management.py
RaynorChavez: I think the following block covers this

```os.makedirs(os.path.dirname(profile_path), exist_ok=True)
        with open(profile_path, 'w') as f:
            f.write(new_query_profile)```

File: src/marqo/core/index_management/index_management.py (hunk 4 above)
Comment (RaynorChavez) [reply src/marqo/core/index_management/index_management.py]: It's my understanding the the _save_marqo_version is persisted after the initial configuration in the if block? I could be wrong

File: src/marqo/core/index_management/index_management.py (hunk 5 above)
Comment (RaynorChavez) [reply src/marqo/core/index_management/index_management.py]: ah yes I missed this. we should return True

File: src/marqo/core/index_management/index_management.py [hunk 6]
```diff
@@ -58,8 +58,8 @@ class IndexManagement:
     _DEFAULT_QUERY_PROFILE_TEMPLATE = textwrap.dedent(
         '''
         <query-profile id="default">
-            <field name="maxHits">1000</field>
-            <field name="maxOffset">10000</field>
+            <field name="maxHits">{max_hits}</field>
```
papa99do: My understanding about our decision on Friday is that we'll need to hard code this config to a extremely large value so we don't get limited by Vespa during search. 
Then we add a validator to the SearchQuery to make sure `limit <= MARQO_MAX_SEARCH_LIMIT` and `offset <= MARQO_MAX_SEARCH_OFFSET`, the default value of these two env vars can be 1000 and 10000 respectively to keep the behaviour same as before. And they can be overridden as needed. 
@farshidz , @RaynorChavez , is my understanding correct?

RaynorChavez: @papa99do, wouldn't this entail asking vespa to return a large number of results each query?

papa99do: I don't think so. We'll need to make sure the limit and offset parameter in each query does not exceed the limit set in env vars (or the default limit if env vars are not provided).

tests/tensor_search/test_api.py
```diff
@@ -62,6 +63,54 @@ def test_memory_disabled_403(self):
             response = self.client.get("/memory")
             self.assertEqual(response.status_code, 403)
 
+    def test_custom_search_limit(self):
+        """
+        Test that the search endpoint returns the expected search limit when MARQO_MAX_SEARCH_LIMIT is set.
+        """
+        custom_limits = [2000, 1000000]
+        for custom_limit in custom_limits:
+            with patch.dict('os.environ', {
```
papa99do, line 72: What if these two env vars are not set? I think we still want to have a default max value, right? which is 1000 and 10000 respectively. And if user want larger Max limit, they can provide it by setting the env var.

RaynorChavez: Hey @papa99do, I think these default values are already set in config.py in api, right?

https://github.com/marqo-ai/marqo/blob/mainline/src/marqo/api/configs.py

RaynorChavez: Or were you referring to making a test without setting these env vars explicitly?

papa99do: Got it. Looks good.

papa99do (approved):

PR #921, opened 2024-07-28
Manage index settings in Vespa application package

papa99do:
Reliability improvements

- Index settings are maintained in a Vespa schema, which has consistency issue
- Marqo config is stored as a document in Vespa
- Vespa Bootstrapping is done once-off

- Index settings are maintained within the Vespa application package
- Marqo config is stored within the Vespa application package
- Bootstrapping of the Vespa app is done automatcially on every Marqo release
- Provided a way to safely rollback to the previous version of Marqo

No, there are no changes to the external APIs or customer-facing behaviour of Marqo

Yes

N/A

N/A

* **Other information**:


requirements.dev.txt
```diff
@@ -61,3 +61,6 @@ httpx==0.25.0
 # test requirements
 pyvespa==0.37.1
 pytest==7.4.3
+pytest-cov==5.0.0
+pytest-profiling==1.7.0
+setuptools==69.5.1
```
papa99do: This is required by clip.py when running with tox.

src/marqo/core/index_management/index_management.py
```diff
@@ -314,322 +197,121 @@ def get_all_indexes(self) -> List[MarqoIndex]:
...
-        for document in documents_element.findall("document"):
-            if document.get("type") == name:
-                documents_element.remove(document)
+        should_deploy = yield app
 
-                if deleted:
-                    logger.warn(f"Multiple schemas with name {name} found in services.xml, removing all")
+        if should_deploy is None or should_deploy is True:
+            prepare_response = self.vespa_client.prepare(prepare_url)
+            # TODO handle prepare configChangeActions
```
papa99do, line 299: This `configChangeActions` field in the response from Vespa contains a list of actions (possibly empty) that must be performed in order to apply some config changes between the current active application and this next prepared application. hese actions are organized into three categories; restart, reindex, and refeed See https://docs.vespa.ai/en/reference/deploy-rest-api-v2.html#prepare-session for an example.  

What should we do when this list is not empty? 
- For requests modifying schemas (in the future), if we get any action in the list, should we reject the request?
- For bootstrap and rollback (in the future if they use deployment session after the Vespa bugfix), should we allow restart? 
@farshidz, what's your thought on this?

src/marqo/core/index_management/vespa_application_package.py
```diff
@@ -0,0 +1,644 @@
...
+        self._config_index_setting_components()
+
+    def _cleanup_container_config(self):
+        """
+        Components config needs to be in sync with the components in the jar files. This method cleans up the
+        custom components config, so we can always start fresh. This assumes that the container section of the
+        services.xml file only has `node` config and empty `document-api` and `search` elements initially. Please
+        note that any manual config change in container section will be overwritten.
+        """
+        container_element = self._ensure_only_one('container')
```
papa99do, line 87: Our approach of fresh bootstrapping components config of container is based on following assumptions:
1. There's only one container element in the services.xml file. 
2. All configs in the container element except the nodes section is fully controlled by Marqo. 

These assumption might not be true in following cases:
- non-HA to HA migration in Marqo Cloud. A transiting state will have two container elements. In this case, we'd like to fail the bootstrapping process anyways to avoid entering any weird state. Also Marqo Cloud will make sure the upgrade of Marqo is blocked when processing migration
- OS customers could have multiple container clusters defined. (one use case is the separate the index and search traffic). Do we need to support this?
- OS customers could have other configuration in container section. should we keep those? 

If we have to support the cases above, I'll need to change the fresh bootstrapping process to an incremental one, which will make the logic more complicated. @farshidz , what's your thought?

src/marqo/core/index_management/vespa_application_package.py
```diff
@@ -0,0 +1,644 @@
...
+        app_sem_version = semver.VersionInfo.parse(app_version, optional_minor_and_patch=True)
+
+        return app_sem_version < marqo_sem_version
+
+    def bootstrap(self, marqo_version: str, existing_index_settings: List[MarqoIndex] = ()) -> None:
+        logger.info(f'Bootstrapping Vespa app to {marqo_version}')
+        # Migrate existing index settings from previous versions of Marqo
+        if not self.is_configured:
+            if existing_index_settings is not None and len(existing_index_settings) > 0:
+                logger.info(f'Importing existing index settings {[index.json for index in existing_index_settings]}')
```
papa99do: @farshidz, Do we have any security concerns of logging out the existing index settings?

src/marqo/core/index_management/vespa_application_package.py
```diff
@@ -0,0 +1,644 @@
...
+            f'''
+            <validation-overrides>
+                 <allow until='{datetime.utcnow().strftime('%Y-%m-%d')}'>schema-removal</allow>
+            </validation-overrides>
+            '''
+        ).strip()
+        self._store.save_file(content, 'validation-overrides.xml')
+
+    def _validate_services_xml_for_rollback(self, services_xml_backup: str) -> None:
+        services_xml_old = ServiceXml(services_xml_backup)
```
papa99do: During the rollback, if we detect the changes in 'content/documents' or any 'nodes' elements in services.xml, we will abort the rollback. This is to avoid rolling back to an inconsistent state. Should we capture any diffs of the services.xml file in the log for debugging purpose? @farshidz

tests/core/index_management/existing_vespa_app/search/query-profiles/default.xml
```diff
@@ -0,0 +1,4 @@
+<query-profile id="default">
+    <field name="maxHits">1000</field>
```
RaynorChavez, line 2: should this be 1,000,000?

papa99do: This is used for testing. I used a different value to test if the bootstrapping process overrides this value

src/marqo/tensor_search/api.py
```diff
@@ -503,6 +504,16 @@ def rollback_marqo(req: RollbackRequest, marqo_config: config.Config = Depends(g
     rollback_runner.rollback(from_version=req.from_version, to_version=req.to_version)
 
 
+@app.post("/rollback-vespa")
+@utils.enable_upgrade_api()
```
papa99do: remove this flag

src/marqo/tensor_search/on_start_script.py
```diff
@@ -65,6 +65,7 @@ def run(self):
                 "ensure that Marqo is configured properly for this. See "
                 f"{marqo_docs.configuring_marqo()} for more details. Error: {e}"
             )
+            raise e
```
papa99do, line 68: This change is requested by Cloud team. If anything goes wrong during bootstrapping, Marqo instance will be shutdown. This guarantees that Marqo will always run on a properly bootstrapped Vespa application. Marqo Cloud can leverage this to capture and manage upgrade failure properly as well.

src/marqo/core/index_management/index_management.py
```diff
@@ -1,71 +1,43 @@
...
 
 logger = marqo.logging.get_logger(__name__)
 
 
-class _MarqoConfig(ImmutableStrictBaseModel):
-    version: str
-
-
 class IndexManagement:
+    MINIMUM_VESPA_VERSION_TO_SUPPORT_UPLOAD_BINARY_FILES = semver.VersionInfo.parse('8.382.22')
```
farshidz: Does this need to be public?

src/marqo/core/index_management/index_management.py
```diff
@@ -75,35 +47,66 @@ def __init__(self, vespa_client: VespaClient, zookeeper_client: Optional[Zookeep
...
+        self._deployment_timeout_seconds = deployment_timeout_seconds
+        self._convergence_timeout_seconds = convergence_timeout_seconds
 
-    def _instantiate_deployment_lock(self) -> Optional[ZookeeperDistributedLock]:
-        """Instantiate a ZookeeperDistributedLock."""
-        if self._zookeeper_client is None:
-            return None
-        else:
-            return get_deployment_lock(self._zookeeper_client)
+    @staticmethod
```
farshidz: Does this need to be static? I tend to avoid one random static method just because it doesn't need any instance variables, to keep things consistent. I could be wrong though

src/marqo/core/index_management/index_management.py
```diff
@@ -75,35 +47,66 @@ def __init__(self, vespa_client: VespaClient, zookeeper_client: Optional[Zookeep
...
 
         Returns:
-            True if Vespa was configured, False if it was already configured
+            True if Vespa was bootstrapped, False if it was already up-to-date
         """
-        app = self.vespa_client.download_application()
-        configured = self._marqo_config_exists(app)
+        with self._vespa_deployment_lock():
+            application = self._get_application_context_manager_for_bootstrapping_and_rollback(check_configured=False)
+            with application as app:
```
farshidz: Is it safer if you inline this or is there no difference?

src/marqo/core/index_management/index_management.py
```diff
@@ -75,35 +47,66 @@ def __init__(self, vespa_client: VespaClient, zookeeper_client: Optional[Zookeep
...
+
+                marqo_version = version.get_version()
+                has_marqo_settings_schema = app.has_schema(self._MARQO_SETTINGS_SCHEMA_NAME)
+                marqo_config_doc = self._get_marqo_config() if has_marqo_settings_schema else None
+
+                if not app.need_bootstrapping(marqo_version, marqo_config_doc):
```
farshidz: Add some debug messages to this process. They've proven very useful in the past

src/marqo/core/index_management/index_management.py
```diff
@@ -75,35 +47,66 @@ def __init__(self, vespa_client: VespaClient, zookeeper_client: Optional[Zookeep
...
         """
-        app = self.vespa_client.download_application()
-        configured = self._marqo_config_exists(app)
+        with self._vespa_deployment_lock():
+            application = self._get_application_context_manager_for_bootstrapping_and_rollback(check_configured=False)
+            with application as app:
+
+                marqo_version = version.get_version()
+                has_marqo_settings_schema = app.has_schema(self._MARQO_SETTINGS_SCHEMA_NAME)
+                marqo_config_doc = self._get_marqo_config() if has_marqo_settings_schema else None
```
farshidz: Is it possible the schema exists but the confid doesn't? I assume this isn't atomic

src/marqo/core/index_management/index_management.py
```diff
@@ -75,35 +47,66 @@ def __init__(self, vespa_client: VespaClient, zookeeper_client: Optional[Zookeep
...
+
+                if not app.need_bootstrapping(marqo_version, marqo_config_doc):
+                    application.gen.send(False)  # tell context manager to skip deployment
```
farshidz: Why application and not app? Why do we need both? I'm a bit confused

src/marqo/core/index_management/index_management.py
```diff
@@ -75,35 +47,66 @@ def __init__(self, vespa_client: VespaClient, zookeeper_client: Optional[Zookeep
...
-        return False
+    def rollback_vespa(self) -> None:
+        with self._vespa_deployment_lock():
+            application = self._get_application_context_manager_for_bootstrapping_and_rollback()
+            with application as app:
+                try:
+                    app.rollback(version.get_version())
+                except ApplicationRollbackError as e:
+                    logger.error(e.message)
+                    application.gen.send(False)  # tell context manager to skip deployment
```
farshidz: What happens if this call fails or we fail before this call? Also does the context manager not suffice?

src/marqo/core/index_management/index_management.py
```diff
@@ -122,58 +125,7 @@ def create_index(self, marqo_index_request: MarqoIndexRequest) -> MarqoIndex:
...
-            api_exceptions.InvalidArgError: If several settings are invalid in the context of the Marqo API.
-            Check the errors in to_marqo_index_request method of IndexSettings model for more details.
-
-        Returns:
-            None: If the validation is successful, nothing is returned,
-            else InvalidArgumentError is raised.
-        """
-        index_settings = IndexSettings(**settings_dict)
-        index_settings.to_marqo_index_request(index_name)
+        return self.batch_create_indexes([marqo_index_request])[0]
```
farshidz, line 123: Nice

src/marqo/core/index_management/index_management.py
papa99do: I need to access the context manager to send result (whether we need to deploy) so I cannot inline here.

src/marqo/core/index_management/index_management.py
papa99do: no, will change to private. A topic for future: a lot of version check has been added to our code base, does it make sense to extract all of them to one class?

src/marqo/core/index_management/index_management.py
papa99do: actually, I didn't touch this method, and I think it's added by Ali for Cloud usage only. I can change it to instance method or class method.

src/marqo/core/index_management/index_management.py
papa99do: I've added a lot of info level log to vespa_application_package.py for this process, I think they should be enough for debugging bootstrapping and rollback issues. Since we don't often do bootstrapping and rollback, I think info level log should be fine. Please also have a look at the vespa_application_package.py

src/marqo/core/index_management/index_management.py
papa99do: So do you mean the marqo__settings schema exists, but there's no marqo_config doc in it? This will raise a 404 error for now. I think this only happens if the previous version isn't bootstrapped properly, or in Marqo 2.0? 

What could be a solution for this? should we just use None if we got a 404?

src/marqo/core/index_management/index_management.py
papa99do: application is the context manager, I'll rename it.

src/marqo/core/index_management/index_management.py
papa99do: This call tells context manager that we need to skip the deployment due to some failures. You made a good point that this call could fail. But only in two circumstances: 
- the context manager yields all the values already, and raises StopIteration error, which should not happen here since we only use the yielded value once. 
- any exception type when handling the sent value, in this case, it's just a bool value, it should be fine.

src/marqo/core/index_management/index_management.py
```diff
@@ -75,35 +47,67 @@ def __init__(self, vespa_client: VespaClient, zookeeper_client: Optional[Zookeep
...
 
-        if not configured:
-            self._add_marqo_config(app)
-            self.vespa_client.deploy_application(app)
-            self.vespa_client.wait_for_application_convergence()
-            self._save_marqo_version(version.get_version())
-            return True
+                marqo_version = version.get_version()
+                has_marqo_settings_schema = app.has_schema(self._MARQO_SETTINGS_SCHEMA_NAME)
+                marqo_config_doc = self._get_marqo_config() if has_marqo_settings_schema else None
```
farshidz: Note `has_marqo_settings_schema` will be False in 2.0, exists only from 2.1

src/marqo/core/index_management/index_management.py
```diff
@@ -314,322 +207,129 @@ def get_all_indexes(self) -> List[MarqoIndex]:
 
         return [
             MarqoIndex.parse_raw(document.fields['settings'])
-
             for document in batch_response.documents
             if not document.id.split('::')[-1].startswith(constants.MARQO_RESERVED_PREFIX)
         ]
 
-    def get_index(self, index_name) -> MarqoIndex:
+    def _get_marqo_config(self) -> Optional[MarqoConfig]:
```
farshidz: Perhaps the name should make it clear this is for older versions and not current

src/marqo/core/index_management/index_management.py
```diff
@@ -314,322 +207,129 @@ def get_all_indexes(self) -> List[MarqoIndex]:
...
+        This method is only used during upgrade and rollback
+        TODO check if this is still needed
         """
-        Create the default query profile if it does not exist.
-        Args:
-            app: Path to Vespa application package
+        application = self._vespa_application_with_deployment_session()
+        with application as app:
+            marqo_version = app.get_marqo_config().version
+            application.gen.send(False)  # do not deploy
```
farshidz: This seems error prone. Should we have a different context manager when we don't need to deploy? Do we need a context manager?

src/marqo/core/index_management/index_management.py
```diff
@@ -314,322 +207,129 @@ def get_all_indexes(self) -> List[MarqoIndex]:
...
+            return marqo_version
 
-        Returns:
-            True if query profile was created, False if it already existed
+    @contextmanager
+    def _vespa_application(self, check_configured: bool = True):
```
farshidz: Type annotation missing

src/marqo/core/index_management/index_management.py
```diff
@@ -314,322 +207,129 @@ def get_all_indexes(self) -> List[MarqoIndex]:
...
+        if should_deploy is not None:
+            yield
 
-        documents_elements = root.findall("content/documents")
+    @contextmanager
+    def _vespa_application_with_deployment_session(self, check_configured: bool = True):
+        """
+        A context manager that handles the deployment of a Vespa application package in a deployment session
+        This is a recommended way to deploy a Vespa application package. It leverages the optimistic locking mechanism
+        to avoid race conditions. Changes to the application package that do not touch any binary files should use
```
farshidz: Statement doesn't apply to the latest Vespa version

src/marqo/core/index_management/vespa_application_package.py
```diff
@@ -0,0 +1,636 @@
...
+            if backup is not None:
+                backup.backup_file(self.read_binary_file(*paths), *paths)
+            os.remove(path)
+
+
+class ApplicationPackageDeploymentSessionStore(VespaApplicationStore):
+    def __init__(self, content_base_url: str, vespa_client: VespaClient):
+        self._content_base_url = content_base_url
+        self._vespa_client = vespa_client
+        self._all_contents = vespa_client.list_contents(content_base_url)
```
farshidz: I/O operation in constructor is discouraged

src/marqo/core/index_management/vespa_application_package.py
```diff
@@ -0,0 +1,636 @@
...
+        marqo_config = MarqoConfigStore(old_backup.read_text_file(self._MARQO_CONFIG_FILE)).get()
+        if marqo_config is not None and marqo_config.version != marqo_version:
+            raise ApplicationRollbackError(f"Cannot rollback to {marqo_config.version}, current Marqo version is {marqo_version}")
+
+        # Copy components jar files from the old Marqo version
+        self._copy_components_jar()
+
+        new_backup = VespaAppBackup()
+        for paths in old_backup.files_to_remove():
+            logger.info(f'Removing file {paths}')
```
farshidz: These are too verbose for debug

src/marqo/core/index_management/vespa_application_package.py
```diff
@@ -0,0 +1,636 @@
...
+        The version of the application is retrieved in the following order:
+        - from the 'marqo_config.json' file (Post v2.12.0)
+        - from the marqo_config_doc passed in which is marqo__config doc saved in marqo__settings schema (Post v2.1.0)
+        - if neither is available, return 2.0.0 as default (Pre v2.1.0 or not bootstrapped yet)
+        """
+        logger.info(f'Marqo version is {marqo_version}')
+
+        if self.is_configured and self.get_marqo_config() is not None:
+            app_version = self.get_marqo_config().version
+            logger.info(f'Vespa app version is detected in marqo_config.json: {app_version}')
```
farshidz: Vector store in info messages. Also might be too verbose for info

src/marqo/core/index_management/index_management.py
farshidz: What I mean is if bootstrapping ran partially and was interrupted. We need to be able to recover from this. e.g. user killed the image

src/marqo/tensor_search/on_start_script.py
farshidz: Can we change the wording of the error to "Failed to bootstrap Marqo"?

src/marqo/core/index_management/vespa_application_package.py
farshidz: > All configs in the container element except the nodes section is fully controlled by Marqo.

This is fine but needs to be in Marqo docs, and also cloud migration guide

src/marqo/core/index_management/index_management.py
farshidz: Confirmed the process is deployed atomically

src/marqo/core/index_management/vespa_application_package.py
papa99do: will add it in the configuration.md file, and cloud migration guide

src/marqo/tensor_search/on_start_script.py
papa99do: It makes sense to me to mention vector store here since we are bootstrapping the vector store to work with current version of Marqo.

farshidz (approved):

PR #922, opened 2024-07-30
Hybrid search lexical/tensor score modifiers fix

vicilliar:
Bug fix

Currently, when using hybrid search with LEXICAL retrieval method and TENSOR ranking method, using `scoreModifiersLexical` will raise a 400. This is a bug.

Now both `scoreModifiersLexical` and `scoreModifiersTensor` are available and function for HYBRID search with LEXICAL retrieval method and TENSOR ranking.

`ranking.rerankCount` was used to limit the number of results from phase 1 (lexical ranking) into phase 2.

no

yes

* **Other information**:


src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -544,7 +554,8 @@ def _to_vespa_hybrid_query(self, marqo_query: MarqoHybridQuery) -> Dict[str, Any
             'searchChain': 'marqo',
             'yql': 'PLACEHOLDER. WILL NOT BE USED IN HYBRID SEARCH.',
             'ranking': common.RANK_PROFILE_HYBRID_CUSTOM_SEARCHER,
-
+            'ranking.rerankCount': marqo_query.limit,       # limits the number of results going to phase 2
```
papa99do, line 557: From Vespa doc: `ranking.rerankCount` specifies the number of hits that should be ranked in the second ranking phase. So my understanding is that past first ranking phase, only the top N result will be reranked by phase 2 ranker. By default N is 100, but can be overridden by `rerankCount`.  This might impact relevance of the result if the first phase ranker does not perform well and we rely on the second phase to give us more accurate result.
Since we are not using expensive second-phase ranking, I think it make sense to use the default value unless limit is larger than the default value. @farshidz , keen to know your thought on this.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -329,16 +328,29 @@ Query createSubQuery(
...
         Tensor fieldsToRank = extractTensorRankFeature(query, featureNameFieldsToRank);
         Iterator<Cell> cells = fieldsToRank.cellIterator();
-        cells.forEachRemaining(
-                (cell) ->
-                        addFieldToRankFeatures(
-                                cell, queryNew, retrievalMethod, rankingMethod, verbose));
+        cells.forEachRemaining((cell) -> addFieldToRankFeatures(cell, queryNew, verbose));
+
+        // Extract using RANKING method (second-phase)
+        if (!(retrievalMethod.equals(rankingMethod))) {
```
papa99do, line 343: Can we have a Java unit test case for this change?

src/marqo/core/models/hybrid_parameters.py
```diff
@@ -69,9 +69,11 @@ def validate_properties(cls, values):
...
-            if values.get('rankingMethod') not in [RankingMethod.Lexical, RankingMethod.RRF]:
+            if not (values.get('rankingMethod') in [RankingMethod.Lexical, RankingMethod.RRF] or
+                    values.get('retrievalMethod') == RetrievalMethod.Lexical):
                 raise ValueError(
-                    "'scoreModifiersLexical' can only be defined for 'lexical', 'rrf' ranking methods")  # TODO: re-add normalize_linear
+                    "'scoreModifiersLexical' can only be defined for 'lexical', 'rrf' ranking methods or "
+                    "'lexical' retrieval method.")  # TODO: re-add normalize_linear
 
         # score_modifiers_tensor can only be defined for Tensor, RRF, NormalizeLinear
         if values.get('scoreModifiersTensor') is not None:
```
papa99do, line 79: So just to confirm, the scoreModifiersTensor will not be applied to the first ranking phase of the tensor search, right?

src/marqo/core/structured_vespa_index/structured_vespa_index.py
vicilliar: After talking to the vespa team, `ranking.rerankCount` will not shave off any results, but only pass a certain number to 2nd phase for ranking. All results not passed to 2nd phase will still be in the final result, just with a low score, effectively letting them be dropped at the end.

Regarding relevance, this is the behavior we want right? We only want the top N results out of phase 1 to be considered in phase 2.

src/marqo/core/models/hybrid_parameters.py
vicilliar: scoreModifiersTensor will be used in:
1. lexical/tensor search, in second phase.
2. tensor/lexical search, in first phase.

src/marqo/core/structured_vespa_index/structured_vespa_index.py
papa99do: Let me use an example to elaborate my point. Let's say user has only one shard in Vespa. User queries with limit set to 10. The query is executed in that shard, retrieved 100 docs, first phase ranking ranked all 100 docs. Let's assume the first phase ranking does not do a very good job, some docs with high relevance falls out of top 10 results. We expect second phase ranking to improve the result. 
With the default setting, all 100 docs will be included in the second phase ranking. But with 'ranking.rerankCount' set to 10, only the top 10 docs will be included in the second phase ranking, which means we do not have a chance to rectify other docs' scores, leading to lower relevance in the returned result.

src/marqo/core/models/hybrid_parameters.py
papa99do: So we'll need to add the same logic as `scoreModifiersLexical`, right?  I mean we should also allow this score modifier when `values.get('retrievalMethod') == RetrievalMethod.Tensor`

vicilliar: No, since tensor/lexical is single phase (only lexical ranking)

src/marqo/core/structured_vespa_index/structured_vespa_index.py
papa99do, line 557: should we rerank `marqo_query.limit+marqo_query.offset` docs?

papa99do (approved):

PR #931, opened 2024-08-12
Add multimodal model support

RaynorChavez:
Feature

We can only index images and text

Implementing support video-audio-image-text models for indexing and search.

No

Yes

https://github.com/marqo-ai/py-marqo/pull/249

https://github.com/marqo-ai/marqodocs/pull/304

* **Other information**:


src/marqo/tensor_search/add_docs.py
```diff
@@ -61,59 +87,202 @@ def threaded_download_and_preprocess_images(allocated_docs: List[dict], image_re
...
+            chunks.append(process_chunk(chunk, modality))
+            memory_pool.return_chunk(chunk)
+            buffer.seek(0)
+            buffer.truncate()
+
+    c = pycurl.Curl()
+    c.setopt(c.URL, url)
+    c.setopt(c.WRITEFUNCTION, write_function)
+    for key, value in headers.items():
+        c.setopt(c.HTTPHEADER, [f"{key}: {value}"])
```
papa99do: have you tested if headers will be added? This is what we did before: `c.setopt(pycurl.HTTPHEADER, [f"{k}: {v}" for k, v in headers.items()])`

src/marqo/tensor_search/add_docs.py
```diff
@@ -61,59 +87,202 @@ def threaded_download_and_preprocess_images(allocated_docs: List[dict], image_re
...
+    c.perform()
+    c.close()
+
+    if buffer.tell() > 0:
+        chunk = memory_pool.get_chunk()
+        chunk[:buffer.tell()] = buffer.getvalue()
+        chunks.append(process_chunk(chunk[:buffer.tell()], modality))
```
papa99do: We limit the size of downloaded content, but not the chunks. How large can these chunks be for a batch containing about 10 videos? Would there by any concerns if content repo takes too much memory?

src/marqo/tensor_search/add_docs.py
```diff
@@ -61,59 +87,202 @@ def threaded_download_and_preprocess_images(allocated_docs: List[dict], image_re
...
+    buffer = BytesIO()
+
+    def write_function(data):
+        buffer.write(data)
+        if buffer.tell() >= memory_pool.chunk_size:
+            chunk = memory_pool.get_chunk()
+            buffer.seek(0)
+            chunk[:len(buffer.getvalue())] = buffer.getvalue()
+            chunks.append(process_chunk(chunk, modality))
+            memory_pool.return_chunk(chunk)
```
papa99do: chunk needs to be returned to memory pool even there's an exception.

.github/workflows/largemodel_unit_test_CI.yml
```diff
@@ -62,6 +62,12 @@ jobs:
           repository: marqo-ai/marqo-base
           path: marqo-base
 
+      - name: Install FFmpeg
```
farshidz: What alternatives did we consider for this package? Why did we pick this?

requirements.txt
```diff
@@ -12,13 +11,79 @@ urllib3==1.26.0
...
+pynvml==11.5.0
 kazoo==2.10.0
 pycurl==7.45.3
 certifi==2019.11.28
-transformers==4.41.2
-optimum==1.20.0
\ No newline at end of file
+transformers==4.31.0 # Consider 4.31.0 for _expand_mask
+
+# s2_inference:
```
farshidz: Where did all these new requirements come from?

src/marqo/core/unstructured_vespa_index/unstructured_vespa_schema.py
```diff
@@ -54,13 +54,16 @@ def _generate_unstructured_marqo_index(self, schema_name: str) -> UnstructuredMa
...
+            video_preprocessing=self._index_request.video_preprocessing,
+            audio_preprocessing=self._index_request.audio_preprocessing,
             distance_metric=self._index_request.distance_metric,
             vector_numeric_type=self._index_request.vector_numeric_type,
             hnsw_config=self._index_request.hnsw_config,
             marqo_version=self._index_request.marqo_version,
             created_at=self._index_request.created_at,
             updated_at=self._index_request.updated_at,
             treat_urls_and_pointers_as_images=self._index_request.treat_urls_and_pointers_as_images,
+            treat_urls_and_pointers_as_media=self._index_request.treat_urls_and_pointers_as_media,
```
farshidz, line 66: Does this supersede `treat_urls_and_pointers_as_images` and deprecate it? Otherwise having both will be a weird interface

src/marqo/s2_inference/languagebind/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+    'video': LanguageBindVideo,
+    'audio': LanguageBindAudio
+}
+transform_dict = {
+    'video': LanguageBindVideoProcessor,
+    'audio': LanguageBindAudioProcessor,
+    'image': LanguageBindImageProcessor,
+}
+
+class LanguageBind(nn.Module):
```
farshidz, line 33: Is this test code? Otherwise init file isn't the right place

src/marqo/s2_inference/languagebind/audio/configuration_audio.py
```diff
@@ -0,0 +1,430 @@
+import copy
```
farshidz, line 1: Is there any commonality between this and ImageBind that can be used to have one abstraction for both?

requirements.txt
RaynorChavez: @farshidz I had to copy the entire `.dev.txt` into the main `.txt` file since a bunch of requirements installed in the image are not compatible with languagebind. So this is temporary until we can sort out the ones that need to change in docker.

In addition to the ones already in `.dev.txt` I got the new requirements for languagebind from the languagebind repo https://github.com/PKU-YuanGroup/LanguageBind/blob/main/requirements.txt 

Had to upgrade/downgrade some existing ones to support languagebind. I would need your and @papa99do 's opinion on these changes esp on the existing ones that were changed.

src/marqo/core/unstructured_vespa_index/unstructured_vespa_schema.py
RaynorChavez: Yeah it does. I couldn't remove `treat_urls_and_pointers_as_image` since it would cause backwards compatibility issues, but using only `treat_urls_and_pointers_as_media` for video and audio links and `treat_urls_and_pointers_as_image` for images would be a bit confusing.

So I decided that setting either to `True` would put all urls through the process of identifying whether it is an image, audio, video file. Effectively, `treat_urls_and_pointers_as_image` isn't needed anymore, but just kept for backwards compatibility. 

What are your thoughts on this?

src/marqo/s2_inference/languagebind/__init__.py
RaynorChavez: This isn't test code. These are pretty important for using LanguageBind (loading, file processing, inference). The `/languagebind/` folder was copied from their repo, so I didn't modify anything there. 

Do you think we should move these?

src/marqo/s2_inference/languagebind/audio/configuration_audio.py
RaynorChavez: We will have to look into imagebind's code as well to determine this. I would imagine there are commonalities that we can use, since I think they both just build on the same CLIP architecture.

.github/workflows/largemodel_unit_test_CI.yml
RaynorChavez: Initially, I only downloaded chunks of the file through pycurl. I then realized that downloading chunks do not make them viable video/audio files. So we needed to find a video processing library. There are a few options

1. GStreamer - similar to ffmpeg, highly performant, can stream, more customizable with plugins, but more complex to setup. Has multiple python bindings. Good for complex multimedia tasks.
2. PyAV - Python binding for ffmpeg. Roughly near native ffmpeg performance. Might be easier to use in the future for when we change multithreading logic, but with the current threading logic (default 20 threads, configurable), we won't be able to take advantage of its built in automatic memory management. (I also need to study how it manages multithreading and async more). I discovered this after I started using the ffmpeg-python binding.
3. ffmpeg - Well maintained, has high performance (I think people often decide between using ffmpeg and gstreamer), also has multiple python bindings (aside from PyAV). 
4. OpenCV - Has video processing capabilities, but not great for audio processing
5. VLC Python - seems like it isn't as fast as either GStreamer or ffmpeg for our purposes.

https://news.ycombinator.com/item?id=24883756

So, given that we only need to download chunks of a video or audio file, I don't see the need to use GStreamer and all the complexities that come with it. Since ffmpeg is also just about as performant and is simpler to use, this is what I think is best to use right now.

RaynorChavez:
Once review is done, I'll remove all the print debug statements in this PR.

.gitignore
```diff
@@ -151,4 +151,5 @@ dump.rdb
 # Tester app for unit tests
 scripts/vespa_local/vespa_tester_app.zip
 
-src/marqo/tensor_search/cache_dir/*
\ No newline at end of file
+src/marqo/tensor_search/cache_dir/*
+tests/tensor_search/integ_tests/cache_dir/*
```
RaynorChavez: Will be removed. I've set the cache dir to the same location other models are saved to.

.github/workflows/largemodel_unit_test_CI.yml
papa99do: Action item for later: find a way to run these tests inside the marqo-base image so that we don't need to install these deps in each build

requirements.txt
```diff
@@ -12,13 +11,20 @@ urllib3==1.26.0
...
+
+# LanguageBind
+
+eva-decord==0.6.1
+einops==0.6.1
+pytorchvideo==0.1.5
+torchaudio==0.12.1
+SoundFile==0.12.1
+python-magic==0.4.27
+ffmpeg-python==0.2.0
```
papa99do: Can be removed once the base image has these deps

src/marqo/s2_inference/model_registry.py
```diff
@@ -2008,7 +2058,9 @@ def _get_model_load_mappings() -> Dict:
             "fp16_clip": FP16_CLIP,
             'random':Random,
             'hf':HF_MODEL,
-            "no_model": NO_MODEL}
+            "no_model": NO_MODEL,
+            #"languagebind": LANGUAGEBIND,
```
papa99do: why is this commented out?

tests/tensor_search/integ_tests/test_search_combined.py
```diff
@@ -172,6 +198,73 @@ def tearDown(self) -> None:
         super().tearDown()
         self.device_patcher.stop()
 
+    @pytest.mark.skipif(torch.cuda.is_available() is True, reason="We skip this test if we have cuda support. This model is 5gb and is very slow on g4dn.xlarge and may crash it")
+    def test_search_video(self):
+        documents = [
+            {"video_field_1": "https://marqo-k400-video-test-dataset.s3.amazonaws.com/videos/---QUuC4vJs_000084_000094.mp4", "_id": "1"},
+            # Replace the audio link with something marqo-hosted
+            {"audio_field_1": "https://audio-previews.elements.envatousercontent.com/files/187680354/preview.mp3", "_id": "2"},
```
papa99do: better to upload the file to our public s3 bucket just in case envato moves or deletes this file in the future.

RaynorChavez: Yes, I'll still need to replace this audio link with one from Li's dataset

src/marqo/s2_inference/model_registry.py
RaynorChavez: Because currently we use different loading logic for Multimodal models https://github.com/marqo-ai/marqo/blob/450fea3adc74929f3c64a18375ca17b9e0ef2afd/src/marqo/s2_inference/s2_inference.py#L551

src/marqo/s2_inference/model_registry.py
```diff
@@ -1932,6 +1932,56 @@ def _get_onnx_clip_properties() -> Dict:
...
+        'LanguageBind/Audio_FT': {
+            "name": "LanguageBind/Audio_FT",
+            "dimensions": 768,
+            "type": "languagebind",
+            "loader": "languagebind",
+            "supported_modalities": ["video", "language"],
+            "video_chunk_length": 20,
+            "audio_chunk_length": 20,
+        },
+        'LanguageBind/Video_1.5_FT': {
```
RaynorChavez: Change name to `LanguageBind/Video_1.5_FT'` once memory tests are done

requirements.dev.txt
```diff
@@ -61,3 +61,14 @@ httpx==0.25.0
...
+
+# LanguageBind
+
+eva-decord==0.6.1
+einops==0.6.1
+pytorchvideo==0.1.5
+torchaudio==0.12.1
+SoundFile==0.12.1
+python-magic==0.4.27
+ffmpeg-python==0.2.0
```
papa99do, line 73: These can be removed, only eva-decord is needed (for local test)

requirements.txt
```diff
@@ -14,11 +13,20 @@ httpx==0.25.0
...
+optimum==1.20.0
+
+#eva-decord==0.6.1
+decord==0.6.0
+einops==0.6.1
+pytorchvideo==0.1.5
+torchaudio==0.12.1
+SoundFile==0.12.1
+python-magic==0.4.27
+ffmpeg-python==0.2.0
```
papa99do: Only these dep are needed in this file:

pydantic==1.10.11
httpx==0.25.0
semver==3.0.2
scipy==1.10.1
memory-profiler==0.61.0
cachetools==5.3.1
pynvml==11.5.0 # For cuda utilization
readerwriterlock==1.0.9
kazoo==2.10.0
pycurl==7.45.3

requirements.dev.txt
RaynorChavez: Right, but the ones in `requirements.txt` right, not in this `.dev.txt` one? We still need all these other libraries for local installation

requirements.txt
RaynorChavez: Thanks, changed!

src/marqo/core/unstructured_vespa_index/unstructured_vespa_schema.py
RaynorChavez: Just a note on this, expected behavior:

1. When both are False: all content is treated as text.
2. When treat_urls_and_pointers_as_images is True and treat_urls_and_pointers_as_media is False: will attempt to process URLs and pointers as images, but not as other types of media (like video or audio).
3. When treat_urls_and_pointers_as_images is False and treat_urls_and_pointers_as_media is True: will attempt to process URLs and pointers as various types of media, including images, videos, and audio.
4. When both are True: This case is effectively the same as when only treat_urls_and_pointers_as_media is True. will attempt to process URLs and pointers as various types of media, including images, videos, and audio (same as the third one).

src/marqo/s2_inference/languagebind/__init__.py
RaynorChavez: Update, I modified some things in their code so we can reduce our dependencies

src/marqo/api/configs.py
```diff
@@ -29,6 +29,7 @@ def default_env_vars() -> dict:
         EnvVars.MARQO_THREAD_EXPIRY_TIME: 1800,  # 30 minutes
         EnvVars.MARQO_ENABLE_THROTTLING: "TRUE",
         EnvVars.MARQO_LOG_LEVEL: "info",
+        EnvVars.MARQO_MEDIA_DOWNLOAD_THREAD_COUNT: 20,
```
wanliAlex: Be more specific here. This is the download thread per request

src/marqo/s2_inference/multimodal_utils.py
```diff
@@ -0,0 +1,286 @@
...
+            buffer.write(chunk)
+            if buffer.tell() >= sample_size:
+                break
+        buffer.seek(0)
+        yield buffer
+    finally:
+        buffer.close()
+        response.close()
+
+def infer_modality(content: Union[str, List[str], bytes]) -> Modality:
```
farshidz: This needs to be unit tested

farshidz: Public methods need docstring

src/marqo/s2_inference/multimodal_utils.py
```diff
@@ -0,0 +1,286 @@
...
+            try:
+                with fetch_content_sample(content) as sample:
+                    #print(f"did not find extension, but found url: {content}")
+                    mime = magic.from_buffer(sample.read(), mime=True)
+                    #print(f"mime: {mime}")
+                    if mime.startswith('image/'):
+                        return Modality.IMAGE
+                    elif mime.startswith('video/'):
+                        return Modality.VIDEO
+                    elif mime.startswith('audio/'):
```
farshidz: Else it's text or an error? This is returnign text currently

src/marqo/core/models/marqo_index.py
```diff
@@ -250,6 +259,8 @@ class MarqoIndex(ImmutableBaseModel, ABC):
     normalize_embeddings: bool
     text_preprocessing: TextPreProcessing
     image_preprocessing: ImagePreProcessing
+    video_preprocessing: Optional[VideoPreProcessing] = None
```
wanliAlex, line 262: Can this be `None`? we don't have optional for image and text

src/marqo/s2_inference/multimodal_utils.py
```diff
@@ -0,0 +1,286 @@
...
+        if mime.startswith('image/'):
+            return Modality.IMAGE
+        elif mime.startswith('video/'):
+            return Modality.VIDEO
+        elif mime.startswith('audio/'):
+            return Modality.AUDIO
+        else:
+            return Modality.TEXT
+    
+    return None
```
farshidz: Return type isn't optional. Also how is the caller supposed to interpret None? Maybe you need to raise an exception?

src/marqo/core/models/marqo_index.py
```diff
@@ -324,6 +335,7 @@ def _cache_or_get(self, key: str, func):
 class UnstructuredMarqoIndex(MarqoIndex):
     type = IndexType.Unstructured
     treat_urls_and_pointers_as_images: bool
+    treat_urls_and_pointers_as_media: Optional[bool] = None
```
wanliAlex, line 338: Same, can this be `None`?

src/marqo/s2_inference/languagebind/audio/modeling_audio.py
```diff
@@ -0,0 +1,1012 @@
...
+        hidden_states = self.mlp(hidden_states)
+        hidden_states = residual + hidden_states
+
+        outputs = (hidden_states,)
+
+        if output_attentions:
+            outputs += (attn_weights,)
+
+        return outputs
+
```
farshidz, line 157: Please use your IDE to format all code

src/marqo/core/unstructured_vespa_index/unstructured_vespa_schema.py
wanliAlex: It makes sense to me

src/marqo/tensor_search/tensor_search.py
```diff
@@ -144,19 +146,26 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
...
+    content_repo = {}
 
     text_chunk_prefix = marqo_index.model.get_text_chunk_prefix(add_docs_params.text_chunk_prefix)
 
     docs, doc_ids = config.document.remove_duplicated_documents(add_docs_params.docs)
 
+    # Check if model is Video/Audio. If so, manually set thread_count to 5
+    media_download_thread_count = add_docs_params.image_download_thread_count
+    if marqo_index.model.get_properties().get('type') in [ModelType.LanguageBind]:
+        #if media_download_thread_count > 5:
```
farshidz: What is this?

wanliAlex (changes requested):

src/marqo/s2_inference/multimodal_utils.py
wanliAlex: You can move this to a separate Python file called `multimdal_download.py` and get rid of the `utils`. As Farshid mentioned before, having a utils file is not the best practice.

src/marqo/s2_inference/multimodal_utils.py
```diff
@@ -0,0 +1,286 @@
...
+            self.clip_type = {
+                'audio': 'LanguageBind_Audio_FT',
+            }
+        elif self.model_name == "LanguageBind/Video_V1.5_FT":
+            self.clip_type = {
+                'video': 'LanguageBind_Video_V1.5_FT',
+            }
+        else:
+            raise ValueError(f"Unsupported LanguageBind model: {self.model_name}")
+        model = LanguageBind(clip_type=self.clip_type, cache_dir=ModelCache.languagebind_cache_path)
```
wanliAlex: nit: `model = LanguageBind(clip_type=self.clip_type, cache_dir=ModelCache.languagebind_cache_path).to(self.device)`

src/marqo/s2_inference/s2_inference.py
```diff
@@ -40,33 +44,11 @@
...
-        enable_cache: A flag to enable or disable the cache
-
-    Returns:
-        List[List[float]]: _description_
-
-    Raises:
-        VectoriseError: if the content can't be vectorised, for some reason.
-    """
-
+              model_auth: ModelAuth = None, enable_cache: bool = False, modality: Modality = None, **kwargs,) -> List[List[float]]:
```
wanliAlex: Should the default Modality be `Text` instead of `None`?

src/marqo/tensor_search/add_docs.py
```diff
@@ -6,27 +6,43 @@
...
+
+def threaded_download_and_preprocess_content(allocated_docs: List[dict], 
+                                                content_repo: dict, 
+                                                tensor_fields: List[str],
+                                                image_download_headers: dict,
+                                                device: str = None,
+                                                media_field_types_mapping: Optional[Dict[str, FieldType]] = None,
+                                                download_headers: Optional[Dict] = None,  # Optional for now
+                                                metric_obj: Optional[RequestMetrics] = None,
+                                                preprocessors: Optional[Dict[str, Compose]] = None,
```
wanliAlex: Can we make the preprocessors a Pydantic class and add it to the abstraction?

src/marqo/tensor_search/add_docs.py
```diff
@@ -64,120 +80,186 @@ def threaded_download_and_preprocess_images(
...
-                    except PIL.UnidentifiedImageError as e:
-                        image_repo[doc[field]] = e
-                        metric_obj.increment_counter(f"{doc.get(field, '')}.UnidentifiedImageError")
-                        continue
-                    # preprocess image to tensor
-                    if preprocessor:
-                        if not device:
-                            raise ValueError("Device must be provided for preprocessing images")
+                if isinstance(doc[field], str) or force_download:
+                    modality = infer_modality(doc[field])
```
wanliAlex, line 91: What if the inferred modality is different from the modality from the structured index field feature?

wanliAlex (changes requested):

src/marqo/s2_inference/s2_inference.py
```diff
@@ -166,6 +164,15 @@ def get_marqo_inference_cache() -> MarqoInferenceCache:
     """Returns the _marqo_inference_cache object"""
     return _marqo_inference_cache
 
+def get_encoder(model):
+    if isinstance(model, MultimodalModel):
+        if model.properties.loader == "languagebind":
+            return LanguageBindEncoder(model)
+        elif model.properties.loader == "imagebind":
+            # Return ImageBind encoder when implemented
+            pass
```
wanliAlex: Raise NotImplementedError

src/marqo/tensor_search/models/add_docs_objects.py
```diff
@@ -30,7 +30,7 @@ class Config:
     modelAuth: Optional[ModelAuth] = None
     mappings: Optional[dict] = None
     documents: Union[Sequence[Union[dict, Any]], np.ndarray]
-    imageDownloadThreadCount: int = 20
+    imageDownloadThreadCount: int = read_env_vars_and_defaults_ints(EnvVars.MARQO_MEDIA_DOWNLOAD_THREAD_COUNT)
```
wanliAlex: This is not a good place to read an environment variable. Inject it into index_management.

src/marqo/tensor_search/add_docs.py
```diff
@@ -64,120 +80,186 @@ def threaded_download_and_preprocess_images(
...
+                            ModelType.LanguageBind] and modality not in marqo_index.model.properties.get(
+                                'supported_modalities'):
+                            content_repo[doc[field]] = UnsupportedModalityError(
+                                f"Model {marqo_index.model.name} does not support {modality}")
+                            continue
+                        try:
+                            processed_chunks = download_and_chunk_media(doc[field], device, download_headers, modality,
+                                                                        marqo_index, preprocessors)
+                            content_repo[doc[field]] = processed_chunks
+                        except Exception as e:
```
wanliAlex: A general catch for Exception is not a good practice here. Try to catch more specific errors.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -144,19 +146,26 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
...
+        #    media_download_thread_count = 5
+        pass
+
     with ExitStack() as exit_stack:
-        if marqo_index.treat_urls_and_pointers_as_images:
+        if marqo_index.treat_urls_and_pointers_as_images or marqo_index.treat_urls_and_pointers_as_media: # review this logic
```
farshidz: At index creation, we must error out if these parameters disagree i.e. if treat_urls_and_pointers_as_media is true but treat_urls_and_pointers_as_images false. Also are we deprecating treat_urls_and_pointers_as_images? I think we should do that

src/marqo/tensor_search/tensor_search.py
```diff
@@ -144,19 +146,26 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
...
             with RequestMetricsStore.for_request().time(
-                    "image_download.full_time",
+                    "media_download.full_time",
                     lambda t: logger.debug(
                         f"add_documents image download: took {t:.3f}ms to concurrently download "
```
farshidz: media download?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -165,17 +174,20 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
                 tensor_fields_and_multimodal_subfields = copy.deepcopy(add_docs_params.tensor_fields) \
                     if add_docs_params.tensor_fields else []
                 tensor_fields_and_multimodal_subfields.extend(multimodal_sub_fields)
-                image_repo = exit_stack.enter_context(
-                    add_docs.download_and_preprocess_images(
-                        docs=docs, thread_count=add_docs_params.image_download_thread_count,
+                content_repo = exit_stack.enter_context(
```
farshidz: media_repo would be more consistent with your naming

src/marqo/tensor_search/tensor_search.py
```diff
@@ -165,17 +174,20 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
...
+                    add_docs.download_and_preprocess_content(
```
farshidz, line 177: Again I would stick to one terminology and media seems better than content

src/marqo/tensor_search/tensor_search.py
```diff
@@ -165,17 +174,20 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
...
+                        docs=docs,
+                        thread_count=media_download_thread_count,
```
farshidz, line 179: A coupe of issues here. 
1. You're using the same thread count for both image and video/audio. These should be configurable separately
2. You're just using image thread count anyway. 
I suggest adding a new video_audio_thread_count (or someting like that) to the API

src/marqo/tensor_search/tensor_search.py
```diff
@@ -272,10 +282,13 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
                 # C) field type is standard -> chunking and vectorisation
                 # D) field type is multimodal -> use vectorise_multimodal_combination_field (does chunking and vectorisation)
                 # Do step D regardless. It will generate separate chunks for multimodal.
+                # E) field type is video_pointer -> use video chunking and vectorisation
```
farshidz: Aren't e and f basically c?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -307,17 +320,58 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
...
+                                    id=doc_id if doc_id is not None else '',
+                                    error=str(e),
+                                    message=str(e),
+                                    status=int(errors.InvalidArgError.status_code),
+                                    code=errors.InvalidArgError.code)
+                                )
+                            )
+                            break
+
+                    elif isinstance(field_content, (str, Image.Image)) and not video_audio_check:
```
farshidz: Reverting mainline changes

src/marqo/tensor_search/tensor_search.py
```diff
@@ -307,17 +320,58 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
...
                 if len(chunks) == 0:  # Not using existing tensors or didn't find it
-                    if isinstance(field_content, str):
+                    modality = infer_modality(field_content)
+                    video_audio_check = modality in [Modality.VIDEO, Modality.AUDIO] and marqo_index.treat_urls_and_pointers_as_media
+
+                    if video_audio_check:
+                        try:
+                            # Check for UnsupportedModalityError in content_repo
+                            if isinstance(content_repo[field_content], s2_inference_errors.S2InferenceError):
+                                raise content_repo[field_content]
```
farshidz: We should never fail the whole batch. Not sure if this is handled somewhere?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -307,17 +320,58 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
...
                 # C) field type is standard
                 if len(chunks) == 0:  # Not using existing tensors or didn't find it
-                    if isinstance(field_content, str):
+                    modality = infer_modality(field_content)
+                    video_audio_check = modality in [Modality.VIDEO, Modality.AUDIO] and marqo_index.treat_urls_and_pointers_as_media
+
+                    if video_audio_check:
+                        try:
+                            # Check for UnsupportedModalityError in content_repo
+                            if isinstance(content_repo[field_content], s2_inference_errors.S2InferenceError):
```
farshidz: What if there's another error?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -307,17 +320,58 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
...
+                                raise content_repo[field_content]
+                                
+                            media_chunks = content_repo[field_content]
```
farshidz: A pydantic model is generally preferable to a dict

src/marqo/tensor_search/tensor_search.py
```diff
@@ -307,17 +320,58 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
...
+                            for chunk_index, media_chunk in enumerate(media_chunks):
+                                chunk_start = media_chunk['start_time']
+                                chunk_end = media_chunk['end_time']
+                                chunk_id = f"{field}::start_{chunk_start:.2f}::end_{chunk_end:.2f}"
+                                chunks.append(chunk_id)
```
farshidz, line 338: What can we do with the chunk ids we're storing? Is there any point?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -307,17 +320,58 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
...
+                                    content=[media_chunk['tensor']],
+                                    model_properties=marqo_index.model.get_properties(),
+                                    device=add_docs_params.device,
+                                    normalize_embeddings=marqo_index.normalize_embeddings,
+                                    infer=True,
+                                    model_auth=add_docs_params.model_auth,
+                                    modality=modality
+                                )
+                                embeddings.extend(vector)
+                        except Exception as e:
```
farshidz: Handling Exception is almost always a bad idea. You will see in existing code below we handle specific exception classes, you should follow that pattern here or reuse that code below

src/marqo/tensor_search/tensor_search.py
```diff
@@ -307,17 +320,58 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
...
+                    elif isinstance(field_content, (str, Image.Image)) and not video_audio_check:
                         # 1. check if urls should be downloaded -> "treat_pointers_and_urls_as_images":True
                         # 2. check if it is a url or pointer
                         # 3. If yes in 1 and 2, download blindly (without type)
                         # 4. Determine media type of downloaded
                         # 5. load correct media type into memory -> PIL (images), videos (), audio (torchaudio)
                         # 6. if chunking -> then add the extra chunker
 
-                        if not _is_image(field_content):
+                        if isinstance(field_content, str) and not _is_image(field_content):
```
farshidz: Reverting mainline?

src/marqo/tensor_search/tensor_search.py
farshidz: Isn't `not video_audio_check` always true if you reach this line?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -336,12 +390,12 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
...
                                 if isinstance(field_content, str) and marqo_index.treat_urls_and_pointers_as_images:
-                                    if not isinstance(image_repo[field_content], Exception):
-                                        image_data = image_repo[field_content]
+                                    if not isinstance(content_repo[field_content], Exception):
+                                        image_data = content_repo[field_content]
                                     else:
                                         raise s2_inference_errors.S2InferenceError(
-                                            f"Could not find image at `{field_content}`. \n"
-                                            f"Reason: {str(image_repo[field_content])}"
+                                            f"Could not process the media file found at `{field_content}`. \n"
```
farshidz, line 395: Revering mainline change? Also here we know it's an image

src/marqo/s2_inference/multimodal_utils.py
```diff
@@ -0,0 +1,286 @@
...
+                break
+        buffer.seek(0)
+        yield buffer
+    finally:
+        buffer.close()
+        response.close()
+
+def infer_modality(content: Union[str, List[str], bytes]) -> Modality:
+    if isinstance(content, str):
+        if not validators.url(content):
```
farshidz: I think there was a change to encode the url. Please check with Li on this

farshidz (changes requested):

src/marqo/tensor_search/tensor_search.py
```diff
@@ -574,39 +626,57 @@ def _add_documents_structured(config: Config, add_docs_params: AddDocsParams, ma
...
 
-                image_repo = exit_stack.enter_context(
-                    add_docs.download_and_preprocess_images(
-                        docs=docs, thread_count=add_docs_params.image_download_thread_count,
-                        tensor_fields=image_fields,
-                        image_download_headers=add_docs_params.image_download_headers,
+                content_repo = exit_stack.enter_context(
+                    add_docs.download_and_preprocess_content(
+                        docs=docs, 
+                        thread_count=media_download_thread_count,
```
farshidz, line 676: Similar issue of limiting image thread count as well. See unstructured comment

src/marqo/tensor_search/tensor_search.py
```diff
@@ -759,8 +832,44 @@ def _add_documents_structured(config: Config, add_docs_params: AddDocsParams, ma
...
+                                    model_properties=marqo_index.model.get_properties(),
+                                    device=add_docs_params.device,
+                                    normalize_embeddings=marqo_index.normalize_embeddings,
+                                    infer=True,
+                                    model_auth=add_docs_params.model_auth,
+                                    modality=Modality.VIDEO if marqo_field.type == FieldType.VideoPointer else Modality.AUDIO
+                                )
+                                embeddings.extend(vector)  # vectorise returns a list of vectors
+                                
+                        except Exception as e:
```
farshidz: Handling generic Exception class. See comment on this above

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1519,20 +1632,21 @@ def construct_vector_input_batches(query: Optional[Union[str, Dict]], index_info
         A tuple of string batches. The first is text content the second is image content.
     """
     # TODO - infer this from model
-    treat_urls_as_images = True
+    treat_urls_as_media = True
+
     if isinstance(query, str):
-        if treat_urls_as_images and _is_image(query):
+        if treat_urls_as_media and _is_image(query):
```
farshidz: Are we not supporting audio/video as search query?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1770,7 +1886,8 @@ def get_content_vector(possible_jobs: List[VectorisedJobPointer], job_to_vectors
 
     Raises runtime error if is not found
     """
-    content_type = 'image' if treat_urls_as_images and _is_image(content) else 'text'
+    content_type = 'text' if infer_modality(content) == Modality.TEXT else 'media'
```
farshidz, line 1926: Do we ever get `media` from `jobs[vec_job_pointer.job_hash].content_type`?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1756,12 +1872,12 @@ def get_query_vectors_from_jobs(
 
 def get_content_vector(possible_jobs: List[VectorisedJobPointer], job_to_vectors: Dict[JHash, Dict[str, List[float]]],
                        jobs: Dict[JHash, VectorisedJobs],
-                       treat_urls_as_images: bool, content: str) -> List[float]:
+                       treat_urls_as_media: bool, content: str) -> List[float]:
```
farshidz: `treat_urls_as_media` is unused. Why? You can use your IDE to find instnaces of unused variables as there are quite a few, and investigate these

src/marqo/tensor_search/tensor_search.py
```diff
@@ -2051,91 +2172,140 @@ def vectorise_multimodal_combination_field_unstructured(field: str,
...
+    video_content_to_vectorise = []
+    audio_field_names = []
+    audio_content_to_vectorise = []
 
     normalize_embeddings = marqo_index.normalize_embeddings
     infer_if_image = marqo_index.treat_urls_and_pointers_as_images
+    infer_if_media = marqo_index.treat_urls_and_pointers_as_media
 
-    if infer_if_image is False:
+    if infer_if_image is False and infer_if_media is False:
```
farshidz: `not infer_if_image and not infer_if_media` would be more idiomatic

src/marqo/tensor_search/tensor_search.py
RaynorChavez: Yeah, we only use one thread count variable. I don't think we can have different thread counts for different modalities (under the current paradigm at least). For example a batch has documents containing 3 modalities at the same time. To make different thread counts work, we'd have to split up the documents into categories, then multithread download each modality group one after the other with the corresponding number of threads the user specified for that modality, or perhaps mini group them and dynamically change the number of threads as different modalities come in. Can you think of a better way to implement this idea?

This would require quite a bit of refactoring of our code in this file. Is it super necessary at this point? Perhaps we can do this in a future release?

src/marqo/tensor_search/tensor_search.py
RaynorChavez: we do, that's why it is treat_urls_as_media. I made a test for video search.

src/marqo/tensor_search/tensor_search.py
RaynorChavez: yes, for when the user uses a video audio or image link in the query

src/marqo/tensor_search/tensor_search.py
RaynorChavez: This was just leftover code. Will delete this one and investigate other ones.

src/marqo/core/unstructured_vespa_index/unstructured_vespa_schema.py
farshidz: Looks like 3 should be a validation error

src/marqo/tensor_search/tensor_search.py
RaynorChavez: True, thanks

src/marqo/tensor_search/tensor_search.py
RaynorChavez: similar comment as well, copying here


Yeah, we only use one thread count variable. I don't think we can have different thread counts for different modalities (under the current paradigm at least). For example a batch has documents containing 3 modalities at the same time. To make different thread counts work, we'd have to split up the documents into categories, then multithread download each modality group one after the other with the corresponding number of threads the user specified for that modality, or perhaps mini group them and dynamically change the number of threads as different modalities come in. Can you think of a better way to implement this idea?

This would require quite a bit of refactoring of our code in this file. Is it super necessary at this point? Perhaps we can do this in a future release?

src/marqo/tensor_search/tensor_search.py
RaynorChavez: ah yeah that's redundant

src/marqo/tensor_search/tensor_search.py
RaynorChavez: This should be an `s2_inference_errors.S2InferenceError as e:` I will change it

src/marqo/tensor_search/add_docs_utils.py
```diff
@@ -0,0 +1,138 @@
...
+from typing import List, Dict
+import time
+import ffmpeg
+
+from torchvision.transforms import Compose
+from marqo.core.models.marqo_index import *
+from marqo.s2_inference.s2_inference import  Modality
+
+
+class StreamingMediaProcessor:
```
farshidz: Why is this module not named after the class? As noted before utility modules are almost always a bad idea

src/marqo/tensor_search/tensor_search.py
RaynorChavez: Yes, the purpose of saving this chunk_id is to tell the user the start and end time for the chunk and their associated embedding. For example in tensor_facets

'_tensor_facets': [
        {'video_field': 'start_0.00::end_10.00', 
        '_embedding': [
            -0.9606754779815674, 
            -0.17251494526863098, 
            0.5819228291511536, 
            ...]},
        {'video_field': 'start_6.56::end_16.56', 
        '_embedding': [
            1.3152823448181152, 
            0.7909719347953796, 
            -1.2539052963256836, 
            ...]},
        ]

src/marqo/tensor_search/tensor_search.py
RaynorChavez: Yep, should be `s2_inference_errors.S2InferenceError as e:` like the others

src/marqo/tensor_search/tensor_search.py
RaynorChavez: ah but the _is_image check is confusing right? 

It probably works right now because _is_image only really checks if it is a valid url so this passes. I will change it to something clearer

src/marqo/tensor_search/add_docs_utils.py
RaynorChavez: Thanks, I made utility modules because I did not recall that we were phasing them out. I'll rename this and try to get rid of multimodal_utils

src/marqo/s2_inference/multimodal_utils.py
RaynorChavez: Thanks, that's right

src/marqo/s2_inference/multimodal_utils.py
RaynorChavez: Default should be text right? So it's alright that we currently return text as well if it's not detected as either image, video, or audio.

src/marqo/s2_inference/multimodal_utils.py
RaynorChavez: Right, let's just raise an error. If the content is not a string, a list of strings, or bytes, there's probably something wrong

src/marqo/tensor_search/tensor_search.py
RaynorChavez: This is deleted. Leftover for when we wanted to limit media thread count to 5 or less for video and audio

src/marqo/tensor_search/tensor_search.py
RaynorChavez: will change to `add_documents media download: took ...`

src/marqo/tensor_search/tensor_search.py
RaynorChavez: will replace

src/marqo/tensor_search/tensor_search.py
RaynorChavez: will replace with media_repo

src/marqo/tensor_search/tensor_search.py
RaynorChavez: This is true

src/marqo/api/configs.py
RaynorChavez: Agree

src/marqo/core/models/marqo_index.py
RaynorChavez: If the user doesn't provide them then they should be populated in `class IndexSettings` so at the end of index creation they should have values.

I think we need to make them Optional now to avoid backwards compatibility issues, like if an older Index does not have these fields, they would error out or at least give a warning of missing field.

src/marqo/s2_inference/multimodal_utils.py
RaynorChavez: Great!

src/marqo/s2_inference/s2_inference.py
RaynorChavez: Yeah good idea

src/marqo/tensor_search/add_docs.py
RaynorChavez: Isn't this case covered by 
`if modality == Modality.IMAGE or (marqo_index.type == IndexType.Structured and media_field_types_mapping[field] == FieldType.ImagePointer):` ?

src/marqo/tensor_search/add_docs.py
RaynorChavez: even if we set the content[repo] value to an s2inference error?

except Exception as e:
                            logger.error(f"Error processing {modality} file: {str(e)}")
                            media_repo[doc[field]] = S2InferenceError(f"Error processing {modality} file: {str(e)}")

src/marqo/tensor_search/tensor_search.py
RaynorChavez: Yeah it would be good to depreciate it

src/marqo/s2_inference/multimodal_utils.py
RaynorChavez: But what if it's a number? Should be treated as text right? I will return Text for now

src/marqo/s2_inference/multimodal_utils.py
RaynorChavez: Done

src/marqo/tensor_search/tensor_search.py
RaynorChavez: removed

src/marqo/s2_inference/multimodal_utils.py
RaynorChavez: done

src/marqo/s2_inference/s2_inference.py
RaynorChavez: done

src/marqo/s2_inference/s2_inference.py
RaynorChavez: Done

src/marqo/s2_inference/multimodal_utils.py
RaynorChavez: done

src/marqo/api/configs.py
RaynorChavez: Done

src/marqo/tensor_search/tensor_search.py
RaynorChavez: done

src/marqo/tensor_search/tensor_search.py
RaynorChavez: done

src/marqo/tensor_search/tensor_search.py
RaynorChavez: done

src/marqo/tensor_search/tensor_search.py
RaynorChavez: done

src/marqo/s2_inference/multimodal_utils.py
RaynorChavez: done

src/marqo/s2_inference/s2_inference.py
RaynorChavez: done

src/marqo/tensor_search/tensor_search.py
RaynorChavez: replaced

src/marqo/s2_inference/languagebind/audio/modeling_audio.py
RaynorChavez: done for add_docs_utils

src/marqo/tensor_search/tensor_search.py
RaynorChavez: done

src/marqo/core/models/marqo_index.py
RaynorChavez: I don't think we can do this since index_management deals with index settings right? And this one is an add docs object.

but I did push new code which:
- adds a new env var. Now we have two env vars for download threads: (1) image threadcount, and (2) media thread count. This is for depreciating the image one. Both can't be set explicitly at the same time or it will error.
- in tensor_search, a new method which selects the proper threadcount variable to use andsets the default media for languagebind to 5 threads.

src/marqo/tensor_search/add_docs.py
RaynorChavez: We don't use memory pool anymore, so I think it's safe to close this!

src/marqo/tensor_search/add_docs.py
RaynorChavez: We don't use memory pool anymore, or limit chunks by bytsize so I think it's safe to close this!

src/marqo/tensor_search/add_docs.py
RaynorChavez: We don't use this techinque anymore so I will be closing the issue!

requirements.txt
RaynorChavez: done

src/marqo/core/unstructured_vespa_index/unstructured_vespa_schema.py
RaynorChavez: done

.gitignore
RaynorChavez: done

requirements.txt
RaynorChavez: done

tests/tensor_search/integ_tests/test_search_combined.py
RaynorChavez: done

src/marqo/s2_inference/model_registry.py
RaynorChavez: done

src/marqo/tensor_search/models/add_docs_objects.py
RaynorChavez: I don't think we can do this since index_management deals with index settings right? And this one is an add docs object.

but I did push new code which:

- adds a new env var. Now we have two env vars for download threads: (1) image threadcount, and (2) media thread count. This is for depreciating the image one. Both can't be set explicitly at the same time or it will error.
- in tensor_search, a new method which selects the proper threadcount variable to use andsets the default media for languagebind to 5 threads.

src/marqo/s2_inference/multimodal_utils.py
RaynorChavez: renamed the file

src/marqo/tensor_search/tensor_search.py
RaynorChavez: This code block shouldn't fail the whole batch, it just checks the value of that specific field. If the value of the field is an `S2InferenceError`, then the code goes to the `except` block

except s2_inference_errors.S2InferenceError as e:
                            document_is_valid = False
                            unsuccessful_docs.append(
                                (i, MarqoAddDocumentsItem(
                                    id=doc_id if doc_id is not None else '',
                                    error=str(e),
                                    message=str(e),
                                    status=int(errors.InvalidArgError.status_code),
                                    code=errors.InvalidArgError.code)
                                )
                            )
                            break
                            
and that specific field is appended to the unsuccessful docs list.

src/marqo/tensor_search/tensor_search.py
RaynorChavez: I think I just followed the existing logic here. Perhaps we could make it a pydantic model in the future?

src/marqo/tensor_search/tensor_search.py
RaynorChavez: This code block is just to get the document appended into the `unsuccessful_docs` list in the except block, 

except s2_inference_errors.S2InferenceError as e:
                            document_is_valid = False
                            unsuccessful_docs.append(
                                (i, MarqoAddDocumentsItem(
                                    id=doc_id if doc_id is not None else '',
                                    error=str(e),
                                    message=str(e),
                                    status=int(errors.InvalidArgError.status_code),
                                    code=errors.InvalidArgError.code)
                                )
                            )
                            break
                            
Other errors related to that video/audio doc should be handled in `StreamingMediaProcessor`. And beyond that, it should just follow existing erroring logic in `tensor_search`

src/marqo/tensor_search/add_docs.py
RaynorChavez: Good idea. Do you think we can do this in the next release?

src/marqo/tensor_search/tensor_search.py
RaynorChavez: fixed

src/marqo/tensor_search/tensor_search.py
RaynorChavez: fixed

src/marqo/s2_inference/languagebind/audio/modeling_audio.py
RaynorChavez: done

src/marqo/s2_inference/multimodal_utils.py
RaynorChavez: @wanliAlex , could you please confirm?

src/marqo/core/models/marqo_index.py
farshidz: I'm guessing also a backwards compatibility consideration so we can deserialize old indexes

src/marqo/tensor_search/tensor_search.py
RaynorChavez: done

src/marqo/tensor_search/add_docs.py
wanliAlex: Not, it is still a bad practice. For example, you can even catch the `KeyboardInterrupt` in this case. 

And if you really have no idea which error will be raised, try to get inside the function `download_and_chunk_media`, catch the error inside it and raise `download_and_chunk_media` error, and catch it here again.

src/marqo/s2_inference/multimodal_utils.py
wanliAlex: Yes, when we validate the url, we use `if validators.url(content) or validators.url(encoded(content))`

src/marqo/tensor_search/add_docs.py
farshidz: I agree catching Exception is almost always bad

src/marqo/core/models/marqo_index.py
wanliAlex: Good consideration on the backward compatibility issue! Thanks

src/marqo/tensor_search/models/add_docs_objects.py
wanliAlex: got it

src/marqo/s2_inference/multimodal_utils.py
RaynorChavez: done, switched to using `validate_url`

src/marqo/tensor_search/add_docs.py
wanliAlex: If your inferred modality is Image and the field type is video_pointer, what will happen?

src/marqo/tensor_search/add_docs.py
wanliAlex: No problem

wanliAlex (changes requested):

src/marqo/tensor_search/add_docs.py
RaynorChavez: done

wanliAlex (approved):

src/marqo/tensor_search/add_docs.py
```diff
@@ -64,119 +82,185 @@ def threaded_download_and_preprocess_images(
...
+                        try:
+                            processed_chunks = download_and_chunk_media(doc[field], device, download_headers, modality,
+                                                                        marqo_index, preprocessors)
+                            media_repo[doc[field]] = processed_chunks
+                        except (ffmpeg.Error, S2InferenceError) as e:
+                            logger.error(f"Error processing {modality} file: {str(e)}")
+                            media_repo[doc[field]] = S2InferenceError(f"Error processing {modality} file: {str(e)}")
+
+                    else:
+                        pass
```
wanliAlex, line 149: Should we raise an error here?

wanliAlex (changes requested):

src/marqo/tensor_search/add_docs.py
RaynorChavez: done

wanliAlex: All good. A key error (which goes to a 500) is not a bad idea in this case.

farshidz (approved):

wanliAlex (approved):

PR #936, opened 2024-08-15
Add encode for url check

wanliAlex:
bug fix

We currently use validators.url to check the validity of a provided URL. However, when the URL contains unencoded characters, the function returns False.

The function has been updated to return True if the URL is valid after encoding.

no

running

* **Other information**:

### DIFF
--- a/src/marqo/s2_inference/clip_utils.py
+++ b/src/marqo/s2_inference/clip_utils.py
@@ -106,7 +106,7 @@ def load_image_from_path(image_path: str, image_download_headers: dict, timeout_
     """
     if os.path.isfile(image_path):
         img = Image.open(image_path)
-    elif validators.url(image_path):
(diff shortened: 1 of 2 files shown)


papa99do (dismissed):

src/marqo/s2_inference/clip_utils.py
```diff
@@ -348,7 +363,7 @@ def load(self) -> None:
                 self.model_path = self._download_from_repo()
             elif os.path.isfile(path):
                 self.model_path = path
-            elif validators.url(path):
+            elif validate_url(path):
                 self.model_path = download_model(url=path)
```
papa99do, line 367: we don't encode url here. So the validation can pass but the download can fail. (if download_mode cannot not handle unencoded characters in the url)

PR #958, opened 2024-09-10
Patch errors made for when user enters wrong modality into a structured pointer field

RaynorChavez:
Bug fix

Generic Internal error when user 
- adds text to a video/audio pointer field in structured
- adds video link to audio pointer and vice version in structured

Only the specific document is failed to 400, but the batch is not.

No

* **Other information**:

### DIFF
--- a/src/marqo/s2_inference/errors.py
+++ b/src/marqo/s2_inference/errors.py
@@ -7,6 +7,8 @@ class S2InferenceError(Exception):
             self.message = message
             super().__init__(self.message)
 
+class MediaMismatchError(S2InferenceError):
+    pass
 
 class ChunkerError(S2InferenceError):
     pass
@@ -64,6 +66,8 @@ class BatchInferenceSizeNotMatchError(S2InferenceError):
 class ImageDownloadError(S2InferenceError):
     pass
 
+class MediaDownloadError(S2InferenceError):
+    pass
 
 class UnsupportedModalityError(S2InferenceError):
     pass
--- a/src/marqo/s2_inference/multimodal_model_load.py
+++ b/src/marqo/s2_inference/multimodal_model_load.py
@@ -152,13 +152,10 @@ def infer_modality(content: Union[str, List[str], bytes]) -> Modality:
 
         extension = encoded_url.split('.')[-1].lower()
         if extension in ['jpg', 'jpeg', 'png', 'gif', 'webp']:
-            print(f"inferred image")
             return Modality.IMAGE
         elif extension in ['mp4', 'avi', 'mov']:
-            print(f"inferred video")
             return Modality.VIDEO
         elif extension in ['mp3', 'wav', 'ogg']:
-            print(f"inferred audio")
             return Modality.AUDIO
 
         if validate_url(encoded_url):
@@ -167,13 +164,10 @@ def infer_modality(content: Union[str, List[str], bytes]) -> Modality:
                 with fetch_content_sample(encoded_url) as sample:
                     mime = magic.from_buffer(sample.read(), mime=True)
                     if mime.startswith('image/'):
-                        print(f"inferred image from url")
                         return Modality.IMAGE
                     elif mime.startswith('video/'):
-                        print(f"inferred video from url")
                         return Modality.VIDEO
                     elif mime.startswith('audio/'):
-                        print(f"inferred audio from url")
                         return Modality.AUDIO
             except requests.exceptions.RequestException as e:
                 pass
--- a/src/marqo/s2_inference/multimodal_model_load.py
+++ b/src/marqo/s2_inference/multimodal_model_load.py
@@ -176,9 +176,7 @@ def infer_modality(content: Union[str, List[str], bytes]) -> Modality:
                 raise MediaDownloadError(f"Error determining MIME type for {encoded_url}: {e}") from e
             except IOError as e:
                 raise MediaDownloadError(f"IO error while processing {encoded_url}: {e}") from e
-            except Exception as e:
-                raise MediaDownloadError(f"Unexpected error while processing {encoded_url}: {e}") from e
-
+            
         return Modality.TEXT
 
     elif isinstance(content, bytes):
--- a/src/marqo/s2_inference/multimodal_model_load.py
+++ b/src/marqo/s2_inference/multimodal_model_load.py
@@ -171,7 +171,13 @@ def infer_modality(content: Union[str, List[str], bytes]) -> Modality:
                     elif mime.startswith('audio/'):
                         return Modality.AUDIO
             except requests.exceptions.RequestException as e:
-                raise MediaDownloadError(f"Error downloading media file {content}: {e}")
+                raise MediaDownloadError(f"Error downloading media file {content}: {e}") from e
+            except magic.MagicException as e:
+                raise MediaDownloadError(f"Error determining MIME type for {encoded_url}: {e}") from e
+            except IOError as e:
+                raise MediaDownloadError(f"IO error while processing {encoded_url}: {e}") from e
+            except Exception as e:
+                raise MediaDownloadError(f"Unexpected error while processing {encoded_url}: {e}") from e
 
         return Modality.TEXT
 
--- a/src/marqo/s2_inference/multimodal_model_load.py
+++ b/src/marqo/s2_inference/multimodal_model_load.py
@@ -14,6 +14,8 @@ from abc import ABC, abstractmethod
 from typing import List, Dict, Any, Union
 from PIL.Image import Image
 import torch
+from urllib.parse import quote
+
 
 from marqo.s2_inference.multimodal_model_load import *
 from marqo.s2_inference.languagebind import (
@@ -143,13 +145,19 @@ def infer_modality(content: Union[str, List[str], bytes]) -> Modality:
     if isinstance(content, str):
         if not validate_url(content):
             return Modality.TEXT
+        
+        # Encode the URL
+        encoded_url = quote(content, safe=':/')
 
-        extension = content.split('.')[-1].lower()
+        extension = encoded_url.split('.')[-1].lower()
         if extension in ['jpg', 'jpeg', 'png', 'gif', 'webp']:
+            print(f"inferred image")
             return Modality.IMAGE
         elif extension in ['mp4', 'avi', 'mov']:
+            print(f"inferred video")
             return Modality.VIDEO
         elif extension in ['mp3', 'wav', 'ogg']:
+            print(f"inferred audio")
             return Modality.AUDIO
 
         if validate_url(content):
@@ -158,13 +166,17 @@ def infer_modality(content: Union[str, List[str], bytes]) -> Modality:
                 with fetch_content_sample(content) as sample:
                     mime = magic.from_buffer(sample.read(), mime=True)
                     if mime.startswith('image/'):
+                        print(f"inferred image from url")
                         return Modality.IMAGE
                     elif mime.startswith('video/'):
+                        print(f"inferred video from url")
                         return Modality.VIDEO
                     elif mime.startswith('audio/'):
+                        print(f"inferred audio from url")
                         return Modality.AUDIO
-            except Exception as e:
-                pass
+            except requests.exceptions.RequestException as e:
+                print(f"Request failed: {e}")
+                raise ValueError("Failed to fetch content from URL.")
 
         return Modality.TEXT
 
--- a/src/marqo/s2_inference/multimodal_model_load.py
+++ b/src/marqo/s2_inference/multimodal_model_load.py
@@ -23,6 +23,7 @@ from marqo.s2_inference.languagebind import (
     LanguageBindVideoProcessor, LanguageBindAudioProcessor, LanguageBindImageProcessor,
     to_device
 )
+from marqo.s2_inference.errors import MediaDownloadError
 from marqo.core.inference.image_download import encode_url
 from marqo.s2_inference.clip_utils import download_image_from_url, validate_url
 from marqo.s2_inference.languagebind.image.tokenization_image import LanguageBindImageTokenizer
@@ -170,7 +171,7 @@ def infer_modality(content: Union[str, List[str], bytes]) -> Modality:
                     elif mime.startswith('audio/'):
                         return Modality.AUDIO
             except requests.exceptions.RequestException as e:
-                pass
+                raise MediaDownloadError(f"Error downloading media file {content}: {e}")
 
         return Modality.TEXT
 
@@ -188,8 +189,7 @@ def infer_modality(content: Union[str, List[str], bytes]) -> Modality:
 
     else:
         return Modality.TEXT
-        # raise ValueError(f"Unsupported content type: {type(content)}.
-        # It is neither a string, list of strings, nor bytes.")
+
 
 
 class LanguageBindEncoder(ModelEncoder):
--- a/src/marqo/s2_inference/multimodal_model_load.py
+++ b/src/marqo/s2_inference/multimodal_model_load.py
@@ -23,6 +23,7 @@ from marqo.s2_inference.languagebind import (
     LanguageBindVideoProcessor, LanguageBindAudioProcessor, LanguageBindImageProcessor,
     to_device
 )
+from marqo.core.inference.image_download import encode_url
 from marqo.s2_inference.clip_utils import download_image_from_url, validate_url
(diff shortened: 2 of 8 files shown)


src/marqo/tensor_search/add_docs.py
```diff
@@ -91,7 +91,7 @@ def threaded_download_and_preprocess_content(allocated_docs: List[dict],
                     modality = infer_modality(doc[field])
```
wanliAlex: rename it to `inferred_modality` for clarificaiton

src/marqo/tensor_search/add_docs.py
```diff
@@ -129,25 +129,40 @@ def threaded_download_and_preprocess_content(allocated_docs: List[dict],
...
+                        if is_structured_index:
+                            if modality is Modality.VIDEO and media_field_types_mapping[
+                            field] is not FieldType.VideoPointer:
+                                media_repo[doc[field]] = S2InferenceError(
+                                    f"Invalid audio file. Error processing media file {doc}, detected as video, but field type is not VideoPointer")
+                                continue
+
+                            if modality is Modality.AUDIO and media_field_types_mapping[
+                            field] is not FieldType.AudioPointer:
+                                media_repo[doc[field]] = S2InferenceError(
```
wanliAlex: Create a new error for this. S2InferenceError is too generic. Maybe `MediaMismatchError`?  Make sure the error is caught properly so it does not break the whole batch but just returns 400 for the document.

src/marqo/tensor_search/add_docs.py
```diff
@@ -91,7 +91,7 @@ def threaded_download_and_preprocess_content(allocated_docs: List[dict],
                     modality = infer_modality(doc[field])
                     if (modality == Modality.IMAGE and is_no_index) or (
```
wanliAlex: It's hard to follow the whole logic here. Could it be simplified? Like raise the error earlier if inferred error does not match the field?

wanliAlex (changes requested):

src/marqo/tensor_search/add_docs.py
wanliAlex: Your infer_modality function is not reliable:
1. It does not encoded the url
2. It shouldn't return text if there is an exception. What if a time-out error happens?

src/marqo/s2_inference/multimodal_model_load.py
```diff
@@ -143,13 +145,19 @@ def infer_modality(content: Union[str, List[str], bytes]) -> Modality:
     if isinstance(content, str):
         if not validate_url(content):
             return Modality.TEXT
+        
+        # Encode the URL
+        encoded_url = quote(content, safe=':/')
```
wanliAlex: Use the function here: https://github.com/marqo-ai/marqo/blob/0f09c36f2f2ab0add8cbb07ce939000fffb932d3/src/marqo/core/inference/image_download.py#L218

src/marqo/s2_inference/multimodal_model_load.py
```diff
@@ -158,13 +166,17 @@ def infer_modality(content: Union[str, List[str], bytes]) -> Modality:
                 with fetch_content_sample(content) as sample:
```
wanliAlex: we should pass the encoded_url when fetching the sample?

src/marqo/s2_inference/multimodal_model_load.py
```diff
@@ -158,13 +166,17 @@ def infer_modality(content: Union[str, List[str], bytes]) -> Modality:
...
+                        print(f"inferred video from url")
                         return Modality.VIDEO
                     elif mime.startswith('audio/'):
+                        print(f"inferred audio from url")
                         return Modality.AUDIO
-            except Exception as e:
-                pass
+            except requests.exceptions.RequestException as e:
+                print(f"Request failed: {e}")
+                raise ValueError("Failed to fetch content from URL.")
```
wanliAlex: Why do we raise a ValueError here? Will this become an internal error?

src/marqo/tensor_search/add_docs.py
wanliAlex: Should we catch the ValueError here?

wanliAlex (changes requested):

src/marqo/tensor_search/add_docs.py
RaynorChavez: done

src/marqo/s2_inference/multimodal_model_load.py
RaynorChavez: done

src/marqo/s2_inference/multimodal_model_load.py
RaynorChavez: done

src/marqo/s2_inference/multimodal_model_load.py
RaynorChavez: revert to pass

src/marqo/tensor_search/add_docs.py
RaynorChavez: done

src/marqo/tensor_search/add_docs.py
RaynorChavez: will refactor in the next release!

src/marqo/tensor_search/add_docs.py
RaynorChavez: This might affect a bunch of tests/logic we have, esp in tensor_search. This will require a bit more investigation. Not sure if it fits here. I think we can leave it at `S2InferenceError` for now

src/marqo/tensor_search/add_docs.py
RaynorChavez: I think the pass and being caught in add_docs.py should suffice for now

src/marqo/tensor_search/add_docs.py
```diff
@@ -88,10 +88,10 @@ def threaded_download_and_preprocess_content(allocated_docs: List[dict],
                 if field not in tensor_fields:
                     continue
                 if isinstance(doc[field], str) or force_download:
-                    modality = infer_modality(doc[field])
-                    if (modality == Modality.IMAGE and is_no_index) or (
-                        modality == Modality.IMAGE and is_unstructured_index) or (
-                        is_structured_index and media_field_types_mapping[field] == FieldType.ImagePointer):
+                    inferred_modality = infer_modality(doc[field])
```
wanliAlex: Why don't we catch the requests error here and return the error message to the user?

src/marqo/tensor_search/add_docs.py
```diff
@@ -88,10 +88,10 @@ def threaded_download_and_preprocess_content(allocated_docs: List[dict],
                 if field not in tensor_fields:
```
wanliAlex, line 90: @farshidz 
There are two ways to solve the issue when users provide an incorrect modality for a structured index:
- We infer the modality and see if there is a mismatch, which is what we are doing in this PR
- We don’t infer the modality and just process it as if it is the expected content and catch the raised error

Option 1 is more effective, but we need to be confident with our inferred modality.
Option 2 is more straightforward but it is more complicated and slower because we need to download the content, process it, and catch the error.

I think option 1 is good because we are using a widely used `mime` package now. Your thoughts?

src/marqo/tensor_search/add_docs.py
```diff
@@ -127,27 +127,42 @@ def threaded_download_and_preprocess_content(allocated_docs: List[dict],
...
                             media_repo[doc[field]] = processed_chunks
                         except (ffmpeg.Error, S2InferenceError) as e:
-                            logger.error(f"Error processing {modality} file: {str(e)}")
-                            media_repo[doc[field]] = S2InferenceError(f"Error processing {modality} file: {str(e)}")
-
+                            logger.error(f"Error processing {inferred_modality} file: {str(e)}")
+                            media_repo[doc[field]] = S2InferenceError(f"Error processing {inferred_modality} file: {str(e)}")
+                    
+                    elif inferred_modality is Modality.TEXT and is_structured_index and media_field_types_mapping[field] in [FieldType.AudioPointer, FieldType.VideoPointer, FieldType.ImagePointer]:
+                        media_repo[doc[field]] = S2InferenceError(f"Error processing media file {doc}, detected as text or failed to determine modality")
```
wanliAlex: @farshidz 
If we had a network issue when inferring the modality, it would not be included in the error message to the user. Working on this will require extra time. Is this generic error message acceptable?

src/marqo/tensor_search/add_docs.py
RaynorChavez: it might lead to an internal error which would fail the whole batch

catching it and passing to all the places which call infer_modality would take a bit of time. 

I think add_docs logic right now catches this failure but only returns something generic like 

`f"Error processing media file {doc}, detected as text or failed to determine modality"`

RaynorChavez: There's surely a better way in the future though

src/marqo/tensor_search/add_docs.py
```diff
@@ -88,17 +88,17 @@ def threaded_download_and_preprocess_content(allocated_docs: List[dict],
                 if field not in tensor_fields:
                     continue
                 if isinstance(doc[field], str) or force_download:
-                    modality = infer_modality(doc[field])
-                    if (modality == Modality.IMAGE and is_no_index) or (
-                        modality == Modality.IMAGE and is_unstructured_index) or (
-                        is_structured_index and media_field_types_mapping[field] == FieldType.ImagePointer):
+                    inferred_modality = infer_modality(doc[field])
+                    if (inferred_modality == Modality.IMAGE and is_no_index) or (
```
farshidz: `is_no_index` is confusing. We shouldn't have code just to make testing easier

src/marqo/tensor_search/add_docs.py
```diff
@@ -88,17 +88,17 @@ def threaded_download_and_preprocess_content(allocated_docs: List[dict],
...
+                        inferred_modality == Modality.IMAGE and is_unstructured_index) or (
+                        is_structured_index and media_field_types_mapping[field] == FieldType.ImagePointer and inferred_modality not in [Modality.TEXT, Modality.AUDIO, Modality.VIDEO]):
```
farshidz: Why not in if image is what you're after?

src/marqo/tensor_search/add_docs.py
```diff
@@ -127,27 +127,42 @@ def threaded_download_and_preprocess_content(allocated_docs: List[dict],
                                 else:
                                     raise e
 
-                    elif (modality in [Modality.VIDEO, Modality.AUDIO] and is_no_index) or (
-                            modality in [Modality.VIDEO, Modality.AUDIO] and is_unstructured_index) or (
-                            is_structured_index and media_field_types_mapping[field] in [FieldType.AudioPointer, FieldType.VideoPointer]):
-                            
+                    elif (inferred_modality in [Modality.VIDEO, Modality.AUDIO] and is_no_index) or (
+                            inferred_modality in [Modality.VIDEO, Modality.AUDIO] and is_unstructured_index) or (
+                            is_structured_index and media_field_types_mapping[field] in [FieldType.AudioPointer, FieldType.VideoPointer] and inferred_modality not in [Modality.TEXT, Modality.IMAGE]):
```
farshidz: not in is still confusing and risky

farshidz (changes requested):

farshidz:
To mitigate regression risk, for structured image pointer we shouldn't use the infer process

src/marqo/s2_inference/multimodal_model_load.py
```diff
@@ -143,27 +146,30 @@ def infer_modality(content: Union[str, List[str], bytes]) -> Modality:
...
+                with fetch_content_sample(encoded_url) as sample:
                     mime = magic.from_buffer(sample.read(), mime=True)
                     if mime.startswith('image/'):
                         return Modality.IMAGE
                     elif mime.startswith('video/'):
                         return Modality.VIDEO
                     elif mime.startswith('audio/'):
                         return Modality.AUDIO
-            except Exception as e:
+            except requests.exceptions.RequestException as e:
```
farshidz, line 173: This should be an error for the document

src/marqo/tensor_search/add_docs.py
RaynorChavez: done

src/marqo/tensor_search/add_docs.py
RaynorChavez: removed this as discussed, infer_modality won't be used for structured image pointers

src/marqo/tensor_search/add_docs.py
RaynorChavez: changed to in [video, audio]

src/marqo/tensor_search/add_docs.py
RaynorChavez: Done, changed to pass the error to doc

src/marqo/tensor_search/add_docs.py
RaynorChavez: scratch the above. catching the error

src/marqo/tensor_search/add_docs.py
RaynorChavez: removed

src/marqo/s2_inference/multimodal_model_load.py
```diff
@@ -143,28 +147,31 @@ def infer_modality(content: Union[str, List[str], bytes]) -> Modality:
...
                     if mime.startswith('image/'):
                         return Modality.IMAGE
                     elif mime.startswith('video/'):
                         return Modality.VIDEO
                     elif mime.startswith('audio/'):
                         return Modality.AUDIO
-            except Exception as e:
-                pass
+            except requests.exceptions.RequestException as e:
+                raise MediaDownloadError(f"Error downloading media file {content}: {e}")
```
wanliAlex: raise MediaDownloadError from e

src/marqo/s2_inference/errors.py
```diff
@@ -64,6 +66,8 @@ class BatchInferenceSizeNotMatchError(S2InferenceError):
 class ImageDownloadError(S2InferenceError):
     pass
 
+class MediaDownloadError(S2InferenceError):
```
wanliAlex, line 69: Nice design!

src/marqo/s2_inference/multimodal_model_load.py
```diff
@@ -143,28 +147,31 @@ def infer_modality(content: Union[str, List[str], bytes]) -> Modality:
...
                     mime = magic.from_buffer(sample.read(), mime=True)
                     if mime.startswith('image/'):
                         return Modality.IMAGE
                     elif mime.startswith('video/'):
                         return Modality.VIDEO
                     elif mime.startswith('audio/'):
                         return Modality.AUDIO
-            except Exception as e:
-                pass
+            except requests.exceptions.RequestException as e:
```
wanliAlex, line 173: Do you think there will be other errors here? e.g., error from magic, mime, etc

wanliAlex (changes requested):

src/marqo/s2_inference/multimodal_model_load.py
RaynorChavez: added

src/marqo/s2_inference/multimodal_model_load.py
RaynorChavez: done, but with slight modification

farshidz (dismissed):

src/marqo/tensor_search/add_docs.py
```diff
@@ -88,18 +90,23 @@ def threaded_download_and_preprocess_content(allocated_docs: List[dict],
...
-                    if (modality == Modality.IMAGE and is_no_index) or (
-                        modality == Modality.IMAGE and is_unstructured_index) or (
-                        is_structured_index and media_field_types_mapping[field] == FieldType.ImagePointer):
-
-                        if (marqo_index is not None
-                                and marqo_index.model.properties.get('type') in [ModelType.LanguageBind]
-                                and marqo_index.model.properties.get('supported_modalities') is not None
-                                and Modality.IMAGE not in marqo_index.model.properties.get('supported_modalities')):
+                    try:    
+                        inferred_modality = infer_modality(doc[field])
```
wanliAlex, line 94: Doest this mean we are still inferring modality for the image field? We shouldn't do this to avoid regression.

wanliAlex (changes requested):

wanliAlex (approved):

PR #965, opened 2024-09-12
Extract test image urls to marqo_test.py

papa99do:
Test improvements

We use various images in tests. URLs of the image are defined in each individual tests

We consolidate these URLs to a dict maintained in marqo_test.py so that we can change them in a central place in the future.

No

Yes

N/A

N/A

* **Other information**:
N/A

### DIFF
--- a/src/marqo/s2_inference/random_utils.py
+++ b/src/marqo/s2_inference/random_utils.py
@@ -12,8 +12,12 @@ def sentence_to_hash(sentence):
     # for speed reasons we hash
     if isinstance(sentence, ImageType):
         pixel_data = list(sentence.getdata())
-        pixel_averages = [sum(channels)/len(channels) for channels in pixel_data]
-        image_average = functools.reduce(lambda x, y: x + y, pixel_averages)/len(pixel_data)
+        if isinstance(pixel_data[0], tuple):
+            pixel_averages = [sum(channels)/len(channels) for channels in pixel_data]
+            image_average = functools.reduce(lambda x, y: x + y, pixel_averages) / len(pixel_data)
+        else:
+            # uses palette
+            image_average = functools.reduce(lambda x, y: x + y, pixel_data)/len(pixel_data)
         return int(hashlib.sha256(str(image_average).encode('utf-8')).hexdigest(), 16) % 10 ** 8
     else:
         return int(hashlib.sha256(sentence.encode('utf-8')).hexdigest(), 16) % 10**8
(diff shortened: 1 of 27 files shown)


tests/marqo_test.py
```diff
@@ -21,6 +21,17 @@
 from marqo.tensor_search.telemetry import RequestMetricsStore
 from marqo.vespa.vespa_client import VespaClient
 
+TEST_IMAGE_URLS = {
```
farshidz: I think an enum for these keys would be great, or a class and one method to get each URL (or a metaclass that generates a class from the list? Probably overkill)

src/marqo/s2_inference/random_utils.py
```diff
@@ -12,8 +12,12 @@ def sentence_to_hash(sentence):
...
     if isinstance(sentence, ImageType):
         pixel_data = list(sentence.getdata())
-        pixel_averages = [sum(channels)/len(channels) for channels in pixel_data]
-        image_average = functools.reduce(lambda x, y: x + y, pixel_averages)/len(pixel_data)
+        if isinstance(pixel_data[0], tuple):
+            pixel_averages = [sum(channels)/len(channels) for channels in pixel_data]
+            image_average = functools.reduce(lambda x, y: x + y, pixel_averages) / len(pixel_data)
+        else:
+            # uses palette
+            image_average = functools.reduce(lambda x, y: x + y, pixel_data)/len(pixel_data)
```
vicilliar: Just clarifying, why are we using `pixel_data` instead of `pixel_averages` here?

tests/marqo_test.py
vicilliar: Is it redundant to have both an enum for image keys and an object for the URLs? In the tests, we'd just want to fetch the actual URL for downloading, so why don't we just have 1 enum with the URLs? Something like:
```
class TestImageURLs(Enum):
    IMAGE0 = 'https://raw.githubusercontent.com/marqo-ai/marqo/mainline/examples/ImageSearchGuide/data/image0.jpg'
    IMAGE1 = 'https://raw.githubusercontent.com/marqo-ai/marqo/mainline/examples/ImageSearchGuide/data/image1.jpg'
```

vicilliar: In the tests, we could then use `TestImageURLs.IMAGE0` instead of `TEST_IMAGE_URLS[ImageKey.IMAGE0]`

src/marqo/s2_inference/random_utils.py
papa99do: The data of png file using palette is just a one-dimension array. It does not store the RGB data for each pixel, instead it stores an index of the palette. I'll make this more explicit by adding a comment

tests/marqo_test.py
papa99do: Good point. @vicilliar . I'll change that. I'll probably need to use `TestImageURLs.IMAGE0.value` in the test

vicilliar (approved):

PR #968, opened 2024-09-15
Support searchable attributes for unstructured indexes

papa99do:
Feature

Searchable attributes is not supported for unstructured index since we store embeddings of all tensor fields in one field, and all string fields in one vespa field. 

Searchable attributes is for unstructured index. When new documents are added to an unstructured index, new lexical and tensor fields are dynamically added to the index. This allows user to specify searchable attributes to selectively search on certain fields when using tensor, lexical or hybrid search. 

No

Yes

N/A. Client will remain the same

To be provided

* **Other information**:
This PR also include the following changes which are required to add this feature
- A refactoring of the add_document logic
- Extracted add_document method in the base MarqoTestCase. it force refresh the index after adding docs to make sure the following get and search requests always get the latest index settings in all tests.
- Introduced a SemiStructuredMarqoIndex type and corresponding VespaIndex implementation. This type of index will be created when an UnstructuredMarqoIndexRequest is received going forward. The UnstructuredMarqoIndex will be kept so unstructured index created before this release can still work as is. 


src/marqo/core/document/add_documents_handler.py
```diff
@@ -0,0 +1,562 @@
...
+
+    def to_add_doc_responses(self, index_name: str) -> MarqoAddDocumentsResponse:
+        processing_time = (timer() - self.start_time) * 1000
+        # since we reversed the doc list to skip duplicate docs, we now need to reverse the response
+        sorted_responses = [response for _, response in sorted(self.responses, key=lambda r: r[0] or 0, reverse=True)]
+        return MarqoAddDocumentsResponse(errors=self.errors, index_name=index_name, items=sorted_responses,
+                                         processingTimeMs=processing_time)
+
+
+class AddDocumentsHandler(ABC):
```
farshidz: Ideally we want one core class Document that provides all functionality. I find the abstractions a bit confusing at this point as `Document` handles partial updates. Perhaps we could have these handlers for add docs, update etc and have `Document` as a facade that everyone else calls?

src/marqo/core/document/add_documents_handler.py
```diff
@@ -0,0 +1,562 @@
...
+        Template method for adding documents to Marqo index
+        """
+        with RequestMetricsStore.for_request().time("add_documents.processing_before_vespa"):
+            for loc, original_doc in enumerate(reversed(self.add_docs_params.docs)):
+                doc = copy.deepcopy(original_doc)
+                original_id = None
+                try:
+                    self.validate_doc(doc)
+
+                    original_id = self.validate_and_pop_doc_id(doc)
```
farshidz: This kind of mutation is generally risky and not very readable. Best to have pure functions

src/marqo/core/document/add_documents_handler.py
```diff
@@ -0,0 +1,562 @@
...
+
+        with RequestMetricsStore.for_request().time("add_documents.postprocess"):
+            self.handle_vespa_response(response)
+            return self.add_docs_response_collector.to_add_doc_responses(self.marqo_index.name)
+
+    @abstractmethod
+    def _create_tensor_fields_container(self) -> TensorFieldsContainer:
+        pass
+
+    @abstractmethod
```
farshidz: Can we have docstrings for these methods so it's clear what the contract is and what the abstraction means?

src/marqo/core/document/add_documents_handler.py
```diff
@@ -0,0 +1,562 @@
...
+                    self.validate_doc(doc)
+
+                    original_id = self.validate_and_pop_doc_id(doc)
+                    doc_id = original_id or str(uuid.uuid4())
+                    marqo_doc = {ORIGINAL_ID: original_id, MARQO_DOC_ID: doc_id}  # keep this info for error report
+
+                    for field_name, field_content in doc.items():
+                        self.handle_field(marqo_doc, field_name, field_content)
+
+                    self.handle_multi_modal_fields(marqo_doc)
```
farshidz: I think these methods should have a _ prefix as they're not to be accessed outside of the class

src/marqo/core/document/add_documents_handler.py
```diff
@@ -0,0 +1,548 @@
...
+                except AddDocumentsError as err:
+                    self.add_docs_response_collector.collect_error_response(original_id, err, loc)
+
+            # retrieve existing docs for existing tensor
+            if self.add_docs_params.use_existing_tensors:
+                # TODO capture the telemetry data for retrieving exiting docs?
+                result = self.config.vespa_client.get_batch(list(self.add_docs_response_collector.visited_doc_ids),
+                                                            self.marqo_index.schema_name)
+                existing_vespa_docs = [r.document for r in result.responses if r.status == 200]
+                self.handle_existing_tensors(existing_vespa_docs)
```
farshidz: Handle is very generic. Perhaps you could have more descriptive names?

src/marqo/core/structured_vespa_index/structured_add_document_handler.py
```diff
@@ -0,0 +1,128 @@
...
+from marqo.core.structured_vespa_index.structured_vespa_index import StructuredVespaIndex
+
+from marqo.vespa.models import VespaDocument
+from marqo.vespa.models.get_document_response import Document
+
+# TODO deps to tensor_search needs to be removed
+from marqo.tensor_search import validation
+
+
+class StructuredAddDocumentsHandler(AddDocumentsHandler):
```
farshidz, line 21: This module inheriting from `core.document` seems problematic. The idea originally was for this module to be low level and not depend on `core.document`. Though I don't have a better suggestion right now, just pointing this out

src/marqo/core/document/add_documents_handler.py
papa99do: makes sense. I'll do validation in one method, and pop id in the next step.

src/marqo/core/document/add_documents_handler.py
papa99do: Will do

src/marqo/core/document/add_documents_handler.py
papa99do: I'm not too sure about python convention for this. usually underscore prefixed method means private method or protected one. Since these method are only used in the class hierarchy and not by other classes. I think it makes sense to have a `_` prefix. I'll make the change

src/marqo/core/document/add_documents_handler.py
papa99do: I think adding documents is quite complex, it makes sense to have its own class.  I like your idea to make Document as a facade to call this handler. (I'll move the add_documents method in tensor_search in Documents at the end of refactoring) For other logic like deletion and partial update, I'm not sure if it worth having their own handler. If the logic is generic, we won't need a class hierarchy to capture index type specific logic, then it makes sense to keep it in Documents class.

src/marqo/core/document/add_documents_handler.py
```diff
@@ -0,0 +1,583 @@
...
+        """
+        Download, preprocess, chunk and vectorise collected tensor fields.
+        Three different batching strategies can be chosen to do tradeoff between resource usage and performance.
+        - Batching by field: Chunk and vectorise field by field (Default?)
+        - Batching by doc: Chunk and vectorise fields of a doc by field type (text, image, audio, video, etc.)
+        - Batching by add-doc batch: Chunk and vectorise all fields of a batch of docs by type
+        """
+        # TODO add a parameter to choose batching strategy? Do a performance test to decide the default approach
+        # self.vectorise_tensor_fields_per_field()
+        self.vectorise_tensor_fields_in_batch_per_doc()
```
farshidz: What's the point of this if it's calling another function in this class?

src/marqo/core/document/tensor_fields_container.py
```diff
@@ -0,0 +1,342 @@
...
+
+    def tensor_fields_to_vectorise(self, *types: FieldType) -> Generator[str, str, TensorFieldContent]:
+        for doc_id in list(self._tensor_field_map.keys()):
+            fields = self._tensor_field_map[doc_id]
+            for field_name, tensor_field_content in fields.items():
+                if doc_id not in self._tensor_field_map:
+                    # removed during interation due to error handling
+                    break
+
+                if tensor_field_content.field_type not in types:
```
farshidz: When does this happen? Should it be an error?

src/marqo/core/document/tensor_fields_container.py
```diff
@@ -0,0 +1,342 @@
...
+        if doc_id in self._tensor_field_map:
+            del self._tensor_field_map[doc_id]
+
+    def _add_tensor_field_content(self, doc_id: str, field_name: str, content: TensorFieldContent) -> None:
+        if doc_id not in self._tensor_field_map:
+            self._tensor_field_map[doc_id] = dict()
+        self._tensor_field_map[doc_id][field_name] = content
+
+    def tensor_fields_to_vectorise(self, *types: FieldType) -> Generator[str, str, TensorFieldContent]:
+        for doc_id in list(self._tensor_field_map.keys()):
```
farshidz: Passing these in the function call (pure function) makes your function more readable and reusable, though I admit it could make the code cluttered at times

src/marqo/core/document/tensor_fields_container.py
```diff
@@ -0,0 +1,342 @@
...
+            # not tensor fields, no need to collect
+            return field_content
+
+        if self.is_custom_tensor_field(field_name):
+            content = field_content['content']
+            embedding = field_content['vector']
+            tensor_field_content = TensorFieldContent(
+                field_content=content,
+                field_type=FieldType.CustomVector,
+                is_tensor_field=True,
```
farshidz: When is this flag false?

src/marqo/core/semi_structured_vespa_index/semi_structured_add_document_handler.py
```diff
@@ -0,0 +1,74 @@
...
+        self._validate_field(field_name, field_content)
+        text_field_type = self._infer_field_type(field_content)
+        content = self.tensor_fields_container.collect(marqo_doc[MARQO_DOC_ID], field_name,
+                                                       field_content, text_field_type)
+        marqo_doc[field_name] = content
+
+        if isinstance(content, str):
+            # Add missing lexical fields to marqo index
+            if field_name not in self.marqo_index.field_map:
+                self.marqo_index.lexical_fields.append(
```
farshidz: This logic feels out of place here. Is there a better place? Do we need a different abstraction?

src/marqo/core/semi_structured_vespa_index/semi_structured_add_document_handler.py
```diff
@@ -0,0 +1,74 @@
...
+                    Field(name=field_name, type=FieldType.Text,
```
farshidz: Do we have a configurable limit on max tensor and lexical fields?

src/marqo/core/document/tensor_fields_container.py
papa99do: subfields of the Multimodal fields that are not top level tensor fields

src/marqo/core/semi_structured_vespa_index/semi_structured_add_document_handler.py
papa99do: No we don't have one yet. what's in your mind would be a good limit?

src/marqo/core/semi_structured_vespa_index/semi_structured_add_document_handler.py
papa99do: I can create a collector class to collect fields to add to the index, but this is probably an overkill given the logic is simple and straightforward here. it's in the handle_field method since we collect the missing field along the way when we process each field of each doc.

src/marqo/core/document/tensor_fields_container.py
papa99do: could you clarify a bit about what should be passed in in the function call?

src/marqo/core/document/tensor_fields_container.py
papa99do: we traverse all tensor fields of a doc here, this line will filter out fields with type that's not in the types we pass in. For example, if we want to traverse ImagePointer field only, all other types will be filtered out.

src/marqo/core/document/add_documents_handler.py
papa99do: This is an incomplete change. My intension is to provide several different batching strategies, and default to one of them. The complete code will be something like this. I'm still working on it. Keen to know your thoughts
```
    def vectorise_tensor_fields(self) -> None:
        """
        Download, preprocess, chunk and vectorise collected tensor fields.
        Three different batching strategies can be chosen to do tradeoff between resource usage and performance.
        - Batching by field: Chunk and vectorise field by field (Default?)
        - Batching by doc: Chunk and vectorise fields of a doc by field type (text, image, audio, video, etc.)
        - Batching by add-doc batch: Chunk and vectorise all fields of a batch of docs by type
        """
        model_config = ModelConfig(
            model_name=self.marqo_index.model.name,
            model_properties=self.marqo_index.model.get_properties(),
            model_auth=self.add_docs_params.model_auth,
            device=self.add_docs_params.device,
            normalize_embeddings=self.marqo_index.normalize_embeddings
        )
        # TODO add a parameter to choose batching strategy? Do a performance test to decide the default approach
        batch_mode = 'PER_DOC'  # extract a enum

        if batch_mode == 'PER_FIELD':
            self.vectorise_tensor_fields_per_field(model_config)
        elif batch_mode == 'PER_DOC':
            self.vectorise_tensor_fields_in_batch_per_doc(model_config)
        elif batch_mode == 'PER_ADD_DOC_BATCH':
            self.vectorise_tensor_fields_in_batch_per_add_doc_batch(model_config)
        else:
            raise api_errors.BadRequestError(
                message=f'Unsupported vectorisation batch mode: {str(batch_mode)}'
            )
```

src/marqo/core/structured_vespa_index/structured_add_document_handler.py
papa99do: That's a good point. In that case, I'll probably need to move the add_documents_handler.py somewhere out of documents. or maybe move all the different types of handlers in to documents. I'll need to give it another thought.

src/marqo/core/document/add_documents_handler.py
papa99do: populate_existing_tensors?

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_schema_template.sd.jinja2
```diff
@@ -0,0 +1,253 @@
...
+            {%- for field in fields -%}{{ add() }}{{ score_macro(field) }}{%- endfor -%}
+        {% endmacro -%}
+
+        {% if index.lexical_fields -%}
+        function lexical_score_sum() {
+            expression: {{ sum_score(lexical_score, index.lexical_fields) }}
+        }
+
+        function lexical_score_avg() {
+            {# TODO calculate average lexical score -#}
```
papa99do: @farshidz , Joshua and I had a discussion regarding this function for structured index here: https://marqo-ai.slack.com/archives/C05K6BFQUTF/p1726537136109449?thread_ts=1726530564.502089&cid=C05K6BFQUTF . 
Could you shed some lights on why we calculate average in this way? Thanks. 
BTW, I just leave it as 0 for now since we only use lexical_score_sum() right now.

src/marqo/core/structured_vespa_index/structured_add_document_handler.py
papa99do: As discussed, I've moved add_documents_handler.py to the new `vespa_index` module, along with `vespa_index.py` and `vespa_schema.py`. I have moved `tensor_fields_container.py` to `core.inference` module because I cannot find a better place to put it. Any suggestions would be appreciated.

src/marqo/vespa/vespa_client.py
```diff
@@ -524,6 +524,36 @@ def get_metrics(self) -> ApplicationMetrics:
 
         return ApplicationMetrics(**resp.json())
 
+    def translate_vespa_document_response(self, status: int, message: Optional[str]=None) -> Tuple[int, Optional[str]]:
```
papa99do, line 574: This piece of code is in document.py. But we need it in add_document_handler, which is a dependency of document.py. To avoid cyclic deps, I moved it here, which seems to be a good place since it converts Vespa status code to Marqo status code for individual docs. If we decided to leave it here, I'll delete the original code from document.py.

tests/conftest.py
```diff
@@ -20,4 +27,10 @@ def pytest_collection_modifyitems(config, items):
     else:
         for item in items:
             if "largemodel" in item.keywords:
-                item.add_marker(skip_largemodel)
\ No newline at end of file
+                item.add_marker(skip_largemodel)
+
+    if config.getoption("--legacy-unstructured"):
```
papa99do: This is my attempt to run integration test suite on old structured index. This kinda works, but it takes about 40min and it also runs tests for structured indexes which are not needed. I will probably revert this commit and find another way to test legacy unstructured index (maybe using compatibility test suite in the future?)

tests/core/document/test_partial_document_update.py
```diff
@@ -133,7 +133,7 @@ def _set_up_for_text_field_test(self):
             "text_field_tensor": "text field tensor",
             "_id": "1"
         }
-        tensor_search.add_documents(config=self.config, add_docs_params=AddDocsParams(
+        self.add_documents(config=self.config, add_docs_params=AddDocsParams(
```
papa99do: I have extracted an add_documents method in the base class in marqo_test.py.  It will force refresh the index after adding documents. This is to guarantee the following search request can get the latest index settings so the integration tests can pass. 
In the following changes, I'll delete the add_documents method from tensor_search module, and use the one I moved to core/document module

src/marqo/core/document/tensor_fields_container.py
farshidz: What's the use case for this? Do we ever process type grouped by type?

src/marqo/core/semi_structured_vespa_index/semi_structured_add_document_handler.py
farshidz: I think we need some generous limit to catch the odd user who is creating fields in a loop. Perhaps 10 each? Or 20

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_schema_template.sd.jinja2
farshidz: This seems like new code. Looking at existing tests the average function looks valid?

src/marqo/core/semi_structured_vespa_index/semi_structured_add_document_handler.py
papa99do: Do you mean we need a limit for each batch or overall? For each batch, it should be fine, since it takes about the same time to add 1 field and 10 fields. We haven't done the experiment to check the max tensor fields Vespa supports, so we will need to pick a random number as an overall limit (or maybe leave it as unlimited to see until we have some extreme use cases which breaks Vespa).

papa99do: Here's the result of my local test:
- adding 2 fields, 'add_documents.update_index': 2159.5356249999895
- adding 10 fields, 2 at a time, first: 2s, following: 6s, 5s, 7s, 5s (there seem to be a cooling down period in Vespa convergence, if adding a 5s sleep in-between, time will reduce to 2s)
- adding 10 fields, 'add_documents.update_index': 2061.9370840000784
- adding 20 fields, 'add_documents.update_index': 1923.3983750000334
- adding 50 fields, 'add_documents.update_index': 2128.181333000043

tests/core/document/test_partial_document_update.py
papa99do: rename the method to reflect the intention to refresh cache

src/marqo/core/semi_structured_vespa_index/semi_structured_add_document_handler.py
papa99do: Added the following two env vars with default value set to 100
`MARQO_MAX_TENSOR_FIELD_COUNT_UNSTRUCTURED`
`MARQO_MAX_LEXICAL_FIELD_COUNT_UNSTRUCTURED`

src/marqo/core/index_management/index_management.py
```diff
@@ -273,6 +274,16 @@ def _get_vespa_application(self, check_configured: bool = True, need_binary_file
...
         """
         vespa_version = semver.VersionInfo.parse(self.vespa_client.get_vespa_version())
+        if vespa_version < self._MINIMUM_VESPA_VERSION_TO_SUPPORT_FAST_FILE_DISTRIBUTION:
+            # Please note that this warning message will only be logged out for OS users running Marqo on external
+            # Vespa servers with version prior to 8.396.18. This will be displayed when Marqo starts up and before
+            # each index CUD operation
+            logger.warning(f'You may encounter slowness when creating a Marqo index or adding documents to indexes '
+                           f'with the current Vespa version {vespa_version}. To improve the performance, please upgrade '
+                           f'Vespa to version {self._MINIMUM_VESPA_VERSION_TO_SUPPORT_FAST_FILE_DISTRIBUTION} or '
+                           f'higher and set the VESPA_FILE_DOWNLOAD_BACKOFF_INITIAL_TIME_MS environment variable to '
```
farshidz: Let's move this to the docs and be clear that it's a Vespa configuration

src/marqo/core/semi_structured_vespa_index/semi_structured_add_document_handler.py
```diff
@@ -67,3 +54,46 @@ def _pre_persist_to_vespa(self):
             with RequestMetricsStore.for_request().time("add_documents.update_index"):
                 self.index_management.update_index(self.marqo_index)
 
+    def _add_lexical_field_to_index(self, field_name):
+        if field_name in self.marqo_index.field_map:
+            return
+
+        max_lexical_field_count = utils.read_env_vars_and_defaults_ints(
```
farshidz: Limits should be passed in the constructor not read within the class

src/marqo/core/index_management/index_management.py
```diff
@@ -277,20 +277,22 @@ def _get_vespa_application(self, check_configured: bool = True, need_binary_file
...
             # Please note that this warning message will only be logged out for OS users running Marqo on external
-            # Vespa servers with version prior to 8.382.22.
-            logger.warning(f'You may run into race conditions when Marqo bootstraps the Vespa application package '
-                           f'with the current Vespa version {vespa_version}. To avoid any changes to the Marqo indexes '
-                           f'from being accidentally overridden, please upgrade '
+            # Vespa servers with version prior to 8.382.22. This will be displayed when Marqo starts up and before
+            # each index CUD operation
+            logger.warning(f'Your Vespa version {vespa_version} is lower than the minimum recommended Vespa version '
+                           f'{self._MINIMUM_VESPA_VERSION_TO_SUPPORT_FAST_FILE_DISTRIBUTION}. This could cause '
+                           f'unexpected unexpected behavior when bootstrapping Marqo. Please upgrade '
```
farshidz: unexpected unexpected

farshidz (changes requested):

farshidz (approved):

PR #970, opened 2024-09-17
Normalize custom vectors passed in at the time of adding documents

adityabharadwaj198:
feature

Currently, if custom vectors are passed as part of the document adding process - we do not normalize them even if normalizeEmbeddings = True.
Also if normalizeEmbeddings = True, during search - there could be a case when division by zero happens, so we throw a ZeroMagnitudeVectorError in that case

1. After this change, any custom vectors passed during document addition will be normalized if normalizeEmbeddings = True 
2. At the time of search, if the query vector has a weight zero, or if the context vector is a zero vector or if the context vector has a weight of zero - then a division by zero may occur and we will throw a ZeroMagnitudeVectorError back to user. 

No

I've tested using pycharm debugger and manually verifying that the vector is normalised

none

none

* **Other information**:


src/marqo/tensor_search/tensor_search.py
```diff
@@ -2648,3 +2653,14 @@ def delete_documents(config: Config, index_name: str, doc_ids: List[str]):
             document_ids=doc_ids,
         )
     )
+
+def normalize_custom_vector(embeddings):
```
vicilliar: Maybe this could be renamed to `normalize_vector` and even used for our query vector normalization: https://github.com/marqo-ai/marqo/blob/662af89f270a51c1706efb3c4643a1284cd12f07/src/marqo/tensor_search/tensor_search.py#L1909

So we don't repeat code. Iirc this normalization only happens in these 2 places in tensor search.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -293,6 +294,10 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
                     # Generate exactly 1 chunk with the custom vector.
                     chunks = [f"{field}::{copied[field]['content']}"]
                     embeddings = [copied[field]["vector"]]
+                    # If normalize_embeddings is true and the index version is > 2.12.0, normalize the embeddings.
+                    # We have added version specific check here to prevent backwards compatibility issues.
+                    if marqo_index.normalize_embeddings and marqo_index.parsed_marqo_version() > semver.VersionInfo.parse("2.12.0"):
```
vicilliar: We could store this version `semver.VersionInfo.parse("2.12.0")` as a property of the index class instead of hardcoding it. Something like how it was done for Hybrid Search: https://github.com/marqo-ai/marqo/blob/7264b9c4532cdacad7f4474c88f25761bf374598/src/marqo/core/structured_vespa_index/structured_vespa_index.py#L62

src/marqo/tensor_search/tensor_search.py
```diff
@@ -2648,3 +2653,14 @@ def delete_documents(config: Config, index_name: str, doc_ids: List[str]):
...
         )
     )
+
+def normalize_custom_vector(embeddings):
+    embeddings_array = np.array(embeddings)
+    magnitude = np.linalg.norm(embeddings_array)
+    if magnitude != 0:
+        embeddings_array = embeddings_array / magnitude
+    else:
+        embeddings_array = embeddings_array
```
vicilliar: In what situations would magnitude be 0? It might be worth testing this edge case.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -2648,3 +2653,14 @@ def delete_documents(config: Config, index_name: str, doc_ids: List[str]):
...
+
+    return embeddings_array.tolist()
+
```
vicilliar, line 2721: We should have some unit tests to make sure the vectors are being normalized properly

src/marqo/tensor_search/tensor_search.py
papa99do: Does this apply to structured index as well? Also can we add a test case?

src/marqo/tensor_search/tensor_search.py
vicilliar: Try to put a parameter and return value datatype, just so it's easier to maintain in the future. In this case, embeddings would be a 2d array like `[[0.1, 0.2, 0.3 ...]]`, we could even make the input just accept the 1d array `[0.1, 0.2, 0.3 ...]` then just wrap it in another list outside the method, so the method is easier to use.

vicilliar (changes requested):
Great work! Just left a few comments :)

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: done

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: It can be zero if the embeddings list passed originally to this function contains a list of 0s. Let me add a test case for this

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: Added same logic for structured index, also added a test case

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: Renamed, and reused for query vector normalization

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: Added a test case with zero vector

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: Added unit tests

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: Added return type & input parameter type

tests/tensor_search/integ_tests/test_custom_vector_field.py
```diff
@@ -235,8 +235,6 @@ def run():
                 self.assertEqual(vespa_fields["marqo__embeddings"], {"0": self.random_vector_1})
 
             elif isinstance(index, StructuredMarqoIndex):
-                print("Printing embeddings" + vespa_fields["marqo__embeddings_my_custom_vector"]["0"].__str__())
-                print("Printing embeddings2" + self.random_vector_1.__str__())
```
adityabharadwaj198: Mistakenly committed this file - reverted it in https://github.com/marqo-ai/marqo/pull/970/commits/420a3678f628a7a93e1de8279e41190ce3c9634e

src/marqo/tensor_search/tensor_search.py
```diff
@@ -2648,3 +2657,33 @@ def delete_documents(config: Config, index_name: str, doc_ids: List[str]):
             document_ids=doc_ids,
         )
     )
+
+def normalize_vector(embeddings: Union[List[List[float]], ndarray]) -> List[List[float]]:
```
farshidz: Was there no existing normalization code? We already normalize for non-custom vectors

tests/tensor_search/integ_tests/test_custom_vector_field.py
```diff
@@ -84,13 +91,73 @@ def setUpClass(cls) -> None:
                            "text_field", "my_multimodal"]
         )
 
+        structured_custom_index_with_normalize_embeddings_true = cls.structured_marqo_index_request(
```
farshidz: custom_index isn't a term we use here. It's a custom vector so the index name is confusing

src/marqo/core/constants.py
```diff
@@ -14,6 +14,7 @@
 
 MARQO_STRUCTURED_HYBRID_SEARCH_MINIMUM_VERSION = '2.10.0'
 MARQO_UNSTRUCTURED_HYBRID_SEARCH_MINIMUM_VERSION = '2.11.0'
+MARQO_CUSTOM_VECTOR_NORMALIZATION_MINIMUM_VERSION = '2.12.0'
```
farshidz: This should be 2.13

src/marqo/tensor_search/tensor_search.py
```diff
@@ -293,6 +295,10 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
                     # Generate exactly 1 chunk with the custom vector.
                     chunks = [f"{field}::{copied[field]['content']}"]
                     embeddings = [copied[field]["vector"]]
+                    # If normalize_embeddings is true and the index version is > 2.12.0, normalize the embeddings.
+                    # We have added version specific check here to prevent backwards compatibility issues.
+                    if marqo_index.normalize_embeddings and marqo_index.parsed_marqo_version() > semver.VersionInfo.parse(constants.MARQO_CUSTOM_VECTOR_NORMALIZATION_MINIMUM_VERSION):
```
farshidz: Better to parse and store as a class variable than to parse every time

src/marqo/tensor_search/tensor_search.py
```diff
@@ -2648,3 +2657,33 @@ def delete_documents(config: Config, index_name: str, doc_ids: List[str]):
...
+
+    Args:
+        embeddings (Union[List[List[float]], ndarray]): A list of vectors or a numpy ndarray of vectors to be normalized.
+
+    Returns:
+        List[List[float]]: A list of normalized vectors.
+    """
+
+    # Convert the input embeddings to a numpy array
+    if embeddings.__class__ == ndarray:
```
farshidz, line 2705: Why do we need to handle ndarray type?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -2648,3 +2657,33 @@ def delete_documents(config: Config, index_name: str, doc_ids: List[str]):
...
+        embeddings_array = np.array(embeddings)
+
+    # Calculate the magnitude (Euclidean norm) of each vector along the last axis
+    magnitude = np.linalg.norm(embeddings_array, axis = -1, keepdims=True)
+
+    # Normalize each vector by dividing by its magnitude, handle zero magnitude case
+    if magnitude != 0:
+        embeddings_array = embeddings_array / magnitude
+    else:
+        embeddings_array = embeddings_array # Handle the zero vector case
```
farshidz: This looks like it should be a ValueError here and a 400 in the API response. Do take care not to fail the full add docs batch, but only the individual document with this vector

farshidz (changes requested):

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: There was no existing function that could be reused to normalize vector, since normalization is being done in 3 places (add docs - structured, add docs - unstructured & query docs) I created a function for reusability purposes.

tests/tensor_search/integ_tests/test_custom_vector_field.py
adityabharadwaj198: I can change this - but this term was already being used in the test cases, ex: https://github.com/marqo-ai/marqo/blob/420a3678f628a7a93e1de8279e41190ce3c9634e/tests/tensor_search/integ_tests/test_custom_vector_field.py#L24

src/marqo/core/constants.py
adityabharadwaj198: Ah, I thought we were releasing this fix in versions post 2.12.0, let me know if not and I can change this

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: makes sense, I'll change this.

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: we're using this function in 3 places, out of which 1 passes a ndarray (here: https://github.com/marqo-ai/marqo/blob/c0d09826a6c662fb96501f3e9f429cde94ba6a18/src/marqo/tensor_search/tensor_search.py#L1917-L1920) 

`merged_vector = np.mean(weighted_vectors, axis=0)` will convert it to an ndarray, which will then be passed to normalize_vector where we don't need to convert it to ndarray again

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: Our earlier logic was not to raise a ValueError/400 in case magnitude was zero: https://github.com/marqo-ai/marqo/blob/mainline/src/marqo/tensor_search/tensor_search.py#L1908-L1911, are we sure we want to throw an error here?

src/marqo/core/constants.py
farshidz: 2.12 is already out and this doesn't qualify for a patch release so I think 2.13 is best (next week)

src/marqo/tensor_search/tensor_search.py
farshidz: But also https://github.com/marqo-ai/marqo/blob/mainline/src/marqo/core/utils/vector_interpolation.py#L171

src/marqo/core/constants.py
adityabharadwaj198: Changed as per offline discussion, here: https://github.com/marqo-ai/marqo/pull/970/files#diff-d562ff1ee5c3bc3e20805164a98160b80530cfca5b9a56d8d3020a6c5e575518R19

tests/tensor_search/integ_tests/test_custom_vector_field.py
adityabharadwaj198: changed as per offline discussion

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: Added logic to throw a new ZeroMagnitudeVectorError: https://github.com/marqo-ai/marqo/pull/970#diff-bc6ddb28e3809766b86804afac01ac632de1e691377c5d3f43d0333b97ba79faR303
https://github.com/marqo-ai/marqo/pull/970#diff-bc6ddb28e3809766b86804afac01ac632de1e691377c5d3f43d0333b97ba79faR848
https://github.com/marqo-ai/marqo/pull/970#diff-bc6ddb28e3809766b86804afac01ac632de1e691377c5d3f43d0333b97ba79faR1950

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: Changed: https://github.com/marqo-ai/marqo/pull/970#diff-d562ff1ee5c3bc3e20805164a98160b80530cfca5b9a56d8d3020a6c5e575518R17-R19

tests/tensor_search/integ_tests/test_custom_vector_field.py
adityabharadwaj198: This variable was no longer needed after some refactoring so removed it

src/marqo/tensor_search/tensor_search.py
papa99do: this is where the multimodal combo field embedding is calculated. According to Owen and Jesse, although it's quite rare, but it is possible that the norm is 0. So it's also good to call normalize_vector method here. Thanks for fixing this.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -2648,3 +2691,33 @@ def delete_documents(config: Config, index_name: str, doc_ids: List[str]):
...
+        embeddings_array = np.array(embeddings)
+
+    # Calculate the magnitude (Euclidean norm) of each vector along the last axis
+    magnitude = np.linalg.norm(embeddings_array, axis = -1, keepdims=True)
+
+    # Normalize each vector by dividing by its magnitude, handle zero magnitude case
+    if magnitude != 0:
+        embeddings_array = embeddings_array / magnitude
+    else:
+        raise core_exceptions.ZeroMagnitudeVectorError(f"Zero magnitude vector detected, cannot normalize.")
```
papa99do, line 2717: Unlike custom vector field, which is provided by user, multimodal combo field embedding is calculated based on embeddings of different fields. If somehow the manitude is zero, there's no easy way for user to fix it. It might not be a good user experience if we just return this error to user. However, this might be too rare to be a real concern for now. @farshidz , what's your thought?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -293,6 +295,23 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
                     # Generate exactly 1 chunk with the custom vector.
                     chunks = [f"{field}::{copied[field]['content']}"]
                     embeddings = [copied[field]["vector"]]
+                    # If normalize_embeddings is true and the index version is > 2.13.0, normalize the embeddings.
+                    # We have added version specific check here to prevent backwards compatibility issues.
+                    if marqo_index.normalize_embeddings and marqo_index.parsed_marqo_version() > MARQO_CUSTOM_VECTOR_NORMALIZATION_MINIMUM_VERSION:
```
vicilliar: Shouldn't this be `>=`, not `>`?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -821,6 +840,24 @@ def _add_documents_structured(config: Config, add_docs_params: AddDocsParams, ma
                     chunks = [copied[field]['content']]
                     embeddings = [copied[field]["vector"]]
 
+                    # If normalize_embeddings is true and the index version is > 2.13.0, normalize the embeddings.
+                    # We have added version specific check here to prevent backwards compatibility issues.
+                    if marqo_index.normalize_embeddings and marqo_index.parsed_marqo_version() > MARQO_CUSTOM_VECTOR_NORMALIZATION_MINIMUM_VERSION:
```
vicilliar: Same comment, shouldn't this be `>=`?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1906,9 +1943,15 @@ def get_query_vectors_from_jobs(
...
             if q.index.normalize_embeddings:
-                norm = np.linalg.norm(merged_vector, axis=-1, keepdims=True)
-                if norm > 0:
-                    merged_vector /= np.linalg.norm(merged_vector, axis=-1, keepdims=True)
+                try:
+                    merged_vector = normalize_vector(merged_vector)
+                except core_exceptions.ZeroMagnitudeVectorError as e:
+                    tb = traceback.extract_tb(e.__traceback__)[-1]
+                    raise core_exceptions.ZeroMagnitudeVectorError(
+                        f"Zero magnitude vector detected at file name: {tb.filename}, line number: {tb.lineno}, inside function name: {tb.name}. "
```
vicilliar: `Zero magnitude vector detected at file name:` isn't a very intuitive error message I think. It could be something like `Magnitude of combined query and context vectors cannot be zero...`, then The filename and other details can be afterwards.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -2648,3 +2691,33 @@ def delete_documents(config: Config, index_name: str, doc_ids: List[str]):
             document_ids=doc_ids,
         )
     )
+
+def normalize_vector(embeddings: Union[List[List[float]], ndarray]) -> List[List[float]]:
```
vicilliar: nit: there might be a need in the future for us to accept `List[float]` as a parameter here as well.

vicilliar: I'd recommend moving this out of the `tensor_search.py` file. We're trying to improve our repo structure and reduce the size of the files. Maybe it could be under `core/utils` or `core/inference`? What do you think @papa99do

papa99do: we can keep it here for now. I'll move it out in my refactoring PR after this is merged.

Edit: I just realised that the logic might not be able to be extracted in my PR given I didn't refactor search code. But I think given the logic here is minimal, it's actually OK if we duplicate it. I might need to raise another error in add_doc anyways. Will see what I could do.

src/marqo/tensor_search/tensor_search.py
farshidz: How are the tests passing then?

src/marqo/tensor_search/tensor_search.py
farshidz: Agreed. The message must be user friendly as it goes to the api. api exception handler will log the stack trace so don't worry about line number etc

src/marqo/tensor_search/tensor_search.py
farshidz: Yeah feels like a 400 still

src/marqo/tensor_search/tensor_search.py
```diff
@@ -2648,3 +2691,33 @@ def delete_documents(config: Config, index_name: str, doc_ids: List[str]):
...
+    """
+
+    # Convert the input embeddings to a numpy array
+    if embeddings.__class__ == ndarray:
+        embeddings_array = embeddings
+    else:
+        embeddings_array = np.array(embeddings)
+
+    # Calculate the magnitude (Euclidean norm) of each vector along the last axis
+    magnitude = np.linalg.norm(embeddings_array, axis = -1, keepdims=True)
```
farshidz, line 2711: Why not reuse the normalization code?

farshidz (commented):
Note API tests also needed

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: Changed this to "Magnitude of combined query and context vectors cannot be zero, If you want to pass a zero vector, please set normalizeEmbeddings = False during index creation"

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: added

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: Not sure what you mean here

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: If you mean the tests written by me, I manually change the MARQO_CUSTOM_VECTOR_NORMALIZATION_MINIMUM_VERSION to something like 2.11.0 and run the tests and make sure they work, then change the value back to 2.13.0 before I commit and push

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: Checked offline with @farshidz and confirmed that this is not applicable

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: changed

adityabharadwaj198:
> Note API tests also needed

@farshidz do you mean tests for https://github.com/marqo-ai/marqo/blob/0a8384e8ef1d26731848ad5cb69de8f6b7d0a585/src/marqo/tensor_search/api.py#L309-L324? I don't think they exist as of now. If they're required - I can probably pick them up as a follow up task?

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: Discussed offline with @farshidz , will handle multimodal combo fields in a follow up PR, tracking Jira ticket creation here: https://s2search.atlassian.net/jira/polaris/projects/MOSD/ideas/view/4863474?selectedIssue=MOSD-73&issueViewLayout=sidebar&issueViewSection=comments&focusedCommentId=10911
Marking this as resolved for now

src/marqo/tensor_search/tensor_search.py
farshidz: I agree should be >= and min should be 2.13 so this will be released in 2.13

adityabharadwaj198: Changed this, also discussed offline with @farshidz

src/marqo/core/search/hybrid_search.py
```diff
@@ -87,17 +87,17 @@ def search(
         # Version checks (different for structured and unstructured)
         marqo_index_version = marqo_index.parsed_marqo_version()
         if isinstance(marqo_index, StructuredMarqoIndex) and \
-                marqo_index_version < semver.VersionInfo.parse(constants.MARQO_STRUCTURED_HYBRID_SEARCH_MINIMUM_VERSION):
+                marqo_index_version < constants.MARQO_STRUCTURED_HYBRID_SEARCH_MINIMUM_VERSION:
             raise core_exceptions.UnsupportedFeatureError(
                 f"Hybrid search is only supported for Marqo structured indexes created with Marqo "
-                f"{constants.MARQO_STRUCTURED_HYBRID_SEARCH_MINIMUM_VERSION} or later. "
+                f"{constants.MARQO_STRUCTURED_HYBRID_SEARCH_MINIMUM_VERSION.__str__()} or later. "
```
farshidz: I think you should use `str(...)`

src/marqo/core/constants.py
```diff
@@ -12,8 +14,9 @@
 MARQO_DOC_HYBRID_TENSOR_SCORE = '_tensor_score'
 MARQO_DOC_HYBRID_LEXICAL_SCORE = '_lexical_score'
 
-MARQO_STRUCTURED_HYBRID_SEARCH_MINIMUM_VERSION = '2.10.0'
-MARQO_UNSTRUCTURED_HYBRID_SEARCH_MINIMUM_VERSION = '2.11.0'
+MARQO_STRUCTURED_HYBRID_SEARCH_MINIMUM_VERSION = semver.VersionInfo.parse('2.10.0')
```
farshidz, line 17: I actually meant to create these in the class using them as a class variable. But this is probably better. Just make sure there's no  regression by finding and reviewing all usages of these

farshidz (changes requested):
Need to upgrade Marqo version to 2.13 as well otherwise your tests won't pass

src/marqo/core/constants.py
adityabharadwaj198: Sure Farshid, I've made sure to review all the usages & made corresponding changes

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1946,11 +1946,9 @@ def get_query_vectors_from_jobs(
                 try:
                     merged_vector = normalize_vector(merged_vector)
                 except core_exceptions.ZeroMagnitudeVectorError as e:
-                    tb = traceback.extract_tb(e.__traceback__)[-1]
                     raise core_exceptions.ZeroMagnitudeVectorError(
-                        f"Zero magnitude vector detected at file name: {tb.filename}, line number: {tb.lineno}, inside function name: {tb.name}. "
+                        f"Magnitude of combined query and context vectors cannot be zero. "
```
vicilliar: Good error message, we can do a few things to increase clarity here:
1. Output the query and context vectors back to the user
2. Add a link to docs (we have an existing pattern for this, using the `marqo_docs.py` module)
3. Let's add this zero magnitude limitation to the docs

tests/tensor_search/integ_tests/test_custom_vector_index_normalize_embeddings_true.py
```diff
@@ -0,0 +1,352 @@
...
+        super().setUp()
+        self.mappings = {
+            "my_custom_vector": {
+                "type": "custom_vector"
+            }
+        }
+
+        # Using arbitrary values so they're easy to eyeball
+        self.random_vector_1 = [1. for _ in range(512)]
+        self.normalized_random_vector_1_nd_array = np.array(self.random_vector_1)
```
vicilliar: This one isn't normalized yet. Maybe it should be `self.random_vector_1_nd_array`

tests/tensor_search/integ_tests/test_custom_vector_index_normalize_embeddings_true.py
```diff
@@ -0,0 +1,352 @@
...
+                        }],
+                        device="cpu",
+                        mappings=self.mappings if isinstance(index, UnstructuredMarqoIndex) else None,
+                        tensor_fields=["my_custom_vector"] if isinstance(index, UnstructuredMarqoIndex) else None
+                    )
+                )
+
+            if isinstance(index, UnstructuredMarqoIndex):
+                self.assertEquals(add_documents_response.errors, True)
+                self.assertEqual(add_documents_response.items[0].message, "Zero magnitude vector detected, cannot "
```
vicilliar: This error message needs to be more descriptive, specify that it's a `custom_vector` and possibly add a docs link as well.

tests/tensor_search/test_api.py
```diff
@@ -51,6 +51,60 @@ def test_add_or_replace_documents_tensor_fields(self):
             self.assertEqual(response.status_code, 200)
             mock_add_documents.assert_called_once()
 
+    def test_add_or_replace_documents_normalize_embeddings_true_zero_vector_passed(self):
```
vicilliar: These could be moved to the actual marqo-api-tests repo

src/marqo/tensor_search/tensor_search.py
```diff
@@ -297,15 +297,15 @@ def _add_documents_unstructured(config: Config, add_docs_params: AddDocsParams,
...
+                    if marqo_index.normalize_embeddings and marqo_index.parsed_marqo_version() >= MARQO_CUSTOM_VECTOR_NORMALIZATION_MINIMUM_VERSION:
                         try:
                             embeddings = normalize_vector(embeddings)
                         except core_exceptions.ZeroMagnitudeVectorError as e:
                             document_is_valid = False
                             unsuccessful_docs.append(
                                 (i, MarqoAddDocumentsItem(
                                     id=doc_id if doc_id is not None else '',
-                                    error=e.message,
+                                    error=e.__str__(),
```
papa99do: error should have the same content as message for compatibility reasons. (also, it's better to use str(e) instead of magic methods directly)

src/marqo/tensor_search/tensor_search.py
```diff
@@ -2692,7 +2690,7 @@ def delete_documents(config: Config, index_name: str, doc_ids: List[str]):
         )
     )
 
-def normalize_vector(embeddings: Union[List[List[float]], ndarray]) -> List[List[float]]:
+def normalize_vector(embeddings: Union[List[List[float]], ndarray, List[float]]) -> List[List[float]]:
```
papa99do, line 2693: What use case uses a List[float]?

adityabharadwaj198: None as of now, this change was based off of Joshua's comment here: https://github.com/marqo-ai/marqo/pull/970 that there might be a future need to accept List[float] type

tests/tensor_search/integ_tests/test_custom_vector_index_normalize_embeddings_true.py
adityabharadwaj198: [update: done] will be changed in the upcoming commit

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: [update: done] will change to error = e.message everywhere

src/marqo/core/search/hybrid_search.py
adityabharadwaj198: [update: done] will be changed in the upcoming commit

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1906,9 +1943,13 @@ def get_query_vectors_from_jobs(
...
 
             if q.index.normalize_embeddings:
-                norm = np.linalg.norm(merged_vector, axis=-1, keepdims=True)
-                if norm > 0:
-                    merged_vector /= np.linalg.norm(merged_vector, axis=-1, keepdims=True)
+                try:
+                    merged_vector = normalize_vector(merged_vector)
+                except core_exceptions.ZeroMagnitudeVectorError as e:
+                    raise core_exceptions.ZeroMagnitudeVectorError(
+                        f"Magnitude of combined query and context vectors cannot be zero. "
```
farshidz: No trailing space or period needed for exception message

tests/tensor_search/integ_tests/test_custom_vector_index_normalize_embeddings_true.py
```diff
@@ -0,0 +1,352 @@
+import os
```
farshidz: `test_custom_vector_field.py` seems like the right place for these tests

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: Agree with it being a convention, however I think trailing space & period make sense here since it's just the first half of the error message. [update: done] I will however remove the trailing space & period from the second half i.e `If you want to pass a zero vector, please set normalizeEmbeddings = False during index creation. `

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: 1. Discussed this with @farshidz and I agree that this may be too verbose of an error message. Also we're not sure how it would provide value to our customers
2. [update: done] Will be done in the next commit
3. Raised a PR for docs update: https://github.com/marqo-ai/marqodocs/pull/327

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: Jira: https://s2search.atlassian.net/browse/MOSD-138?atlOrigin=eyJpIjoiYzZkOTM0NzFhMTYwNDU3NDk5YjkwZDc1ZjYzYzhjNGMiLCJwIjoiaiJ9

tests/tensor_search/test_api.py
adityabharadwaj198: Raised a PR for changes in marqo-api-tests repo: https://github.com/marqo-ai/marqo-api-tests/pull/97

tests/tensor_search/integ_tests/test_custom_vector_index_normalize_embeddings_true.py
adityabharadwaj198: Moved to test_custom_vector_field.py

tests/tensor_search/integ_tests/test_custom_vector_index_normalize_embeddings_true.py
adityabharadwaj198: Changed the error message from where it is being thrown.

src/marqo/tensor_search/tensor_search.py
vicilliar: Thanks Aditya! For #1, that's fine then, as long as we have the `Magnitude of combined query and context vectors cannot be zero` message. Changes look good!

src/marqo/tensor_search/tensor_search.py
papa99do: @vicilliar , could you clarify why we need List[float] here?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1906,9 +1947,15 @@ def get_query_vectors_from_jobs(
             merged_vector = np.mean(weighted_vectors, axis=0)
 
             if q.index.normalize_embeddings:
-                norm = np.linalg.norm(merged_vector, axis=-1, keepdims=True)
-                if norm > 0:
-                    merged_vector /= np.linalg.norm(merged_vector, axis=-1, keepdims=True)
+                try:
+                    merged_vector = normalize_vector(merged_vector)
+                except core_exceptions.ZeroMagnitudeVectorError as e:
+                    if q.index.parsed_marqo_version() >= MARQO_CUSTOM_VECTOR_NORMALIZATION_MINIMUM_VERSION:
```
farshidz: Looks like this will error out for <2.13 anyway so we don't need this. Please verify that vespa errors out for 0 mag query and angular/prenormalized angular metric

farshidz (changes requested):

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: [Adding [slack conversation](https://marqo-ai.slack.com/archives/C05K6BFQUTF/p1727680962155369?thread_ts=1727339033.064519&cid=C05K6BFQUTF)] I'm not getting an error using either angular / prenormalized-angular as a distance metric here. 

This is the result with angular
`{'hits': [{'my_custom_vector': 'custom content is here!!', '_id': 'custom_vector_doc', '_highlights': [{'my_custom_vector': 'custom content is here!!'}], '_score': 0.38898452964834274}], 'query': None, 'limit': 10, 'offset': 0, 'processingTimeMs': 126}
`
the test case was:
```
settings = {
    "treat_urls_and_pointers_as_images": True,
    "model": "ViT-B/32",
    "ann_parameters": {
        "spaceType": "angular",
        "parameters": {"efConstruction": 512, "m": 16},
    },
    "normalize_embeddings": True,
}
mq.create_index("my-seventh-index", **settings)

result = mq.index('my-seventh-index').add_documents(
                documents=[{
                    "_id": "custom_vector_doc",
                                "my_custom_vector": {
                                    "content": "custom content is here!!",
                                    "vector": example_vector_1  # size is 512
                                }
                }],
                device="cpu", tensor_fields=["my_custom_vector", "text_field"],
                mappings={
            "my_custom_vector": {
                "type": "custom_vector"
            }
                })

search_result = mq.index('my-seventh-index').search(
    context={"tensor": [{"vector": zero_vector, "weight": 1} ]
             }
)

```

Result with prenormalised-angular: 

```
settings = {
    "treat_urls_and_pointers_as_images": True,
    "model": "ViT-B/32",
    "normalize_embeddings": True,
    "ann_parameters": {
    "spaceType": "prenormalized-angular",
    "parameters": {
      "efConstruction": 512,
      "m": 16
    }
  },
}

mq.create_index("my-eighth-index", **settings)
result = mq.index('my-eighth-index').add_documents(
    documents=[{
        "_id": "custom_vector_doc",
        "my_custom_vector": {
            "content": "custom content is here!!",
            "vector": example_vector_1  # size is 512
        }
    }],
    device="cpu", tensor_fields=["my_custom_vector", "text_field"],
    mappings={
        "my_custom_vector": {
            "type": "custom_vector"
        }
    })

search_result = mq.index('my-eighth-index').search(
    context={"tensor": [{"vector": zero_vector, "weight": 1} ]
             }
)
print(search_result)


```
`{'hits': [{'my_custom_vector': 'custom content is here!!', '_id': 'custom_vector_doc', '_highlights': [{'my_custom_vector': 'custom content is here!!'}], '_score': 0.5}], 'query': None, 'limit': 10, 'offset': 0, 'processingTimeMs': 5643}`

Since none of these threw an error, I believe we should have a version check, let me know if you don't agree - I can change it in that case.

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: This was closed during standup.

src/marqo/tensor_search/tensor_search.py
vicilliar: I'm alright with the version check if this is the case. Summarizing: if merged vector has magnitude zero: if marqo version is below the minimum, just ignore normalization, same behavior as before. If the version is equal to/greater: error with 400

farshidz (approved):

src/marqo/tensor_search/tensor_search.py
adityabharadwaj198: After having an offline discussion we have reverted logic to throw ZeroMagnitudeVectorError from search API. We will instead do this as part of a 3.x release. [jira](https://s2search.atlassian.net/jira/polaris/projects/MOSD/ideas/view/4863474?selectedIssue=MOSD-141&issueViewSection=comments&atlOrigin=eyJpIjoiOTU2YjIwMTI4NmM1NDE3MGJlNzJkMDhhMzk2OGMwMDgiLCJwIjoiaiJ9)

vicilliar (approved):

PR #1008, opened 2024-10-16
Fix `.`, `"` error in subfields in map score modifiers

RaynorChavez:
Bug fix

Scenario 1: 
- Say we have the document: `{"_id": "1_map", "text_field": "a photo of a cat", "map_score_mods_float": {"a.subsubfield": 0.5}},`
- And we attempt to access the score modifier in a query through: `"add_to_score": [{"field_name": "map_score_mods_float.a.subsubfield", "weight": 2}]`
- Then we will get an error:
```
status_code: 500
ValueError: too many values to unpack (expected 2)
```

Scenario 2:
- with the document `{"_id": "1_map", "text_field": "a photo of a cat", "map_score_mods_float": {"a"subsubfield": 0.5}},`
- we will get an error: ```Failed for special character '"': MarqoWebError: MarqoWebError Error message: {'message': "Marqo encountered an unexpected internal error.\nPlease create an issue on Marqo's GitHub repo (https://github.com/marqo-ai/marqo/issues) if this problem persists.", 'code': None, 'type': None, 'link': ''}
status_code: 500, type: None, code: None, link: ```
- In marqo, the error is: ```marqo.vespa.exceptions.VespaStatusError: 400: {"root":{"id":"toplevel","relevance":1.0,"fields":{"totalCount":0},"errors":[{"code":3,"summary":"Illegal query","message":"Could not set 'ranking.features.query(marqo__add_weights_tensor)' to '{\"map_score_mods_float.a\\\"subsubfield\": 2.0}': Could not parse '{\"map_score_mods_float.a\\\"subsubfield\": 2.0}' as a tensor of type tensor(p{}): At value position 26: Expected a ':' but got 's'"}]}}```

Scenario 3:
- for all subfields unicode characters and `\\` 
- users will be able to index, but will get a silent error when searching because of the discrepancy between the encoding format for these characters between add_docs and the query - thus, search will be possible, but the correct score_modifier will not be matched and applied

- For score modifiers with `.`, expected behavior is that we no longer get an internal error saying too many values to unpack and that the score is correctly applied.
- For score modifiers with `"`, expected behavior is that we no longer have the 500 internal error when searching and ~~that the score modifier is applied to the score~~ As discussed, we will let this silent error like other unicode characters until we hear back from the vespa team. Currently, we will encode " in unicode. I have confirmed that this will not match with documents explicitly containing the `\u0022` encoding.
- For all other unicode characters,  we will let these silent fail for now.

The main changes to enable this were:
- For `.` we simply split the query along the first `.` character, resulting in only 2 partitions, instead of in all `.`, which may result in multiple partitions if the query score modifier has more that one `.`
- For `"`, Vespa seems to have a bug wherein it attempts to escape `"` in the queries by transforming it to `\\\"`, leading to an invalid query. ~~To solve this without modifying vespa code, we switch to url encoding (where `"` is stored as `%22`) in both documents and queries. Note that this only applies to the subfields of score modifier fields in both unstructured and structured indexes~~ . Use `re.sub(r'(")', lambda m: r'\u0022', s)` .
- 
 - Need someone to check

* **Other information**:


src/marqo/core/utils/special_characters_encoding.py
```diff
@@ -0,0 +1,7 @@
+import re
+
+def custom_encode(s: str) -> str:
```
wanliAlex: Can you provide some doc strings for these functions?

tests/core/structured_vespa_index/test_structured_vespa_index.py
wanliAlex, line 1: More tests about `"` are also needed?

wanliAlex (changes requested):

src/marqo/core/utils/special_characters_encoding.py
RaynorChavez: Added!

tests/core/structured_vespa_index/test_structured_vespa_index.py
RaynorChavez: Yes, I had them before, but removed due to transient local integ test failing issue. I thought my tests were the problem.

I've re-added these now

wanliAlex (dismissed):
LGTM

tests/tensor_search/integ_tests/test_search_unstructured.py
```diff
@@ -1441,3 +1442,116 @@ def run():
                       vespa_query_kwargs["yql"])
         self.assertIn("embedding_query", vespa_query_kwargs["query_features"])
         self.assertNotIn("marqo__query_embedding", vespa_query_kwargs["query_features"])
+
+    def test_special_characters_in_map_score_modifiers(self):
+         special_characters = [
```
papa99do: nit. the special chars are duplicated in three tests, can we extract it?

tests/tensor_search/integ_tests/test_search_structured.py
```diff
@@ -1184,3 +1186,115 @@ def test_tensor_search_with_custom_vector_query(self):
...
+                     index_name=self.default_text_index,
+                     text="",
+                     score_modifiers=score_modifiers,
+                 )
+
+                 expected_score = 1.4165449318484857
+                 actual_score = res['hits'][0]['_score']
+
+                 if abs(actual_score - expected_score) > 0.01:
+                     #print(f"Soft fail for special character '{special_character}': Score mismatch")
```
papa99do: can we remove the commented out code from tests?

RaynorChavez: I wanted to leave them commented out for future debugging. What do you think?

tests/tensor_search/integ_tests/test_search_unstructured.py
RaynorChavez: I'll put it in a common file, thanks

src/marqo/core/utils/special_characters_encoding.py
```diff
@@ -0,0 +1,31 @@
...
```
farshidz: Why not have these in the base vespa index class?

src/marqo/core/vespa_index/vespa_index.py
```diff
@@ -129,20 +130,24 @@ def _get_score_modifiers(self, marqo_query: MarqoQuery) -> Optional[Dict[str, Di
         if marqo_query.score_modifiers:
             mult_tensor, add_tensor = self._convert_score_modifiers_to_tensors(marqo_query.score_modifiers)
```
farshidz: You can encode in this function to avoid an extra pass over the fields

tests/core/structured_vespa_index/test_structured_vespa_index.py
```diff
@@ -251,3 +252,39 @@ def test_to_vespa_document_fieldNotInIndex_fails(self):
...
+    def test_score_modifier_with_dot_in_subfield(self):
+        """
+        Test case where the score modifier field contains a period ('.') in the subfield name.
+        Ensures that only the first occurrence of '.' is split, and the rest is retained as part of the subfield.
+        """
+        marqo_doc = {
+            '_id': 'my_id',
+            'title': 'my title',
+            'price': 100.0,
+            "map_float_field": {"sub.field": 1.23},
```
farshidz, line 264: Test one dot as edge case (as you have) but also test more than one as the average case

tests/core/structured_vespa_index/test_structured_vespa_index.py
```diff
@@ -251,3 +252,39 @@ def test_to_vespa_document_fieldNotInIndex_fails(self):
...
+                'title': {
+                    constants.MARQO_DOC_CHUNKS: ['my', 'title'],
+                    constants.MARQO_DOC_EMBEDDINGS: [[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]]
+                }
+            }
+        }
+
+        actual_vespa_doc = self.vespa_index.to_vespa_document(marqo_doc)
+        expected_vespa_doc = {
+            'id': 'my_id',
```
farshidz: You could simplify your test here and just check the ID of the doc returned. Focused tests are more maintainable

src/marqo/core/utils/special_characters_encoding.py
RaynorChavez: Previously also used them to encode add docs to url encoding, so it made sense to separate. But now that we only use it for queries, you're right it's better to just put it in the base vespa index class.

Moved!

src/marqo/core/vespa_index/vespa_index.py
RaynorChavez: Thanks, done

tests/core/structured_vespa_index/test_structured_vespa_index.py
RaynorChavez: Added a new test

tests/core/structured_vespa_index/test_structured_vespa_index.py
RaynorChavez: Thanks, simplified

tests/tensor_search/integ_tests/test_search_unstructured.py
RaynorChavez: Done

tests/tensor_search/integ_tests/test_search_structured.py
RaynorChavez: Removed!

farshidz (approved):

PR #1012, opened 2024-10-17
Video/Audio File Size limits in search, embed, and add docs

RaynorChavez:
Feature

add docs has a hard 100 mb limit to video and audio files
search and embed have a 20s timeout for video and audio files

we introduce a new env var that allows the user to globally set their desired video and audio file size threshold (default 370 MB)

yes

* **Other information**:


src/marqo/s2_inference/clip_utils.py
```diff
@@ -182,15 +186,30 @@ def download_image_from_url(image_path: str, media_download_headers: dict, timeo
     headers.update(media_download_headers)
     c.setopt(pycurl.HTTPHEADER, [f"{k}: {v}" for k, v in headers.items()])
 
+    # callback to check file size for video and audio
```
wanliAlex, line 196: Can you check the `Content-Length` of the file before actually downloading the file?

tests/s2_inference/test_image_downloading.py
wanliAlex, line 1: Let's add an end-to-end test to better guarantee the behaviour for search and embed endpoints. Assert on the error message.

src/marqo/s2_inference/clip_utils.py
```diff
@@ -182,15 +186,30 @@ def download_image_from_url(image_path: str, media_download_headers: dict, timeo
...
+    if modality in ["video", "audio"]:
```
wanliAlex: Use Enum instead of string

src/marqo/s2_inference/clip_utils.py
```diff
@@ -182,15 +186,30 @@ def download_image_from_url(image_path: str, media_download_headers: dict, timeo
...
+        max_size = EnvVars.MARQO_MAX_VIDEO_AUDIO_SEARCH_FILE_SIZE
```
wanliAlex: Try to use the read env method in utils instead of calling this directly.

src/marqo/s2_inference/clip_utils.py
```diff
@@ -182,15 +186,30 @@ def download_image_from_url(image_path: str, media_download_headers: dict, timeo
...
+        def progress(download_total, downloaded, upload_total, uploaded):
+            if downloaded > max_size:
+                return 1  # Abort the download
```
wanliAlex: Can we raise ImageDownloadError directly here?

wanliAlex (changes requested):

src/marqo/s2_inference/clip_utils.py
RaynorChavez: We could, and this was an option, but this would add latency since we're adding another request for each query. I figure a majority of usecases won't exceed 370 mb and it might not be worth adding the extra latency for only a small number of eventualities where file size is > 370 MB.

src/marqo/s2_inference/clip_utils.py
RaynorChavez: done

src/marqo/s2_inference/clip_utils.py
RaynorChavez: done

src/marqo/s2_inference/clip_utils.py
RaynorChavez: done

src/marqo/s2_inference/clip_utils.py
RaynorChavez: done

src/marqo/s2_inference/clip_utils.py
RaynorChavez: currently not (that I know of), we will get a "Callback aborted" error

tests/s2_inference/test_image_downloading.py
RaynorChavez: done

src/marqo/s2_inference/clip_utils.py
wanliAlex: So even if we raise an error here, it will caught and convertd to a `Callback aborted` error?

RaynorChavez: yeah, I tried and it gave me a Callback aborted error when I raised here

src/marqo/tensor_search/add_docs.py
```diff
@@ -195,8 +196,8 @@ def download_and_chunk_media(url: str, device: str, modality: Modality,
     )
 
     if processor.total_size > MAX_FILE_SIZE:
-        raise ValueError(
-            f"File size ({processor.total_size / 1024 / 1024:.2f} MB) exceeds the maximum allowed size of 100 MB")
+        raise S2InferenceError(
```
wanliAlex: Good fix! The error will be caught and only return 400 for the specific document now, instead of failing the whole batch, right? 

Also, create something like the `MediaExceedsMaxSize` error and raise it here.

src/marqo/api/configs.py
```diff
@@ -31,6 +31,7 @@ def default_env_vars() -> dict:
         EnvVars.MARQO_LOG_LEVEL: "info",
         EnvVars.MARQO_MEDIA_DOWNLOAD_THREAD_COUNT_PER_REQUEST: 5,
         EnvVars.MARQO_IMAGE_DOWNLOAD_THREAD_COUNT_PER_REQUEST: 20,
+        EnvVars.MARQO_MAX_VIDEO_AUDIO_SEARCH_FILE_SIZE: 387973120,  # 370 megabytes in bytes
```
wanliAlex, line 34: I thought we use 100MB?

src/marqo/s2_inference/clip_utils.py
```diff
@@ -182,15 +193,31 @@ def download_image_from_url(image_path: str, media_download_headers: dict, timeo
     headers.update(media_download_headers)
     c.setopt(pycurl.HTTPHEADER, [f"{k}: {v}" for k, v in headers.items()])
 
+    # callback to check file size for video and audio
+    if modality in [Modality.VIDEO, Modality.AUDIO]:
+        max_size = read_env_vars_and_defaults_ints(EnvVars.MARQO_MAX_VIDEO_AUDIO_SEARCH_FILE_SIZE)
```
wanliAlex, line 198: @papa99do What do you think here? What is a good way to pass this size limit? This size limit is used in 2 places in our code (you can find the usage [below](https://github.com/marqo-ai/marqo/pull/1012/files#diff-b88aadc8cd6750daead4141b1038ddb2f9c91471f4e61495ca6f7fa4218927f7R190)).

wanliAlex (changes requested):

src/marqo/tensor_search/add_docs.py
RaynorChavez: Yep thats's right this will allow document only to fail not batch

src/marqo/api/configs.py
RaynorChavez: 100 was the old limit on add docs and 20s for search/embed.

370 is about how much we can download within 20s. This is to avoid regression in capability

src/marqo/s2_inference/clip_utils.py
```diff
@@ -182,15 +193,31 @@ def download_image_from_url(image_path: str, media_download_headers: dict, timeo
...
+        def progress(download_total, downloaded, upload_total, uploaded):
+            if downloaded > max_size:
+                return 1
```
papa99do, line 201: What error did we get when this happens? Is user going to get a user-friendly error message indicating the the media size exceeds the threshold?

src/marqo/s2_inference/clip_utils.py
papa99do: I think we can still use this approach temporarily for now, until we find a better way to do dep lookup (in functions) and dep injection (for classes).

src/marqo/s2_inference/clip_utils.py
RaynorChavez: Yes, the error message will be `error_message = f"Media file `{image_path}` exceeds the maximum allowed size for {modality}."`

src/marqo/tensor_search/add_docs.py
RaynorChavez: Addressed

wanliAlex (approved):
LGTM

PR #1018, opened 2024-10-21
Hiding the model internal data when retrieving index settings if the model is a Marqtune model.

keshav:
This PR introduces a feature to hide internal model data if a model is flagged as a Marqtune model in the/indexes/<index>/settings response. Additionally, it includes tests to verify that the model's internal data is appropriately blocked for Marqtune models.

Currently, the model properties in index settings include internal model data, such as the model's name, dimensions, and model location, even if the model is a Marqtune model. This could expose internal model details that should remain within Marqo ecosystem.

With this change, if a model is marked as a Marqtune model using the "isMarqtuneModel" property, the index settings API will block internal model data, returning only the "isMarqtuneModel": True property in the response. This aligns with internal implementation of Marqtune models.

No, this PR does not introduce a breaking change. Users can continue to access model properties as usual, with the exception that Marqtune model data is intentionally removed.

* **Other information**:
  * Added new test cases to verify that internal model properties are hidden for Marqtune models.
  * Verified that non-Marqtune models retain all their original properties in the index settings response.
  * This change does not affect existing index structures or other API functionality.

NA

There is no documentation on this feature yet. 


src/marqo/tensor_search/models/index_settings.py
```diff
@@ -250,6 +250,19 @@ def from_marqo_index(cls, marqo_index: core.MarqoIndex) -> "IndexSettings":
...
+    def get_model_properties(cls, marqo_index):
+        if marqo_index.model.properties is None:
+            return None
+
+        if marqo_index.model.properties.get('isMarqtuneModel', False):
+            # Hide all properties except for isMarqtuneModel
+            marqo_index.model.properties.pop('name', None)
+            marqo_index.model.properties.pop('dimensions')
+            marqo_index.model.properties.pop('model_location')
+            marqo_index.model.properties.pop('type')
```
papa99do, line 263: we only provide a default value for 'name', does that mean all other properties always has value? is it safer to provide default values for everyone of them to avoid KeyError?

src/marqo/tensor_search/models/index_settings.py
```diff
@@ -250,6 +250,19 @@ def from_marqo_index(cls, marqo_index: core.MarqoIndex) -> "IndexSettings":
         else:
             raise api_exceptions.InternalError(f"Unknown index type: {type(marqo_index)}")
 
+    @classmethod
+    def get_model_properties(cls, marqo_index):
```
papa99do, line 254: @keshav , a quick question. Do we also hide this in Cloud control plane?

tests/core/index_management/test_get_settings.py
```diff
@@ -208,4 +206,176 @@ def test_custom_settings(self):
...
+                     },
+                     "auth_required": False
+                 },
+                 "type": "open_clip",
+            },
+            normalizeEmbeddings=False,
+            textPreprocessing=TextPreProcessing(splitLength=3, splitMethod=TextSplitMethod.Word, splitOverlap=1)
+        ).to_marqo_index_request('a' + str(uuid.uuid4()).replace('-', ''))
+
+        indexes = self.create_indexes([
```
papa99do: Do we clean up the index after the test run? Consider adding them to the `setUpClass` method.

tests/core/index_management/test_get_settings.py
```diff
@@ -208,4 +206,176 @@ def test_custom_settings(self):
...
+                 "isMarqtuneModel": True,
+                 "dimensions": 384,
+                 "model_location": {
+                     "s3": {
+                         "Bucket": "marqtune-public-bucket",
+                         "Key": "marqo-test-hf-model/epoch_1.zip",
+                     },
+                     "auth_required": False
+                 },
+                 "type": "hf",
```
papa99do: Is this the only difference compared to the last test? Consider using subtests to reduce the duplicate test code. Also, consider add test cases for structured indexes (since there's change for structured index as well)

src/marqo/tensor_search/models/index_settings.py
keshav: Only name is optional as it wont exist for hf models, all others are mandatory.

src/marqo/tensor_search/models/index_settings.py
keshav: Yes we already do, only Marqo is currently exposing it and this pr should fix it

tests/core/index_management/test_get_settings.py
keshav: Moved the index creation to the setup method.

tests/core/index_management/test_get_settings.py
keshav: Created the subtests and included both structured and unstructured test cases 👍

farshidz (approved):

PR #1030, opened 2024-11-01
Add FFmpeg-cuda Support

wanliAlex:
feature

We only allow the CPU to do ffmpeg video decoding.

* **What is the new behaviour (if this is a feature change)?**
We add support for decoding video with GPU.

no

running

no

no

* **Other information**:
no

- [X] Docs have been added / updated (for bug fixes / features)


src/marqo/tensor_search/on_start_script.py
```diff
@@ -259,6 +265,59 @@ def run(self):
...
+            '-hwaccel', 'cuda',  # Use CUDA for hardware acceleration
+            '-f', 'lavfi',  # Input format is a lavfi (FFmpeg's built-in filter)
+            '-i', 'nullsrc=s=600x400',  # Generate a blank video source of 600x400 resolution
+            '-vframes', '1',  # Process only 1 frame
+            '-c:v', 'h264_nvenc',  # Use NVENC encoder
+            '-f', 'null',  # Output to null (discard the output)
+            '-'  # Output to stdout (discarded)
+        ]
+
+        result = subprocess.run(ffmpeg_command_gpu_check, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
papa99do: Consider handle timeout and other exceptions. Something like this:
```
try:
    result = subprocess.run(
        ffmpeg_command_gpu_check,
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE,
        check=True,          # Raises CalledProcessError for non-zero exit codes
        text=True,           # Decodes bytes to strings using default encoding
        timeout=30           # Optional: Set a timeout in seconds
    )
except subprocess.CalledProcessError as e:
    raise exceptions.StartupSanityCheckError(
                f"Failed to use GPU acceleration for video processing. "
                f"Ensure that your system has the required dependencies installed. "
                f"You can set 'MARQO_ENABLE_VIDEO_GPU_ACCELERATION=FALSE' to disable GPU acceleration. "
                f"Check {marqo_docs.configuring_marqo()} for more information. "
                f"Original error message: {result.stderr}"
            ) from e
except subprocess.TimeoutExpired as e:
     raise exceptions.StartupSanityCheckError(.....) from e
except Exception as e:
     raise exceptions.StartupSanityCheckError(.....) from e
```

src/marqo/tensor_search/streaming_media_processor.py
```diff
@@ -123,48 +125,140 @@ def process_media(self) -> List[Dict[str, torch.Tensor]]:
                     chunk_end = self.duration
                 else:
                     chunk_start = i * (chunk_duration - overlap_duration)
-                    chunk_end = min(chunk_start + chunk_duration, self.duration)
+                    chunk_end = min([chunk_start + chunk_duration, self.duration])
```
papa99do, line 124: Why do we put them in a list?

src/marqo/tensor_search/streaming_media_processor.py
```diff
@@ -123,48 +125,140 @@ def process_media(self) -> List[Dict[str, torch.Tensor]]:
...
+                '-ss', str(start_time),  # Start time
+                '-t', str(duration),  # Duration
+                '-i', self.url,  # Input file
+                '-vcodec', 'libx264',
+                '-acodec', 'aac',
+                '-f', 'mp4',
+                output_file
+            ])
+
+        result = subprocess.run(ffmpeg_command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
papa99do: When calling subprocess.run, we should carefully handle the possible exceptions to avoid failing the whole batch for a failure in one chunk or file. Please refer to https://docs.python.org/3/library/subprocess.html#exceptions for more details about what errors could be thrown. 
Also, subprocess.run is blocking, should we consider introduce timeout (and optional retry?) when downloading video/audio chunks? I'm not sure what could be a good default, but we should not block indefinitely.

src/marqo/tensor_search/streaming_media_processor.py
```diff
@@ -1,40 +1,47 @@
...
-        self.preprocessor = self.preprocessors[modality]
+        self.preprocessor = preprocessors.get_preprocessor(modality)
         self.media_download_headers = self._convert_headers_to_cli_format(media_download_headers)
-
         self.total_size, self.duration = self._fetch_file_metadata()
 
+        if enable_video_gpu_acceleration is None:
+            self.enable_video_gpu_acceleration = (
+                    utils.read_env_vars_and_defaults(EnvVars.MARQO_ENABLE_VIDEO_GPU_ACCELERATION) == 'TRUE'
+            )
```
papa99do: I think a better approach is always inject `enable_video_gpu_acceleration` value in the constructor. This will allow us to unit test this class with out mocking `read_env_vars_and_defaults`.  `download_and_chunk_media` in `add_docs.py` seems to be only function that constructs this class. It's a top-level function, we can call `utils.read_env_vars_and_defaults` there.

papa99do (changes requested):

src/marqo/tensor_search/streaming_media_processor.py
wanliAlex: The IDE raised a type warning before the change so I fixed it. The behaviour is exactly the same.

src/marqo/tensor_search/streaming_media_processor.py
wanliAlex: done. I move the method to add_docs.py

src/marqo/tensor_search/on_start_script.py
wanliAlex: Resolved. I try to avoid catching general exceptions here.

src/marqo/tensor_search/streaming_media_processor.py
wanliAlex: I added more error catches. I propose adding a retry in a different PR.

src/marqo/tensor_search/streaming_media_processor.py
```diff
@@ -14,17 +14,20 @@
...
-from marqo.tensor_search.enums import EnvVars
 from marqo.tensor_search.models.preprocessors_model import Preprocessors
 
 
 class StreamingMediaProcessor:
+
+    VIDEO_CPU_TIMOUT_OUT_MULTIPLIER = 10
+    AUDIO_CPU_TIMOUT_OUT_MULTIPLIER = 10
+    VIDEO_GPU_TIMOUT_OUT_MULTIPLIER = 10
+
```
papa99do, line 25: Why do we pick these values specifically to calculate the timeout?

papa99do (dismissed):
LGTM

papa99do (dismissed):

papa99do (approved):
LGTM

PR #1034, opened 2024-11-07
Fix the validation regressions for unstructured indexes

papa99do:
Bug fix

- List type of unstructured index is not validated correctly. numeric list is not captured in the validation and a generic error is raised when converting it to Vespa array type, causing the whole batch to fail
- Custom vector field not included in tensor fields will be silently ignored

- Numeric list type is captured earlier in the validation phase, which only cause the individual doc to fail, with a proper error message in the response
- Custom vector field not included in tensor fields will cause the whole batch to fail. 

No

Yes

No

No

* **Other information**:

### DIFF
--- a/src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
+++ b/src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
@@ -46,7 +46,7 @@ class SemiStructuredVespaDocument(MarqoBaseModel):
     @classmethod
     def from_vespa_document(cls, document: Dict, marqo_index: SemiStructuredMarqoIndex) -> "SemiStructuredVespaDocument":
         """
-        Instantiate an UnstructuredVespaDocument from a Vespa document.
+        Instantiate an SemiStructuredVespaDocument from a Vespa document.
         Used in get_document_by_id or get_documents_by_ids
         """
         fields = document.get(cls._VESPA_DOC_FIELDS, {})
@@ -76,10 +76,11 @@ class SemiStructuredVespaDocument(MarqoBaseModel):
 
     @classmethod
     def from_marqo_document(cls, document: Dict, marqo_index: SemiStructuredMarqoIndex) -> "SemiStructuredVespaDocument":
-        """Instantiate an UnstructuredVespaDocument from a valid Marqo document for feeding to Vespa"""
+        """Instantiate an SemiStructuredVespaDocument from a valid Marqo document for feeding to Vespa"""
 
         if index_constants.MARQO_DOC_ID not in document:
-            raise VespaDocumentParsingError(
+            # Please note we still use unstructured in the error message since it will be exposed to user
+            raise MarqoDocumentParsingError(
                 f"Unstructured Marqo document does not have a {index_constants.MARQO_DOC_ID} field. "
                 f"This should be assigned for a valid document")
 
@@ -116,7 +117,7 @@ class SemiStructuredVespaDocument(MarqoBaseModel):
                         instance.fixed_fields.float_fields[f"{field_name}.{k}"] = float(v)
                         instance.fixed_fields.score_modifiers_fields[f"{field_name}.{k}"] = v
             else:
-                raise VespaDocumentParsingError(
+                raise MarqoDocumentParsingError(
                     f"In document {doc_id}, field {field_name} has an "
                     f"unsupported type {type(field_content)} which has not been validated in advance.")
 
--- a/src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
+++ b/src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
@@ -7,29 +7,31 @@ from marqo import marqo_docs
 from marqo.api import exceptions as api_errors
 from marqo.core import constants
 from marqo.core.constants import MARQO_DOC_ID
-from marqo.core.vespa_index.add_documents_handler import AddDocumentsHandler, AddDocumentsError
-from marqo.core.models.add_docs_params import AddDocsParams
 from marqo.core.inference.tensor_fields_container import TensorFieldsContainer, MODALITY_FIELD_TYPE_MAP
 from marqo.core.models import UnstructuredMarqoIndex
+from marqo.core.models.add_docs_params import AddDocsParams
 from marqo.core.models.marqo_index import FieldType
 from marqo.core.unstructured_vespa_index.common import MARQO_DOC_MULTIMODAL_PARAMS
 from marqo.core.unstructured_vespa_index.unstructured_validation import validate_tensor_fields, validate_field_name, \
     validate_mappings_object_format, validate_coupling_of_mappings_and_doc
 from marqo.core.unstructured_vespa_index.unstructured_vespa_index import UnstructuredVespaIndex
+from marqo.core.vespa_index.add_documents_handler import AddDocumentsHandler, AddDocumentsError
 from marqo.s2_inference.errors import MediaDownloadError
 from marqo.s2_inference.multimodal_model_load import infer_modality, Modality
 
 from marqo.vespa.models import VespaDocument
 from marqo.vespa.models.get_document_response import Document
+from marqo.vespa.vespa_client import VespaClient
 
 # TODO deps to tensor_search needs to be removed
 from marqo.tensor_search.constants import ALLOWED_UNSTRUCTURED_FIELD_TYPES
-from marqo.tensor_search.validation import list_types_valid, validate_custom_vector, \
-    validate_multimodal_combination, validate_map_numeric_field
-from marqo.vespa.vespa_client import VespaClient
+from marqo.tensor_search.validation import validate_custom_vector, \
+    validate_map_numeric_field
 
 
 class UnstructuredAddDocumentsHandler(AddDocumentsHandler):
+    _MINIMUM_MARQO_VERSION_SUPPORTS_MAP_NUMERIC_FIELDS = semver.VersionInfo.parse("2.9.0")
+
     def __init__(self, marqo_index: UnstructuredMarqoIndex, add_docs_params: AddDocsParams, vespa_client: VespaClient):
         self._validate_add_docs_params(add_docs_params)
         super().__init__(marqo_index, add_docs_params, vespa_client)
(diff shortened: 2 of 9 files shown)


src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
```diff
@@ -100,27 +102,29 @@ def _validate_field(self, field_name: str, field_content: Any) -> None:
...
+                            f"which is not a string. Unstructured Marqo index only supports string lists."
+                        )
+
+            is_tensor_field = field_name in self.add_docs_params.tensor_fields
 
             if isinstance(field_content, dict):
                 if self.tensor_fields_container.is_custom_tensor_field(field_name):
-                    # TODO is_non_tensor_field check can be move out to AddDocsParams level
-                    validate_custom_vector(field_content, False, self.marqo_index.model.get_dimension())
-                elif self.tensor_fields_container.is_multimodal_field(field_name):
```
papa99do, line 114: Confirmed with Li that this logic is only required for Marqo 1.x. In Marqo 2.x, multimodal_combo fields are defined in the mappings instead of each individual doc. This was copied from the legacy validation code.

src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
```diff
@@ -100,27 +102,29 @@ def _validate_field(self, field_name: str, field_content: Any) -> None:
...
-                    # FIXME, multimodal field should not be present in the doc
-                    # TODO This validation should be done at AddDocsParams level
-                    validate_multimodal_combination(field_content, False,
-                                                    self.tensor_fields_container.get_multimodal_field_mapping(field_name))
-                elif self.marqo_index.parsed_marqo_version() < semver.VersionInfo.parse("2.9.0"):
-                    # TODO This check should not happen at root level
+                    # TODO should is_non_tensor_field check be moved out to AddDocsParams validation?
+                    # Please note that if one of the documents in the batch has a custom field which does not exist
+                    # in the tensor field, the whole batch will fail and user will get a 400 pydantic.ValidationError.
```
papa99do, line 120: The error message is not quite user-friendly. Should we capture it and convert it to a better message?
```
MarqoWebError: MarqoWebError: MarqoWebError Error message: {'message': '[{"loc": ["__root__"], "msg": "Cannot create custom_vector field (given field content: \'{\'content\': \'hello\', \'vector\': [1.0, 1.0.....]}\') as a non-tensor field. Add this field to \'tensor_fields\' to fix this problem.", "type": "value_error"}]', 'code': 'invalid_argument', 'type': 'invalid_request', 'link': ''} 
```

src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
```diff
@@ -76,10 +76,11 @@ def extract_field(cls, fields, name: str, default: Any):
 
     @classmethod
     def from_marqo_document(cls, document: Dict, marqo_index: SemiStructuredMarqoIndex) -> "SemiStructuredVespaDocument":
-        """Instantiate an UnstructuredVespaDocument from a valid Marqo document for feeding to Vespa"""
+        """Instantiate an SemiStructuredVespaDocument from a valid Marqo document for feeding to Vespa"""
 
         if index_constants.MARQO_DOC_ID not in document:
-            raise VespaDocumentParsingError(
+            # Please note we still use unstructured in the error message since it will be exposed to user
+            raise MarqoDocumentParsingError(
```
papa99do, line 83: Reason of the change: We are converting Marqo doc to Vespa doc, so we should raise MarqoDocumentParsingError instead of VespaDocumentParsingError

RaynorChavez (approved):

src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
wanliAlex: I think we can change the error message. It's not a breaking change

wanliAlex (approved):
LGTM

src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
```diff
@@ -7,29 +7,31 @@
...
 from marqo.tensor_search.constants import ALLOWED_UNSTRUCTURED_FIELD_TYPES
-from marqo.tensor_search.validation import list_types_valid, validate_custom_vector, \
-    validate_multimodal_combination, validate_map_numeric_field
-from marqo.vespa.vespa_client import VespaClient
+from marqo.tensor_search.validation import validate_custom_vector, \
+    validate_map_numeric_field
 
 
 class UnstructuredAddDocumentsHandler(AddDocumentsHandler):
+    _MINIMUM_MARQO_VERSION_SUPPORTS_MAP_NUMERIC_FIELDS = semver.VersionInfo.parse("2.9.0")
```
farshidz, line 33: I wonder if we should all of these in one file?

src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
```diff
@@ -100,27 +102,29 @@ def _validate_field(self, field_name: str, field_content: Any) -> None:
                     f"Allowed content types: {[ty.__name__ for ty in ALLOWED_UNSTRUCTURED_FIELD_TYPES]}"
                 )
 
-            if isinstance(field_content, list) and not list_types_valid(field_content):
```
farshidz, line 103: Perils of the utility style shared validation module

src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
```diff
@@ -100,27 +102,29 @@ def _validate_field(self, field_name: str, field_content: Any) -> None:
...
-                raise AddDocumentsError(
-                    f"Field content '{field_content}' "
-                    f"of type {type(field_content).__name__} is not of valid content type! "
-                    f"All list elements must be of the same type and that type must be int, float or string"
-                )
+            if isinstance(field_content, list):
+                for element in field_content:
+                    if not isinstance(element, str):
```
farshidz, line 107: You could use an `all()` statement but perhaps you're doing this for efficiency?

src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
```diff
@@ -100,27 +102,29 @@ def _validate_field(self, field_name: str, field_content: Any) -> None:
...
-                    # TODO This validation should be done at AddDocsParams level
-                    validate_multimodal_combination(field_content, False,
-                                                    self.tensor_fields_container.get_multimodal_field_mapping(field_name))
-                elif self.marqo_index.parsed_marqo_version() < semver.VersionInfo.parse("2.9.0"):
-                    # TODO This check should not happen at root level
+                    # TODO should is_non_tensor_field check be moved out to AddDocsParams validation?
+                    # Please note that if one of the documents in the batch has a custom field which does not exist
+                    # in the tensor field, the whole batch will fail and user will get a 400 pydantic.ValidationError.
+                    # We keep this behaviour unchanged to be compatible with the legacy unstructured index.
+                    validate_custom_vector(field_content, not is_tensor_field, self.marqo_index.model.get_dimension())
```
farshidz, line 122: nit should we start moving things out of validation to decommission the module? Seems like bad coupling of the two index types

src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
```diff
@@ -100,27 +102,29 @@ def _validate_field(self, field_name: str, field_content: Any) -> None:
...
+                elif self.marqo_index.parsed_marqo_version() < self._MINIMUM_MARQO_VERSION_SUPPORTS_MAP_NUMERIC_FIELDS:
+                    # We do not support map of numeric fields prior to 2.9.0
                     raise AddDocumentsError(
-                        f"The field {field_name} is a map field and only supported for indexes created with Marqo 2.9.0 "
-                        f"or later. See {marqo_docs.map_fields()} and {marqo_docs.mappings()}."
+                        f"The field {field_name} is a map field and only supported for indexes created with Marqo 2.9.0"
```
farshidz, line 126: Can move version to a central module and use the variable in this error message

tests/tensor_search/integ_tests/test_add_documents_unstructured.py
```diff
@@ -347,7 +350,9 @@ def test_add_documents_list_data_type_validation(self):
                     )
                 ).dict(exclude_none=True, by_alias=True)
                 assert add_res['errors'] is True
-                assert all(['error' in item for item in add_res['items'] if item['_id'].startswith('to_fail')])
+                assert all(['error' in item for item in add_res['items']])
+                assert all(['Unstructured Marqo index only supports string lists.' in item['message']
```
farshidz, line 354: Exception messages and log messages shouldn't end with a period. I think we're quite inconsistent with this pattern

src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
papa99do: Yes, Ideally. I'll create a ticket for this.

src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
papa99do: I've added a todo to do refactoring of  validation logic. Ideally we should do most of validations in pydantic models. I didn't have time to refactor this in my change.

src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
papa99do: no, just inlined the method it was calling. I thing the reason is we have the type of element exposed in the error message.

src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
papa99do: Yes, this will be done in a follow-up refactoring.

tests/tensor_search/integ_tests/test_add_documents_unstructured.py
papa99do: The full error message is not just one sentence though:
```
f"Field content {field_content} includes an element of type {type(element).__name__} "
 f"which is not a string. Unstructured Marqo index only supports string lists."
```

papa99do: Need to do more research. will be addressed in this ticket: https://s2search.atlassian.net/browse/MOSD-203

src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
papa99do: will be addressed in this ticket: https://s2search.atlassian.net/browse/MOSD-205

PR #1036, opened 2024-11-13
Skip convergence check when bootstrapping Marqo

papa99do:
Bug fix

Bootstrapping and rollback process will check the convergence of Vespa app first. If not converged, it will error out. 
See [this decision record](https://www.notion.so/Skip-convergence-check-when-bootstrapping-rolling-back-Marqo-0c65ef5ab70b44ee992665f0ab2cd482?pvs=4) for details

Bootstrapping and rollback process will skip the convergence check

No

Yes

No

No

* **Other information**:

### DIFF
--- a/src/marqo/core/index_management/index_management.py
+++ b/src/marqo/core/index_management/index_management.py
@@ -88,8 +88,7 @@ class IndexManagement:
         """
 
         # We skip the Vespa convergence check here so that Marqo instance can be bootstrapped even when Vespa is
-        # not converged. This will reduce the risk of having downtime when the Marqo instance crashes while waiting
-        # for Vespa app to converge.
+        # not converged.
         to_version = version.get_version()
         vespa_app_for_version_check = self._get_vespa_application(check_configured=False, need_binary_file_support=True,
                                                                   check_for_application_convergence=False)
@@ -101,9 +100,12 @@ class IndexManagement:
             return False
 
         with self._vespa_deployment_lock():
-            # Initialise another session based on the current active session
+            # Initialise another session based on the latest active Vespa session. The reason we do this again while
+            # holding the distributed lock is that the Vespa application might be changed by other operations when
+            # we wait for the lock. This time, we error out if the Vespa application is not converged, which reduces
+            # the chance of running into race conditions.
             vespa_app = self._get_vespa_application(check_configured=False, need_binary_file_support=True,
-                                                    check_for_application_convergence=False)
+                                                    check_for_application_convergence=True)
 
             # Only retrieving existing index when the vespa app is not configured and the index settings schema exists
             existing_indexes = self._get_existing_indexes() if not vespa_app.is_configured and \
@@ -118,10 +120,7 @@ class IndexManagement:
         Roll back Vespa application package to the previous version backed up in the current app package.
         """
         with self._vespa_deployment_lock():
-            # We skip the Vespa convergence check so that the rollback can be done even when Vespa is not converged.
-            # This allows us to rollback Vespa app when Vespa is in a weird state and fails to converge.
-            vespa_app = self._get_vespa_application(need_binary_file_support=True,
-                                                    check_for_application_convergence=False)
+            vespa_app = self._get_vespa_application(need_binary_file_support=True)
             vespa_app.rollback(version.get_version())
 
     def create_index(self, marqo_index_request: MarqoIndexRequest) -> MarqoIndex:
@@ -329,14 +328,11 @@ class IndexManagement:
 
         if need_binary_file_support and vespa_version < self._MINIMUM_VESPA_VERSION_TO_SUPPORT_UPLOAD_BINARY_FILES:
             # Binary files are only supported using VespaApplicationFileStore prior to Vespa version 8.382.22
-
-            # Please note that we don't pass in `check_for_application_convergence` parameter because we will always
-            # wait for convergence in this case. This will reduce the risk of overriding a newer version of the Vespa
-            # application package.
             application_package_store = VespaApplicationFileStore(
                 vespa_client=self.vespa_client,
                 deploy_timeout=self._deployment_timeout_seconds,
-                wait_for_convergence_timeout=self._convergence_timeout_seconds
+                wait_for_convergence_timeout=self._convergence_timeout_seconds,
+                check_for_application_convergence=check_for_application_convergence
             )
         else:
             application_package_store = ApplicationPackageDeploymentSessionStore(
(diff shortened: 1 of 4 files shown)


src/marqo/core/index_management/index_management.py
```diff
@@ -86,15 +86,24 @@ def bootstrap_vespa(self) -> bool:
...
+            # skip bootstrapping if already bootstrapped to this version or later
+            return False
 
-            if from_version and semver.VersionInfo.parse(from_version) >= semver.VersionInfo.parse(to_version):
-                # skip bootstrapping if already bootstrapped to this version or later
-                return False
+        with self._vespa_deployment_lock():
+            # Initialise another session based on the current active session
+            vespa_app = self._get_vespa_application(check_configured=False, need_binary_file_support=True,
+                                                    check_for_application_convergence=False)
```
adityabharadwaj198: I'm curious to know if we need to call _get_vespa_application() twice? Once here and previously on L#94

src/marqo/core/index_management/index_management.py
```diff
@@ -311,6 +329,10 @@ def _get_vespa_application(self, check_configured: bool = True, need_binary_file
 
         if need_binary_file_support and vespa_version < self._MINIMUM_VESPA_VERSION_TO_SUPPORT_UPLOAD_BINARY_FILES:
             # Binary files are only supported using VespaApplicationFileStore prior to Vespa version 8.382.22
+
+            # Please note that we don't pass in `check_for_application_convergence` parameter because we will always
+            # wait for convergence in this case. This will reduce the risk of overriding a newer version of the Vespa
+            # application package.
             application_package_store = VespaApplicationFileStore(
```
adityabharadwaj198, line 331: just curious, are we not checking convergence for this case due to our design decision? https://www.notion.so/Skip-convergence-check-when-bootstrapping-rolling-back-Marqo-0c65ef5ab70b44ee992665f0ab2cd482?pvs=4#13d75d43da4c80388506f913eeb4c272

src/marqo/core/index_management/index_management.py
papa99do: Hello @adityabharadwaj198 , good question. The reason we call it inside the distributed locking block is that the application package might be changed by other operations when we wait to acquire the lock. If we use the previous vespa session, it is more likely to fail the deployment due to race condition.

src/marqo/core/index_management/index_management.py
papa99do: Yes, correct. Old version of Vespa does not support optimistic locking mechanism for bootstrapping, which is more likely to cause the newer version of app be overridden if we do not wait for convergence.

src/marqo/core/index_management/index_management.py
```diff
@@ -86,15 +86,24 @@ def bootstrap_vespa(self) -> bool:
         Returns:
             True if Vespa was bootstrapped, False if it was already up-to-date
         """
-        with self._vespa_deployment_lock():
-            vespa_app = self._get_vespa_application(check_configured=False, need_binary_file_support=True)
 
-            to_version = version.get_version()
-            from_version = vespa_app.get_marqo_config().version if vespa_app.is_configured else None
+        # We skip the Vespa convergence check here so that Marqo instance can be bootstrapped even when Vespa is
+        # not converged. This will reduce the risk of having downtime when the Marqo instance crashes while waiting
```
farshidz: Description is confusing. Do you mean crashes when a deployment triggered by Marqo is converging? I think we're worried more about external deployments converging when Marqo has no changes to make and can start anyway

src/marqo/core/index_management/index_management.py
farshidz: Did we decide that the convergence check was superfluous anyway? Is there a scenario where we actually want this check? I would've thought it would reduce the chance of a concurrency failure if we did the check here

farshidz: @papa99do perhaps clarify this in code so others don't change it later

src/marqo/core/index_management/index_management.py
```diff
@@ -105,8 +114,15 @@ def bootstrap_vespa(self) -> bool:
...
     def rollback_vespa(self) -> None:
+        """
+        Roll back Vespa application package to the previous version backed up in the current app package.
+        """
         with self._vespa_deployment_lock():
-            self._get_vespa_application(need_binary_file_support=True).rollback(version.get_version())
+            # We skip the Vespa convergence check so that the rollback can be done even when Vespa is not converged.
+            # This allows us to rollback Vespa app when Vespa is in a weird state and fails to converge.
+            vespa_app = self._get_vespa_application(need_binary_file_support=True,
+                                                    check_for_application_convergence=False)
```
farshidz: Do we really need this change? What specific scenario are we addressing here?

src/marqo/core/index_management/index_management.py
farshidz: Also what happens if a convergence is in progress but we go and deploy anyway before it converges? Does the deployment go through or fail?

src/marqo/core/index_management/index_management.py
```diff
@@ -311,6 +329,10 @@ def _get_vespa_application(self, check_configured: bool = True, need_binary_file
 
         if need_binary_file_support and vespa_version < self._MINIMUM_VESPA_VERSION_TO_SUPPORT_UPLOAD_BINARY_FILES:
             # Binary files are only supported using VespaApplicationFileStore prior to Vespa version 8.382.22
+
+            # Please note that we don't pass in `check_for_application_convergence` parameter because we will always
+            # wait for convergence in this case. This will reduce the risk of overriding a newer version of the Vespa
+            # application package.
```
farshidz: Add a warning when we're ignoring the value here and waiting anyway

src/marqo/core/index_management/index_management.py
farshidz: @adityabharadwaj198 the opposite actually, in this case we're always checking convergence

src/marqo/vespa/vespa_client.py
```diff
@@ -193,6 +198,8 @@ def wait_for_application_convergence(self, timeout: int = 120) -> None:
             except (httpx.TimeoutException, httpcore.TimeoutException):
                 logger.error("Marqo timed out waiting for Vespa application to converge. Will retry.")
 
+        logger.debug(f"Vespa application did not converge within {timeout} seconds, the convergence status is {self._get_convergence_status()}")
```
farshidz: Why don't you add the status to the error message instead? This debug message seems redundant

src/marqo/core/index_management/index_management.py
papa99do: > when a deployment triggered by Marqo is converging? 
I does not matter who triggers the deployment, it can be Marqo or other processes.

src/marqo/vespa/vespa_client.py
papa99do: I've moved it to error message

farshidz (approved):

PR #1040, opened 2024-11-15
Move nltk download to start_on_script and remove the unsafe punkt package

wanliAlex:
bug fix

We download the nltk data when text preprocessing method is called. This could lead to a cold start and racing issue if this download is called by multiple threads.

We are also using the safe punkt package.

move the nltk download and check to start_on script

we don't download the unsafe punkt package anymore

no

yes

no

no

* **Other information**:
no


src/marqo/tensor_search/on_start_script.py
```diff
@@ -259,7 +258,34 @@ def run(self):
...
+
+
+class CheckNLTKTokenizers:
+    """Check if NLTK tokenizers are available, if not, download them.
+
+    NLTK tokenizers are included in the base-image, we do a sanity check to ensure they are available.
+    """
+    def run(self):
+        try:
+            nltk.data.find("tokenizers/punkt")
```
papa99do: this is for old version only, right? new version uses `tokenizers/punkt_tab`.
https://github.com/nltk/nltk/issues/3293#issuecomment-2284039835

src/marqo/tensor_search/on_start_script.py
```diff
@@ -259,7 +258,34 @@ def run(self):
...
+        except LookupError:
+            logger.info("NLTK punkt tokenizer not found. Downloading...")
+            nltk.download("punkt")
```
papa99do: In what case do we need to download it given we've already downloaded them in the base image? I'm referring to this: https://github.com/marqo-ai/marqo/pull/1040/files#r1847654510

src/marqo/tensor_search/on_start_script.py
wanliAlex: The newer version requires `tokenizers/punkt_tab`, which is added below.

src/marqo/tensor_search/on_start_script.py
wanliAlex: E.g., starting Marqo locally via `uvicorn`. Generally, we don't need to download it again since they are in the base-image.

src/marqo/tensor_search/on_start_script.py
papa99do: Sorry, I mean why do we still need to support the old version?

wanliAlex: We need both. Previously `punkt_tab` was part of `punkt` and then it was move out of `punkt` and we have to download both of them.

papa99do (approved):
LGTM

PR #1054, opened 2024-11-24
Rollback tests 

adityabharadwaj198:
* feature

* Currently we have no rollback tests  

* Functionality to do Rollback testing added 

No

Yes: https://github.com/marqo-ai/marqo/actions/runs/12047295783

NA

* NA

* **Other information**:


tests/backwards_compatibility_tests/compatibility_test_runner.py
```diff
@@ -625,10 +717,23 @@ def copy_state_from_container(
...
+    # Check for Major version compatibility
+    from_major_version = from_version.major
+    to_major_version = to_version.major
+
+    #If from major version & to major version aren't the same we cannot run backwards compatibility tests or rollback tests
+    if from_major_version != to_major_version:
+        logger.error(f"from Major version {from_major_version} & to Major version {to_major_version} cannot "
+                     f"be used for running backwards compatibility tests or rollback tests"
+                     f"since they are from different major versions")
+        sys.exit(0)
```
papa99do: when validation errors out, should we exit with an error code to indicate the test fails to run? Or maybe just throw an exception?

.github/workflows/backwards_compatibility_marqo_execution.yml
```diff
@@ -116,18 +116,34 @@ jobs:
...
-              --to_version_image_identifier "$TO_VERSION_IMAGE_IDENTIFIER" \
+            export PYTHONPATH=${{ github.workspace }}:$PYTHONPATH
+            python tests/backwards_compatibility_tests/compatibility_test_runner.py \
+            --mode=backwards_compatibility \
+            --from_version "$FROM_VERSION" \
+            --to_version "$TO_VERSION" \
+            --to_version_image_identifier "$TO_VERSION_IMAGE_IDENTIFIER" \
+
+      # Step to run rollback tests. Only runs if backwards compatibility tests have succeeded
+      - name: Run rollback tests
```
farshidz: Running these sequentially seems unnecessary and inefficient

.github/workflows/backwards_compatibility_marqo_execution.yml
```diff
@@ -116,18 +116,34 @@ jobs:
...
+        if: success()
+        env:
+          FROM_VERSION: ${{ inputs.from_version || github.event.inputs.from_version }}
+          TO_VERSION: ${{ inputs.to_version || github.event.inputs.to_version }}
+          TO_VERSION_IMAGE_IDENTIFIER: ${{ inputs.to_version_image_identifier || github.event.inputs.to_version_image_identifier }}
```
farshidz: I really think to_image will make more sense but perhaps seek a second opinion. Name looks unnecessarily long

tests/backwards_compatibility_tests/compatibility_test_runner.py
```diff
@@ -212,35 +207,46 @@ def start_marqo_from_version_container(from_version: str, from_version_volume, e
...
+
     logger.debug(
-        f"Starting Marqo container with to_version: {to_version}, "
-        f"from_version: {from_version} "
-        f"from_version_volume: {from_version_volume}, to_version_identifier: {to_version_identifier}, source: {source}")
-    container_name = f"marqo-{to_version}"
-    to_version = semver.VersionInfo.parse(to_version)
-    from_version = semver.VersionInfo.parse(from_version)
+        f"Starting Marqo container with target version: {target_version}, "
+        f"source version: {source_version} "
```
farshidz, line 238: Are we being consistent with the terminology? We shouldn't use both source version and from version especially in input args and outputs

tests/backwards_compatibility_tests/compatibility_test_runner.py
```diff
@@ -254,24 +260,22 @@ def start_marqo_to_version_container(to_version: str, from_version: str, from_ve
...
         "-e", "MARQO_MAX_CPU_MODEL_MEMORY=1.6"
     ]
-    # Append environment variables passed via the method
-    if env_vars:
-        for var in env_vars:
-            cmd.extend(["-e", var])
-
 
-    if from_version >= marqo_transfer_state_version and to_version >= marqo_transfer_state_version:
+    if source_version >= marqo_transfer_state_version and target_version >= marqo_transfer_state_version:
```
farshidz, line 269: Likewise with target version. Are we consistent with terminology?

tests/backwards_compatibility_tests/compatibility_test_runner.py
```diff
@@ -615,6 +525,188 @@ def copy_state_from_container(
...
+    try:
+        # Step 1: Start from_version container and run tests in prepare mode
+        logger.debug(f"Starting backwards compatibility tests with from_version: {from_version}, to_version: {to_version}, to_version_image_idenfitifer: {to_version_image_idenfitifer}")
+
+        # Generate a volume name to be used with the "from_version" Marqo container for state transfer.
+        from_version_volume = get_volume_name_from_marqo_version(from_version)
+
+        #Start from_version container
+        start_marqo_container(from_version, from_version_volume)
+        logger.debug(f"Started marqo container {from_version}")
```
farshidz: Marqo should be capitalized

tests/backwards_compatibility_tests/compatibility_test_runner.py
```diff
@@ -615,6 +525,188 @@ def copy_state_from_container(
...
+
+        try:
+            run_tests_across_versions("prepare", from_version)
+        except Exception as e:
+            logger.error(f"Error running tests across versions in 'prepare' mode: {e}")
```
farshidz: Logging and raising in an antipattern

tests/backwards_compatibility_tests/compatibility_test_runner.py
```diff
@@ -615,6 +525,188 @@ def copy_state_from_container(
...
+        stop_marqo_container(from_version)
+
+        # Step 3: Start to_version container by transferring state
+        logger.debug(f"Starting marqo to_version: {to_version} container by transferring state from version {from_version} to {to_version}")
+        start_marqo_container_by_transferring_state(to_version, from_version, from_version_volume,
+                                                    to_version_image_idenfitifer, "ECR")
+        logger.debug(f"Started marqo to_version: {to_version} container by transferring state")
+        # Step 4: Run tests
+        try:
+            run_tests_across_versions("test", from_version)
```
farshidz: Can we have an enum for mode? Passing strings around is error prone and confusing

tests/backwards_compatibility_tests/compatibility_test_runner.py
farshidz: Applies to all other cases in this PR

tests/backwards_compatibility_tests/compatibility_test_runner.py
```diff
@@ -366,130 +370,35 @@ def cleanup_volumes():
...
         run_test_mode(from_version)
 
-def full_test_run(to_version: str):
+def run_full_test_suite(from_version: str, to_version: str):
+    logger.debug(f"Running full test suite with from_version: {from_version}, to_version: {to_version}")
+    run_prepare_mode(from_version)
+    run_test_mode(from_version)
+    full_test_run(to_version)
+
+def full_test_run(marqo_version: str):
```
farshidz, line 394: When we run prepare again here, do we use new index names?

tests/backwards_compatibility_tests/compatibility_test_runner.py
```diff
@@ -366,130 +370,35 @@ def cleanup_volumes():
...
-        to_version (str): The target version of the Marqo container.
-        to_version_image_idenfitifer (str): The unique identifier for a to_version image. It can be either be the fully qualified image name with the tag
-                                (ex: 424082663841.dkr.ecr.us-east-1.amazonaws.com/marqo-compatibility-tests:abcdefgh1234)
-                                or the fully qualified image name with the digest (ex: 424082663841.dkr.ecr.us-east-1.amazonaws.com/marqo-compatibility-tests@sha256:1234567890abcdef).
-                                This is constructed in build_push_image.yml workflow and will be the qualified image name with digest for an automatically triggered workflow.
-
-    Raises:
-        ValueError: If the major versions of from_version and to_version are incompatible.
-        Exception: If there is an error during the test process.
+def run_tests_across_versions(mode: str, from_version: str):
```
farshidz: Both name and description appear misleading because this method doesn't actually perform state transfer and is operating on a single version of marqo only

tests/backwards_compatibility_tests/compatibility_test_runner.py
```diff
@@ -615,6 +525,188 @@ def copy_state_from_container(
...
+        to_version_image_idenfitifer (str): The unique identifier for a to_version image. It can be either be the fully qualified image name with the tag
+    """
+    logger.debug(f"Starting marqo rollback tests with from_version: {from_version}, to_version: {to_version}, to_version_image_idenfitifer: {to_version_image_idenfitifer}")
+    try:
+        # Step 0: Generate a volume name to be used with the "from_version" Marqo container for state transfer.
+        from_version_volume = get_volume_name_from_marqo_version(from_version)
+
+        # Step 1: Start a Marqo container using from_version and run tests in prepare mode.
+        start_marqo_container(from_version, from_version_volume)
+        logger.debug(f"Started marqo container {from_version}")
```
farshidz: nit I'm on the fence with these log levels. They feel like info messages. I also see you had to change your default log level to debug

tests/backwards_compatibility_tests/compatibility_test_runner.py
adityabharadwaj198: oops, meant to use a sys.exit(1) here - thanks for catching this. I'll change it to sys.exit(1). I also explored the possibility of throwing an exception here, it will work but it may produce noisy logs which may include python stack trace etc. So I've made changes such that this will throw an exception and the exception will be caught, logged and sys.exit(1) will be called

.github/workflows/backwards_compatibility_marqo_execution.yml
adityabharadwaj198: Rollback tests' initial tests are the same  as backwards compatibility tests (https://www.notion.so/Backwards-compatibility-and-rollback-testing-16d64207c72e486fa25001192f20817b?pvs=4#0acae9b3b5024b4b946e373a30f02c68). The idea here was that if anything is failing during the steps which are in both backwards_compatibility & rollback & they failing in backwards_compatibility why run rollback at all.

tests/backwards_compatibility_tests/compatibility_test_runner.py
adityabharadwaj198: Using target_version / source_version in this method makes more sense because this method is common to both backwards compatibility & rollback. In backwards compatibility it's required to upgrade marqo version (so target_version will be to_version and source_version will be from_version), and in rollback tests it's used to upgrade marqo version + downgrade marqo version (in which case target_version will be from_version and source_version will be to_version since you're downgrading). Here upgrading and downgrading basically mean running a newer marqo version & running an older marqo version resp.

tests/backwards_compatibility_tests/compatibility_test_runner.py
adityabharadwaj198: ^

tests/backwards_compatibility_tests/compatibility_test_runner.py
adityabharadwaj198: No we don't need to do that because after running prepare 1st time (and consequently having run test also), we will delete the indexes. We will then run prepare + test here again

tests/backwards_compatibility_tests/compatibility_test_runner.py
adityabharadwaj198: Changed name & description

tests/backwards_compatibility_tests/compatibility_test_runner.py
adityabharadwaj198: done

tests/backwards_compatibility_tests/compatibility_test_runner.py
adityabharadwaj198: fixed everywhere

tests/backwards_compatibility_tests/compatibility_test_runner.py
adityabharadwaj198: I've changed some logs to info, kept some logs as debug level (which we don't need to see in test runs) . And have changed the default log level to info. So anyone whose developing can change them to debug level and see all the logs (debug + info).

tests/backwards_compatibility_tests/compatibility_test_runner.py
adityabharadwaj198: Fixed everywhere

.github/workflows/backwards_compatibility_marqo_execution.yml
adityabharadwaj198: I have changed it to to_image wherever it made sense to do so, I've also shortened the variable name where-ever I couldn't change it to to_image, hope that helps

.github/workflows/backwards_compatibility_marqo_execution.yml
farshidz: But if they're going to pass, they'll pass faster. Or if they're going to fail for different reasons. Slow tests are a major issue for us currently

tests/backwards_compatibility_tests/compatibility_test_runner.py
```diff
@@ -611,25 +527,228 @@ def copy_state_from_container(
...
+
+        # Generate a volume name to be used with the "from_version" Marqo container for state transfer.
+        from_version_volume = get_volume_name_from_marqo_version(from_version)
+
+        #Start from_version container
+        start_marqo_container(from_version, from_version_volume)
+        logger.info(f"Started Marqo container {from_version}")
+
+        try:
+            run_tests_in_mode(Mode.PREPARE.value, from_version)
```
farshidz: Why do you need `.value`? Your func can just expect the enum

.github/workflows/backwards_compatibility_marqo_execution.yml
adityabharadwaj198: made them run in parallel

tests/backwards_compatibility_tests/compatibility_test_runner.py
adityabharadwaj198: changed it so it accepts enum

tests/backwards_compatibility_tests/compatibility_test_runner.py
```diff
@@ -221,36 +226,36 @@ def start_marqo_container_by_transferring_state(target_version: str, source_vers
...
     # Pull the image before starting the container
-    target_version_image_name = pull_marqo_image(image_identifier, source)
-    logger.debug(f" Printing image name {target_version_image_name}")
+    target_version_image_name = pull_marqo_image(image_name, source)
+    logger.info(f" Printing image name {target_version_image_name}")
     try:
         subprocess.run(["docker", "rm", "-f", container_name], check=True)
     except subprocess.CalledProcessError:
-        logger.debug(f"Container {container_name} not found, skipping removal.")
+        logger.warn(f"Container {container_name} not found, skipping removal.")
```
papa99do: nit. use warning instead, warn is deprecated.

tests/backwards_compatibility_tests/requirements.txt
```diff
@@ -1,6 +1,4 @@
 pytest==7.4.3
-pillow==9.3.0
-numpy==1.23.4
 marqo
```
papa99do, line 2: we do not specify marqo version here, which means we always use the latest one. Would this cause issue since py-marqo has requirements of minimum version of Marqo server. Can we use newer version of py-marqo to send request to older version of Marqo?

farshidz (dismissed):

papa99do (approved):

tests/backwards_compatibility_tests/requirements.txt
adityabharadwaj198: discussed offline that I have tested till version 2.9 with latest py-marqo

PR #1056, opened 2024-11-28
Expose cuda device health status in /healthz endpoint

papa99do:
Operation improvement

Cuda device can silently fail, which causes weird behaviour of Marqo

Expose a /healthz endpoint to be called to check cuda device status. This endpoint will return 500 errors when
- Cuda device becomes unavailable
- Cuda device is out of memory
This endpoint can be used by any scheduling framework as a liveness check for Marqo container.

No

Yes

No

N/A

* **Other information**:

### DIFF
--- a/src/marqo/config.py
+++ b/src/marqo/config.py
@@ -2,6 +2,7 @@ from typing import Optional, Union
 
 from kazoo.handlers.threading import KazooTimeoutError
 
+from marqo.core.inference.device_manager import DeviceManager
 from marqo.vespa.zookeeper_client import ZookeeperClient
 from marqo.core.document.document import Document
 from marqo.core.embed.embed import Embed
@@ -39,6 +40,7 @@ class Config:
 
         self.timeout = timeout
         self.backend = backend if backend is not None else enums.SearchDb.vespa
+        # TODO [Refactoring device logic] deprecate default_device since it's not used
         self.default_device = default_device if default_device is not None else (
             utils.read_env_vars_and_defaults(EnvVars.MARQO_BEST_AVAILABLE_DEVICE))
 
@@ -52,6 +54,7 @@ class Config:
         self.document = Document(vespa_client, self.index_management)
         self.recommender = Recommender(vespa_client, self.index_management)
         self.embed = Embed(vespa_client, self.index_management, self.default_device)
+        self.device_manager = DeviceManager()
 
     def set_is_remote(self, vespa_client: VespaClient):
         local_host_markers = ["localhost", "0.0.0.0", "127.0.0.1"]
--- a/src/marqo/core/exceptions.py
+++ b/src/marqo/core/exceptions.py
@@ -112,3 +112,15 @@ class DuplicateDocumentError(AddDocumentsError):
 
 class TooManyFieldsError(MarqoError):
     pass
+
+
+class DeviceError(MarqoError):
+    pass
+
+
+class CudaDeviceNotAvailableError(DeviceError):
+    pass
+
+
+class CudaOutOfMemoryError(DeviceError):
+    pass
--- a/src/marqo/core/inference/device_manager.py
+++ b/src/marqo/core/inference/device_manager.py
@@ -53,14 +53,27 @@ class DeviceManager:
         return [device for device in self.devices if device.type == DeviceType.cuda]
 
     def cuda_device_health_check(self) -> None:
+        """
+        Checks the status of the CUDA devices, and raises exceptions if it becomes
+        not available or out of memory.
+
+        raises
+          - CudaDeviceNotAvailableError if CUDA device is not available.
+          - CudaOutOfMemoryError if any CUDA device is out of memory.
+        """
         if not self._is_cuda_available_at_startup:
+            # If the instance is initialised without cuda devices, skip the check
             return
 
         if not torch.cuda.is_available():
+            # CUDA devices could become unavailable/unreachable if the docker container running Marqo loses access
+            # to the device symlinks. There is no way to recover from this, we will need to restart the container.
+            # See https://github.com/NVIDIA/nvidia-container-toolkit/issues/48 for more details.
             logger.error('Cuda device becomes unavailable.')
             raise CudaDeviceNotAvailableError('Cuda device becomes unavailable.')
 
         for device in self.cuda_devices:
+            # TODO confirm whether we should check all devices or just the default one
             cuda_device = torch.device(device.name)
             memory_stats = torch.cuda.memory_stats(cuda_device)
             logger.debug(f'Cuda device {device.name} with total memory {device.total_memory}. '
@@ -70,6 +83,8 @@ class DeviceManager:
                 torch.randn(3).to(cuda_device)
             except RuntimeError as e:
                 if 'out of memory' in str(e).lower():
+                    # If we encounter 'CUDA error: out of memory' error consistently, it means some threads are
+                    # holding the memory
                     logger.error(f'Cuda device {device.name} is out of memory. Total memory: {device.total_memory}. '
                                  f'Memory stats: {memory_stats}')
                     raise CudaOutOfMemoryError(f'Cuda device {device.name} is out of memory. '
--- a/src/marqo/core/inference/device_manager.py
+++ b/src/marqo/core/inference/device_manager.py
@@ -1,0 +1,76 @@
+from enum import Enum
+from typing import List
+
+import psutil
+import torch
+from pydantic import BaseModel
+
+from marqo.core.exceptions import CudaDeviceNotAvailableError, CudaOutOfMemoryError
+from marqo.logging import get_logger
+
+logger = get_logger('device_manager')
+
+
+class DeviceType(str, Enum):
+    cpu = 'cpu'
+    cuda = 'cuda'
+
+
+class Device(BaseModel):
+    id: int
+    name: str
+    type: DeviceType
+    total_memory: int
+
+    @classmethod
+    def cpu(cls) -> 'Device':
+        return Device(id=-1, name='cpu', type=DeviceType.cpu, total_memory=psutil.virtual_memory().total)
+
+    @classmethod
+    def cuda(cls, device_id, name, total_memory) -> 'Device':
+        return Device(id=device_id, name=name, type=DeviceType.cuda, total_memory=total_memory)
+
+
+class DeviceManager:
+    def __init__(self):
+        self._is_cuda_available_at_startup: bool = torch.cuda.is_available()
+        self.devices: List[Device] = [Device.cpu()]
+        self.best_available_device_type = DeviceType.cpu
+
+        if self._is_cuda_available_at_startup:
+            self.best_available_device_type = DeviceType.cuda
+            device_count = torch.cuda.device_count()
+            for device_id in range(device_count):
+                self.devices.append(Device.cuda(device_id,
+                                                torch.cuda.get_device_name(device_id),
+                                                torch.cuda.get_device_properties(device_id).total_memory))
+
+        logger.debug(f'Found devices {self.devices}. Best available device set to: '
+                     f'{self.best_available_device_type.value}.')
+
+    @property
+    def cuda_devices(self):
+        return [device for device in self.devices if device.type == DeviceType.cuda]
+
+    def cuda_device_health_check(self) -> None:
+        if not self._is_cuda_available_at_startup:
+            return
+
+        if not torch.cuda.is_available():
+            logger.error('Cuda device becomes unavailable.')
+            raise CudaDeviceNotAvailableError('Cuda device becomes unavailable.')
+
+        for device in self.cuda_devices:
+            cuda_device = torch.device(device.name)
+            memory_stats = torch.cuda.memory_stats(cuda_device)
+            logger.debug(f'Cuda device {device.name} with total memory {device.total_memory}. '
+                         f'Memory stats: {memory_stats}')
+            
+            try:
+                torch.randn(3).to(cuda_device)
+            except RuntimeError as e:
+                if 'out of memory' in str(e).lower():
+                    logger.error(f'Cuda device {device.name} is out of memory. Total memory: {device.total_memory}. '
+                                 f'Memory stats: {memory_stats}')
+                    raise CudaOutOfMemoryError(f'Cuda device {device.name} is out of memory. '
+                                               f'({memory_stats["allocated.all.current"]}/{device.total_memory})')
--- a/src/marqo/core/inference/device_manager.py
+++ b/src/marqo/core/inference/device_manager.py
@@ -19,7 +19,7 @@ class Device(BaseModel):
     id: int
     name: str
     type: DeviceType
-    total_memory: Optional[int] = 0
+    total_memory: Optional[int] = None
 
     @classmethod
     def cpu(cls) -> 'Device':
@@ -68,10 +68,10 @@ class DeviceManager:
             # CUDA devices could become unavailable/unreachable if the docker container running Marqo loses access
             # to the device symlinks. There is no way to recover from this, we will need to restart the container.
             # See https://github.com/NVIDIA/nvidia-container-toolkit/issues/48 for more details.
-            logger.error('Cuda device becomes unavailable')
-            raise CudaDeviceNotAvailableError('Cuda device becomes unavailable')
+            logger.error('CUDA device/s have become unavailable')
+            raise CudaDeviceNotAvailableError('CUDA device/s have become unavailable')
 
-        # TODO confirm whether we should check all devices or just the default one
+        oom_errors = []
         for device in self.cuda_devices:
             try:
                 cuda_device = torch.device(device.name)
@@ -82,13 +82,23 @@ class DeviceManager:
                 torch.randn(3, device=cuda_device)
             except RuntimeError as e:
                 if 'out of memory' in str(e).lower():
-                    # If we encounter 'CUDA error: out of memory' error consistently, it means some threads are
-                    # holding the memory
+                    # `~torch.cuda.empty_cache` doesn't increase the amount of GPU memory available for PyTorch.
+                    # However, it may help reduce fragmentation of GPU memory in certain cases.
+                    torch.cuda.empty_cache()
+
                     logger.error(f'Cuda device {device.name} is out of memory. Total memory: {device.total_memory}. '
                                  f'Memory stats: {str(memory_stats)}')
                     allocated_mem = memory_stats.get("allocated.all.current", None) if memory_stats else None
-                    raise CudaOutOfMemoryError(f'Cuda device {device.name} is out of memory: '
-                                               f'({allocated_mem}/{device.total_memory})')
+                    oom_errors.append(f'Cuda device {device.name} is out of memory:'
+                                      f' ({allocated_mem}/{device.total_memory})')
+                else:
+                    # Log out a warning message when encounter other transient errors.
+                    logger.warning(f'Encountered issue inspecting Cuda device {device.name}: {str(e)}')
             except Exception as e:
                 # Log out a warning message when encounter other transient errors.
                 logger.warning(f'Encountered issue inspecting Cuda device {device.name}: {str(e)}')
+
+        if oom_errors:
+            # We error out if any cuda device is out of memory. If this happens consistently, the memory might be held
+            # by a long-running thread, and Marqo will need to be restarted to get to a healthy status
+            raise CudaOutOfMemoryError(';'.join(oom_errors))
--- a/src/marqo/core/inference/device_manager.py
+++ b/src/marqo/core/inference/device_manager.py
@@ -1,9 +1,10 @@
 from enum import Enum
+from functools import cached_property
 from typing import List, Optional
 
 import torch
-from pydantic import BaseModel
 
+from marqo.base_model import ImmutableBaseModel
 from marqo.core.exceptions import CudaDeviceNotAvailableError, CudaOutOfMemoryError
 from marqo.logging import get_logger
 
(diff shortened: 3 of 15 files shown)


src/marqo/config.py
```diff
@@ -39,6 +40,7 @@ def __init__(
 
         self.timeout = timeout
         self.backend = backend if backend is not None else enums.SearchDb.vespa
+        # TODO [Refactoring device logic] deprecate default_device since it's not used
```
papa99do, line 43: To control the scope of this ticket, and reduce the risk of doing refactoring without enough test coverage, I added TODOs for all the possible improvement to consolidate device check logic.

src/marqo/core/inference/device_manager.py
```diff
@@ -0,0 +1,94 @@
...
+import torch
+from pydantic import BaseModel
+
+from marqo.core.exceptions import CudaDeviceNotAvailableError, CudaOutOfMemoryError
+from marqo.logging import get_logger
+
+logger = get_logger('device_manager')
+
+
+class DeviceType(str, Enum):
```
vicilliar, line 14: We have an existing enum for this in `tensor_search`: https://github.com/marqo-ai/marqo/blob/a262ede62d0d960365cf243ba34f07702117ad8e/src/marqo/tensor_search/enums.py#L30
although this might actually be a better place to put it

src/marqo/tensor_search/on_start_script.py
```diff
@@ -109,6 +110,7 @@ def id_to_device(id):
 
 
 class SetBestAvailableDevice:
+    # TODO [Refactoring device logic] move this logic to device manager, get rid of MARQO_BEST_AVAILABLE_DEVICE envvar
```
vicilliar, line 113: good idea. We could remove the `torch.cuda.is_available()` call here and replace it with `DeviceManager` `_is_cuda_available_at_startup()` right?

src/marqo/core/inference/device_manager.py
```diff
@@ -0,0 +1,94 @@
...
+
+class Device(BaseModel):
+    id: int
+    name: str
+    type: DeviceType
+    total_memory: Optional[int] = 0
+
+    @classmethod
+    def cpu(cls) -> 'Device':
+        return Device(id=-1, name='cpu', type=DeviceType.cpu)
```
vicilliar: Do we have no `total_memory` for `cpu`?

src/marqo/core/inference/device_manager.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        if not self._is_cuda_available_at_startup:
+            # If the instance is initialised without cuda devices, skip the check
+            return
+
+        if not torch.cuda.is_available():
+            # CUDA devices could become unavailable/unreachable if the docker container running Marqo loses access
+            # to the device symlinks. There is no way to recover from this, we will need to restart the container.
+            # See https://github.com/NVIDIA/nvidia-container-toolkit/issues/48 for more details.
+            logger.error('Cuda device becomes unavailable')
```
vicilliar: We could fix the error message: `CUDA device/s have become unavailable`

src/marqo/core/exceptions.py
```diff
@@ -112,3 +112,15 @@ class DuplicateDocumentError(AddDocumentsError):
 
 class TooManyFieldsError(MarqoError):
     pass
+
+
+class DeviceError(MarqoError):
```
vicilliar, line 117: How is this handled by the API layer?

tests/tensor_search/test_api.py
```diff
@@ -528,6 +529,23 @@ def test_invalid_structured_index_field_features(self):
...
+        for cuda_exception in [
+            CudaDeviceNotAvailableError('Cuda device becomes unavailable'),
+            CudaOutOfMemoryError('Cuda device cuda:0 is out of memory')
+        ]:
+            with self.subTest(cuda_exception):
+                with patch("marqo.core.inference.device_manager.DeviceManager.cuda_device_health_check",
+                           side_effect=cuda_exception):
+                    response = self.client.get("/healthz")
+                    self.assertEqual(response.status_code, 500)
+                    self.assertIn(cuda_exception.message, response.json()['message'])
```
vicilliar, line 547: We should have an integration test for this as well, on a marqo container

src/marqo/tensor_search/api.py
```diff
@@ -579,16 +580,22 @@ def schema_validation(index_name: str, settings_object: dict):
...
     health_status = marqo_config.monitoring.get_health()
     return HealthResponse.from_marqo_health_status(health_status)
 
 
-@app.get('/memory', include_in_schema=False)
-@utils.enable_debug_apis()
-def memory():
-    return memory_profiler.get_memory_profile()
+@app.get("/healthz", include_in_schema=False)
+def check_health(marqo_config: config.Config = Depends(get_config)):
```
vicilliar: Is there an issue with having the same function name `check_health` as the `health` endpoint? We don't want this to conflict

src/marqo/core/inference/device_manager.py
```diff
@@ -0,0 +1,94 @@
...
+        # TODO confirm whether we should check all devices or just the default one
+        for device in self.cuda_devices:
+            try:
+                cuda_device = torch.device(device.name)
+                memory_stats = torch.cuda.memory_stats(cuda_device)
+                logger.debug(f'Cuda device {device.name} with total memory {device.total_memory}. '
+                             f'Memory stats: {str(memory_stats)}')
+
+                torch.randn(3, device=cuda_device)
+            except RuntimeError as e:
```
vicilliar, line 92: Is my understanding correct that the very first error encountered will stop this loop? What if multiple CUDA devices are out of memory? Should we distinguish if some of the devices are available and some or not? That could be useful. Maybe we could report on the status of each CUDA device.

src/marqo/core/inference/device_manager.py
```diff
@@ -0,0 +1,94 @@
...
+                if 'out of memory' in str(e).lower():
+                    # If we encounter 'CUDA error: out of memory' error consistently, it means some threads are
+                    # holding the memory
+                    logger.error(f'Cuda device {device.name} is out of memory. Total memory: {device.total_memory}. '
+                                 f'Memory stats: {str(memory_stats)}')
+                    allocated_mem = memory_stats.get("allocated.all.current", None) if memory_stats else None
+                    raise CudaOutOfMemoryError(f'Cuda device {device.name} is out of memory: '
+                                               f'({allocated_mem}/{device.total_memory})')
+            except Exception as e:
```
vicilliar, line 102: This exception catch seems too broad. This may mask CUDA issues in the future

src/marqo/core/monitoring/monitoring.py
```diff
@@ -154,6 +154,7 @@ def _get_vespa_health(self, hostname_filter: Optional[str]) -> VespaHealthStatus
         )
 
     def get_cuda_info(self) -> MarqoCudaInfoResponse:
+        # TODO [Refactoring device logic] move this logic to device manager
```
vicilliar, line 157: It seems like this method overlaps a lot of functionality of `cuda_device_health_check`. But yes, refactoring can be another ticket

tests/core/inference/test_device_manager.py
```diff
@@ -0,0 +1,90 @@
...
+        self.assertTrue(device_manager._is_cuda_available_at_startup)
+
+    def test_cuda_health_check_should_skip_without_cuda_devices(self):
+        device_manager = self._device_manager_without_cuda()
+
+        with mock.patch("marqo.core.inference.device_manager.torch") as mock_cuda:
+            device_manager.cuda_device_health_check()
+        self.assertEqual(0, len(mock_cuda.mock_calls))
+
+    def test_cuda_health_check_should_pass_when_cuda_device_is_healthy(self):
```
vicilliar, line 55: Let's add a check where there are multiple cuda devices. 1 or more is healthy, while some are unhealthy

vicilliar (changes requested):

src/marqo/core/inference/device_manager.py
papa99do: Yes, my plan is to use this one in the future. The one in enums.py will be removed.

src/marqo/core/inference/device_manager.py
papa99do: I used to populate it from psutil. But we never use it in this health check, so I removed it. We can add it easily if we see a use case.

tests/tensor_search/test_api.py
papa99do: Good point. You mean an API test, right? probably not using pymarqo since this is an internal endpont. Will add one once this is merged

src/marqo/tensor_search/api.py
papa99do: good point. I've ignored this, will change the function name.

src/marqo/core/inference/device_manager.py
papa99do: The purpose is to catch all CUDA errors or non-CUDA related errors, so that we don't crash the container on unknown errors.

src/marqo/core/inference/device_manager.py
papa99do: Yes, correct. we can check all cuda devices and error out if any one is out of memory.

src/marqo/core/exceptions.py
papa99do: See https://github.com/marqo-ai/marqo/pull/1056/files#diff-ba0027ca86b6804c6a459ce5153a53376e7c99eae82e93bcde231c854b0d402aR115.

src/marqo/core/inference/device_manager.py
```diff
@@ -0,0 +1,104 @@
...
+
+logger = get_logger('device_manager')
+
+
+class DeviceType(str, Enum):
+    cpu = 'cpu'
+    cuda = 'cuda'
+
+
+class Device(BaseModel):
```
farshidz: We have custom base pydantic classes that may be more suitable for(e.g., immutable base class)

src/marqo/core/inference/device_manager.py
```diff
@@ -0,0 +1,104 @@
...
+    @classmethod
+    def cpu(cls) -> 'Device':
+        return Device(id=-1, name='cpu', type=DeviceType.cpu)
+
+    @classmethod
+    def cuda(cls, device_id, name, total_memory) -> 'Device':
+        return Device(id=device_id, name=name, type=DeviceType.cuda, total_memory=total_memory)
+
+
+class DeviceManager:
```
farshidz, line 38: docstring would be nice so we know what this is expected to do

src/marqo/core/inference/device_manager.py
```diff
@@ -0,0 +1,104 @@
...
+        self.devices: List[Device] = [Device.cpu()]
+        self.best_available_device_type = DeviceType.cpu
+
+        if self._is_cuda_available_at_startup:
+            self.best_available_device_type = DeviceType.cuda
+            device_count = torch.cuda.device_count()
+            for device_id in range(device_count):
+                self.devices.append(Device.cuda(device_id,
+                                                torch.cuda.get_device_name(device_id),
+                                                torch.cuda.get_device_properties(device_id).total_memory))
```
farshidz, line 55: Do CUDA devices have separate memory? I believe it's shared

src/marqo/core/inference/device_manager.py
```diff
@@ -0,0 +1,104 @@
...
+
+        logger.debug(f'Found devices {self.devices}. Best available device set to: '
+                     f'{self.best_available_device_type.value}.')
+
+    @property
+    def cuda_devices(self):
+        return [device for device in self.devices if device.type == DeviceType.cuda]
```
farshidz: Is this called a lot? If yes you could keep a map to avoid the linear filtering

src/marqo/core/inference/device_manager.py
```diff
@@ -0,0 +1,104 @@
...
+        """
+        if not self._is_cuda_available_at_startup:
+            # If the instance is initialised without cuda devices, skip the check
+            return
+
+        if not torch.cuda.is_available():
+            # CUDA devices could become unavailable/unreachable if the docker container running Marqo loses access
+            # to the device symlinks. There is no way to recover from this, we will need to restart the container.
+            # See https://github.com/NVIDIA/nvidia-container-toolkit/issues/48 for more details.
+            logger.error('CUDA device/s have become unavailable')
```
farshidz: Logging and raising is bad practice. Raising suffices

src/marqo/core/inference/device_manager.py
```diff
@@ -0,0 +1,104 @@
...
+            raise CudaDeviceNotAvailableError('CUDA device/s have become unavailable')
```
farshidz: device(s)?

src/marqo/core/inference/device_manager.py
```diff
@@ -0,0 +1,104 @@
...
+                memory_stats = torch.cuda.memory_stats(cuda_device)
+                logger.debug(f'Cuda device {device.name} with total memory {device.total_memory}. '
+                             f'Memory stats: {str(memory_stats)}')
+
+                torch.randn(3, device=cuda_device)
+            except RuntimeError as e:
+                if 'out of memory' in str(e).lower():
+                    # `~torch.cuda.empty_cache` doesn't increase the amount of GPU memory available for PyTorch.
+                    # However, it may help reduce fragmentation of GPU memory in certain cases.
+                    torch.cuda.empty_cache()
```
farshidz: Why are we calling this here?

src/marqo/core/inference/device_manager.py
```diff
@@ -0,0 +1,104 @@
...
+
+                    logger.error(f'Cuda device {device.name} is out of memory. Total memory: {device.total_memory}. '
```
farshidz: I think CUDA should be all caps everywhere

src/marqo/core/inference/device_manager.py
farshidz: I agree catching Exception is dubious. If there's a bug in the code you'll just mask it and log a warning. An error here won't crash the container, it'll cause a 500 API response which is fine for an unhandled situation

src/marqo/core/monitoring/monitoring.py
farshidz, line 157: I wonder if DeviceManager should be a dependency of this class, so API calls this class. That seems for in line with our class structure

src/marqo/tensor_search/api.py
```diff
@@ -579,16 +580,31 @@ def schema_validation(index_name: str, settings_object: dict):
...
     health_status = marqo_config.monitoring.get_health()
     return HealthResponse.from_marqo_health_status(health_status)
 
 
-@app.get('/memory', include_in_schema=False)
-@utils.enable_debug_apis()
-def memory():
-    return memory_profiler.get_memory_profile()
+@app.get("/healthz", include_in_schema=False)
+def liveness_check(marqo_config: config.Config = Depends(get_config)) -> JSONResponse:
```
farshidz, line 596: Is this an internal endpoint (undocumented)?

src/marqo/tensor_search/api.py
```diff
@@ -579,16 +580,31 @@ def schema_validation(index_name: str, settings_object: dict):
...
+    """
+    This liveness check endpoint does a quick status check, and error out if any component encounters unrecoverable
+    issues. This only does a check on the cuda devices right now.
+    Docker schedulers could leverage this endpoint to decide whether to restart the Marqo container.
+
+    Returns:
+        200 - if all checks pass
+        500 - if any check fails
+    """
+    marqo_config.device_manager.cuda_device_health_check()
```
farshidz, line 606: I have two issues with this
1. Our convention is 500/InternalError means unhandled/bug. We're violating that convention
2. 503 seems like the right code

src/marqo/core/inference/device_manager.py
papa99do: This is from AWS Q: 

> G4 instances with multiple GPUs do not share VRAM (video RAM) between the GPUs. Each GPU in a G4 instance has its own dedicated VRAM. This architecture allows for better performance and isolation between workloads running on different GPUs within the same instance.

src/marqo/core/inference/device_manager.py
papa99do: depends on how often the liveness check request is made. In Marqo Cloud, I think it's one per 10s. Since devices should never change, I'll change to use @cached_property

src/marqo/core/inference/device_manager.py
papa99do: Will remove logging. I think raising will also log the error, right?

src/marqo/core/inference/device_manager.py
papa99do: will update

src/marqo/core/monitoring/monitoring.py
papa99do: yes, the plan is move all device related logic to DeviceManager. I think API can directly call Device manager.

src/marqo/tensor_search/api.py
papa99do: That is correct, `include_in_schema=False` makes sure it's not included in the openAPI specs either.

src/marqo/tensor_search/api.py
papa99do: 503 makes sense. It indicates some service upstream (CUDA in this case.) is not working, so the service is down. I'll change it. Liveness check treats all non 2xx code the same way.

src/marqo/core/inference/device_manager.py
papa99do: Thanks for the reminder, will use the immutable base class

src/marqo/core/inference/device_manager.py
papa99do: will add

src/marqo/core/inference/device_manager.py
papa99do: I just added it thinking it could probably defragment the GPU memory and may help with the OOM issue. But given it another thought, health check endpoints should not have side effects. I'll remove it.

src/marqo/core/inference/device_manager.py
```diff
@@ -0,0 +1,104 @@
...
+                    oom_errors.append(f'Cuda device {device.name} is out of memory:'
+                                      f' ({allocated_mem}/{device.total_memory})')
+                else:
+                    # Log out a warning message when encounter other transient errors.
+                    logger.warning(f'Encountered issue inspecting Cuda device {device.name}: {str(e)}')
+            except Exception as e:
+                # Log out a warning message when encounter other transient errors.
+                logger.warning(f'Encountered issue inspecting Cuda device {device.name}: {str(e)}')
+
+        if oom_errors:
```
vicilliar, line 106: This only works if all the errors are oom errors. What about the case where we have multiple CUDA devices:
1. gpu 0 is OOM
2. gpu 1 is unavailable/unreachable for some reason (`CudaDeviceNotAvailableError`)
3. gpu 2 has some unknown exception when being inspected (this will go to the `Exception` catch)
4. gpu 3 is fine

Some questions:
- Should we always return a 500 as long as even 1 device has an issue? If this is the case, the error should have the status of each gpu with a problem and clearly state the reason. eg: "device 0 is out of memory (14.9g/15g); device 1 is unavailable; device 2 encountered an error"

- Should we return a 200 as long as there is even 1 healthy device instead?
cc @farshidz

tests/tensor_search/test_api.py
vicilliar: yep sorry I meant an api test.

vicilliar (changes requested):

src/marqo/tensor_search/api.py
vicilliar: I think there may be upside to returning more info about the devices in this endpoint. Currently we're just using it as a healthy/unhealthy check, but we could also return info for the healthy devices. Like the current memory being used.

src/marqo/core/inference/device_manager.py
papa99do: > Should we always return a 500 as long as even 1 device has an issue?
Yes, untill we improved Marqo to quarantine the problematic devices.

> Should we return a 200 as long as there is even 1 healthy device instead?
No this will cause Marqo to behaviour weirdly. Some requests will fail but some will succeed. We would rather restart Marqo and make sure all devices are healthy.

src/marqo/core/inference/device_manager.py
vicilliar: I agree we should still raise a 500 if other errors occur aside from oom

src/marqo/core/inference/device_manager.py
farshidz: Depends on the caller of course. But the idea is either someone will handle it (and not raise, possibly log) or it'll be raised, in which case it'll be logged, yes

src/marqo/core/inference/device_manager.py
vicilliar: Makes sense. If this is the case, other errors should be attended to the "error list", not just OOM errors. So when the 500 is returned, it has all issues related to all cuda devices

src/marqo/core/inference/device_manager.py
farshidz: Log and raise can lead to logging the same thing multiple times if we raise it later, or higher layers catch and log too

farshidz (approved):

src/marqo/core/inference/device_manager.py
papa99do: I only error out for OOM issues but not other 'unknown issues' because we don't know if other issues are transient or unrecoverable. I just log them out so we can triage later.

src/marqo/core/inference/device_manager.py
papa99do: >  An error here won't crash the container, it'll cause a 500 API response which is fine for an unhandled situation

If this endpoint is used by scheduler, we will need to be cautious. A error might be recoverable or transient. If we restart the container on these errors, we might introduce unnecessary downtime. (Please note if critical error like OOM or cuda unavailable happens, there's already downtime, so it's ok for rescheduler to shut it down and reschedule).

src/marqo/tensor_search/api.py
papa99do: liveness check endpoint should be lightweight, and probably should not return a lot of information. We have a separate endpoint `/device/cuda`  for more detailed information, we could add a `verbose` parameter to include more info in that endpont..

tests/core/inference/test_device_manager.py
papa99do: added

src/marqo/core/inference/device_manager.py
vicilliar: That's a fair point, as long as cloud team consistently looks at these warnings to make sure no we're not silently failing often

vicilliar (approved):

PR #1062, opened 2024-12-08
Fix a bug where attributes_to_retrieve does not work for non-string fields in unstructured indexes created with Marqo 2.13

papa99do:
Bug fix

There are three issues for unstructured index
image

- See the second row and last colume, we'll fix the regression for 2.13 unstructured index, and fix the missing int/float map in legacy unstructured index, but keep the map fields flattened for compatibility reasons. 

Yes, if the user's client code depends on the flattened map fields, they will need to change it since the new behaviour will return maps as is. 

Yes

No

No

* **Other information**:

### DIFF
--- a/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
+++ b/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
@@ -44,14 +44,17 @@ class SemiStructuredVespaIndex(StructuredVespaIndex, UnstructuredVespaIndex):
     def to_vespa_query(self, marqo_query: MarqoQuery) -> Dict[str, Any]:
         # Verify attributes to retrieve, if defined
         if marqo_query.attributes_to_retrieve is not None:
-            # Always retrieve static fields content to extract non-string values from combined fields
-            marqo_query.attributes_to_retrieve.extend([
-                common.VESPA_FIELD_ID,
-                common.STRING_ARRAY,
-                common.INT_FIELDS,
-                common.FLOAT_FIELDS,
-                common.BOOL_FIELDS,
-            ])
+            if len(marqo_query.attributes_to_retrieve) > 0:
+                # Retrieve static fields content to extract non-string values from combined fields
+                marqo_query.attributes_to_retrieve.extend([
+                    common.STRING_ARRAY,
+                    common.INT_FIELDS,
+                    common.FLOAT_FIELDS,
+                    common.BOOL_FIELDS,
+                ])
+
+            marqo_query.attributes_to_retrieve.append(common.VESPA_FIELD_ID)
+
             # add chunk field names for tensor fields
             marqo_query.attributes_to_retrieve.extend(
                 [self.get_marqo_index().tensor_field_map[att].chunk_field_name
--- a/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
+++ b/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
@@ -44,14 +44,20 @@ class SemiStructuredVespaIndex(StructuredVespaIndex, UnstructuredVespaIndex):
     def to_vespa_query(self, marqo_query: MarqoQuery) -> Dict[str, Any]:
         # Verify attributes to retrieve, if defined
         if marqo_query.attributes_to_retrieve is not None:
-            marqo_query.attributes_to_retrieve.append(common.VESPA_FIELD_ID)
+            # Always retrieve static fields content to extract non-string values from combined fields
+            marqo_query.attributes_to_retrieve.extend([
+                common.VESPA_FIELD_ID,
+                common.STRING_ARRAY,
+                common.INT_FIELDS,
+                common.FLOAT_FIELDS,
+                common.BOOL_FIELDS,
+            ])
             # add chunk field names for tensor fields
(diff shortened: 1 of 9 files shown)


src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
```diff
@@ -44,14 +44,20 @@ def to_marqo_document(self, vespa_document: Dict[str, Any], return_highlights: b
     def to_vespa_query(self, marqo_query: MarqoQuery) -> Dict[str, Any]:
         # Verify attributes to retrieve, if defined
         if marqo_query.attributes_to_retrieve is not None:
-            marqo_query.attributes_to_retrieve.append(common.VESPA_FIELD_ID)
+            # Always retrieve static fields content to extract non-string values from combined fields
```
wanliAlex: Will this cause search latency regression if they set `attributes_to_retrieve=[]`?

papa99do: Good point. I'll only add static fields when attributes_to_retrieve is not empty list. This will be a perf improvement comparing with the legacy unstructured where we always retrieve all fields.

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
```diff
@@ -44,14 +44,23 @@ def to_marqo_document(self, vespa_document: Dict[str, Any], return_highlights: b
...
+                # Retrieve static fields content to extract non-string values from combined fields
+                marqo_query.attributes_to_retrieve.extend([
+                    common.STRING_ARRAY,
+                    common.INT_FIELDS,
+                    common.FLOAT_FIELDS,
+                    common.BOOL_FIELDS,
+                ])
+
             marqo_query.attributes_to_retrieve.append(common.VESPA_FIELD_ID)
+
```
wanliAlex, line 57: To confirm, if we have `attributes_to_retrieve=["_id"]`, there won't be any error right?

papa99do: @wanliAlex No, I've tested it. And the field name in vespa is `marqo__id`, it will be retrieved anyhow and converged to `_id` field in marqo document. But having `_id` in the attributes_to_retrieve does trigger the other static field to be added. It's not ideal, but I think it's acceptable for now?

wanliAlex: Yeah i think it's acceptable.

wanliAlex (approved):
LGTM

tests/tensor_search/integ_tests/test_get_documents_by_ids.py
```diff
@@ -33,15 +27,29 @@ def setUpClass(cls) -> None:
             fields=[
                 FieldRequest(name='title1', type=FieldType.Text),
                 FieldRequest(name='desc2', type=FieldType.Text),
+                FieldRequest(name='int_field', type=FieldType.Int),
```
farshidz, line 30: We need to cover all data types for structured

farshidz: Likewise for search

src/marqo/core/unstructured_vespa_index/unstructured_document.py
```diff
@@ -23,7 +23,7 @@ class UnstructuredVespaDocumentFields(MarqoBaseModel):
     score_modifiers_fields: Dict[str, Any] = Field(default_factory=dict, alias=unstructured_common.SCORE_MODIFIERS)
     vespa_chunks: List[str] = Field(default_factory=list, alias=unstructured_common.VESPA_DOC_CHUNKS)
     vespa_embeddings: Dict[str, Any] = Field(default_factory=dict, alias=unstructured_common.VESPA_DOC_EMBEDDINGS)
-    vespa_multimodal_params: Dict[str, str] = Field(default_factory=str,
+    vespa_multimodal_params: Dict[str, str] = Field(default_factory=dict,
```
farshidz: Please check all reference of `vespa_multimodal_params` and make sure we don't rely on the default value (if we do, is it assuming str?)

farshidz (changes requested):

tests/tensor_search/integ_tests/test_get_documents_by_ids.py
papa99do: Added test cases for all field types of structured (both search and get_by_ids)

src/marqo/core/unstructured_vespa_index/unstructured_document.py
papa99do: I'll change it back to avoid introducing new regressions. Will change it to use the correct default value (or make it optional) in the next refactoring.

wanliAlex: Your fix is correct. It was a mistake I made before. It is working because there is a check [here](https://github.com/marqo-ai/marqo/blob/fa0d695f7b03227caad6aab5fbed7b79b4dc06b0/src/marqo/core/unstructured_vespa_index/unstructured_document.py#L190) so even if it's an empty string there isn't any issue. It's fine that you leave it as it is and I will fix it.

farshidz (approved):

PR #1068, opened 2024-12-12
Ensure the OOM error is captured in cuda health check

papa99do:
Bug fix

- OOM issue might be missed since we only try to create a small tensor which fits in the reserved memory

- We try to create a larger tensor (20MiB) which does not fit in reserved memory. It always triggeres memory allocation which reveals OOM issue consistently. 

No

Yes

No

No

* **Other information**:

### DIFF
--- a/src/marqo/core/inference/device_manager.py
+++ b/src/marqo/core/inference/device_manager.py
@@ -82,20 +82,30 @@ class DeviceManager:
 
         oom_errors = []
         for device in self.cuda_devices:
+            memory_stats = None
             try:
                 cuda_device = torch.device(f'cuda:{device.id}')
                 memory_stats = torch.cuda.memory_stats(cuda_device)
                 logger.debug(f'CUDA device {device.full_name} with total memory {device.total_memory}. '
                              f'Memory stats: {str(memory_stats)}')
 
-                torch.randn(3, device=cuda_device)
+                # Marqo usually allocates 20MiB cuda memory at a time when processing media files. When OOM happens,
+                # there are usually a few MiB free in reserved memory. To consistently trigger OOM from this health
+                # check request, we try to create a tensor that won't fit in the reserved memory to always trigger
+                # another allocation. Please note this is not leaky since we don't store the reference to this tensor.
+                # Once it's out of scope, the memory it takes will be returned to reserved space. Our assumption is
+                # that this method will not be called too often or with high concurrency (only used by liveness check
+                # for now which run once every few seconds).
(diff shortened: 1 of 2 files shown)


src/marqo/core/inference/device_manager.py
```diff
@@ -82,20 +82,30 @@ def cuda_device_health_check(self) -> None:
...
                 cuda_device = torch.device(f'cuda:{device.id}')
                 memory_stats = torch.cuda.memory_stats(cuda_device)
                 logger.debug(f'CUDA device {device.full_name} with total memory {device.total_memory}. '
                              f'Memory stats: {str(memory_stats)}')
 
-                torch.randn(3, device=cuda_device)
+                # Marqo usually allocates 20MiB cuda memory at a time when processing media files. When OOM happens,
+                # there are usually a few MiB free in reserved memory. To consistently trigger OOM from this health
+                # check request, we try to create a tensor that won't fit in the reserved memory to always trigger
+                # another allocation. Please note this is not leaky since we don't store the reference to this tensor.
```
farshidz, line 95: How sure are we though? Torch memory behavior has been nothing if not unpredictable

src/marqo/core/inference/device_manager.py
```diff
@@ -82,20 +82,30 @@ def cuda_device_health_check(self) -> None:
...
+                # Once it's out of scope, the memory it takes will be returned to reserved space. Our assumption is
+                # that this method will not be called too often or with high concurrency (only used by liveness check
+                # for now which run once every few seconds).
+                tensor_size = int(20 * 1024 * 1024 / 4)  # 20MiB, float32 (4 bytes)
+                torch.randn(tensor_size, device=cuda_device)
```
farshidz, line 100: How long does this operation take? If this were called very frequently would it affect other APIs?

farshidz: I would try spamming this to see what happens

papa99do: I've spammed this endpoint with 40 rps for a few min, and the reserved memory only increased 20MiB in the beginning and stayed at the same high watermark for the rest of the test. So I think it should be safe to use this endpoint only for liveness check. (please note this endpoint is not exposed to our customer via RP, I got 404 when directly send requests to it via the Cloud url)

src/marqo/core/inference/device_manager.py
papa99do: I didn't observe any cuda memory leak during my stress test with 40 request per second for several minutes on /healthz endpoint.

farshidz (approved):

PR #1071, opened 2024-12-16
Multinode-HA Vespa Setup for Local Testing

vicilliar:
Testing improvement

Current vespa setup only uses a single node. 

We implement a multinode setup for local vespa, so we can simulate cloud shards and replicas.
`vespa_local.py start` function now accepts `--Shards` and `--Replicas` as parameters. If Shards > 1 or Replicas > 0, multinode vespa setup is used. Multinode vespa setup has 3 config server nodes, max(2, total_content_nodes / 4) API nodes, and shards * (1 + replicas) content nodes.
Unit test github workflow now accepts shards and replicas as parameters.
Orchestrator workflow was created, which runs 4 unit tests setups:
(1) 0 replicas, 1 shard
(2) 1 replica, 1 shard
(3) 0 replicas, 2 shards
(4) 1 replica, 2 shards

Unit tests on multinode vespa will ignore the following directories: tests/integ_tests/core/inference, tests/integ_tests/processing, tests/integ_tests/s2_inference

Multinode vespa tests will use m6i.2xlarge instead of m6i.xlarge due to the higher memory usage from many vespa nodes. Config and API nodes are ~1gb and content nodes are ~500mb. A 9 node system (3 config, 2 API, 4 content) needs roughly 7gb for vespa alone.

No

In progress

* **Other information**:


.github/workflows/unit_test_200gb_CI.yml
```diff
@@ -14,14 +41,48 @@ on:
...
+            # Convert inputs to integers
+            NUMBER_OF_SHARDS_INT=$(echo "$NUMBER_OF_SHARDS" | awk '{print int($0)}')
+            NUMBER_OF_REPLICAS_INT=$(echo "$NUMBER_OF_REPLICAS" | awk '{print int($0)}')
+
+            # Evaluate the conditions
+            if [[ "$NUMBER_OF_SHARDS_INT" -gt 1 || "$NUMBER_OF_REPLICAS_INT" -gt 0 ]]; then
+              echo "Using multi-node Vespa setup. Shards are $NUMBER_OF_SHARDS_INT and replicas are $NUMBER_OF_REPLICAS_INT."
+              echo "VESPA_MULTINODE_SETUP=true" >> $GITHUB_OUTPUT
+              # If multinode vespa, ignore unrelated tests to save time
+              echo "MULTINODE_TEST_ARGS=--multinode --ignore=tests/core/index_management/test_index_management.py --ignore=tests/core/inference --ignore=tests/processing --ignore=tests/s2_inference" >> $GITHUB_OUTPUT
```
papa99do: We should exclude all the unit tests that do not use Vespa to save some time (some long running tests do not use vespa at all). Let's create a ticket to track this improvement.

.github/workflows/unit_test_200gb_CI.yml
```diff
@@ -135,18 +202,53 @@ jobs:
...
 
           # Deploy application with test schema
           curl --header "Content-Type:application/zip" --data-binary @vespa_tester_app.zip http://localhost:19071/application/v2/tenant/default/prepareandactivate
-
-          # wait for vespa to start (document url):
-          timeout 10m bash -c 'until curl -f -X GET $VESPA_DOCUMENT_URL >/dev/null 2>&1; do echo "  Waiting for Vespa document API to be available..."; sleep 10; done;' || \
-            (echo "Vespa (Document URL) did not start in time" && exit 1)
-
-          echo "Vespa document API is available. Local Vespa setup complete."
+
```
papa99do: We can consolidate this by querying the convergence status from CONFIG_URL in an interval. If the the cluster is convergent after deploying the vespa app package, then all nodes must be up and running.

.github/workflows/unit_test_200gb_CI.yml
```diff
@@ -165,7 +267,7 @@ jobs:
 
           cd marqo
           export PYTHONPATH="./tests:./src:."
-          pytest --ignore=tests/test_documentation.py --ignore=tests/backwards_compatibility_tests --durations=100 --cov=src --cov-branch --cov-context=test --cov-report=html:cov_html --cov-report=lcov:lcov.info tests
+          pytest ${{ env.MULTINODE_TEST_ARGS }} --ignore=tests/test_documentation.py --ignore=tests/backwards_compatibility_tests --durations=100 --cov=src --cov-branch --cov-context=test --cov-report=html:cov_html --cov-report=lcov:lcov.info tests
```
papa99do: This is outdated, please merge the mainline change first.

.github/workflows/unit_tests_with_shards_and_replicas.yml
```diff
@@ -0,0 +1,52 @@
...
+    types: [submitted]
+    branches:
+      - mainline
+      - 'releases/*'
+
+permissions:
+  contents: read
+
+jobs:
+  Unit-Tests-1-Shard-0-Replica:
```
papa99do, line 22: Is this a duplicate run?

.github/workflows/unit_test_200gb_CI.yml
```diff
@@ -135,18 +202,53 @@ jobs:
           done
           echo -e "\nDone waiting."
 
-          # Zip up schemas and services
+          # Zip up schemas, hosts, and services
           sudo apt-get install zip -y
-          zip -r vespa_tester_app.zip services.xml schemas
+          zip -r vespa_tester_app.zip . -x README.md .env .gitignore "*.yml"  # zip the whole directory except for the listed files
```
papa99do: If we don't want to keep the zip file, we can pipe the output of zip to curl command. In this case, we don't need to clean up the files or even copy files around. Example:
`zip -r - . -x README.md .env .gitignore "*.yml" |  curl --header Content-Type:application/zip --data-binary @- http://localhost:19071/application/v2/tenant/default/prepareandactivate`

scripts/vespa_local/vespa_local.py
```diff
@@ -1,22 +1,417 @@
 import argparse
 
 import os
+import yaml
+import docker
```
papa99do, line 24: We will need these deps in `requirements.dev.txt` file

scripts/vespa_local/vespa_local.py
```diff
@@ -1,22 +1,417 @@
...
+                  "--publish 8080:8080 --publish 19071:19071 --publish 2181:2181 --publish 127.0.0.1:5005:5005 "
+                  f"vespaengine/vespa:{VESPA_VERSION}")
+
+        # Copy services.xml to base dir. It can be zipped from here.
+        os.system("cp singlenode/services.xml services.xml")
+        os.system("rm -f hosts.xml")
+        print("Copied services.xml to base directory. Removed hosts.xml.")
+
+
+class VespaLocalMultiNode:
```
papa99do: The logic in this class is worth testing (to make sure the generated docker compose file and config files are as expected). Can we write a test for this class?

.github/workflows/unit_test_200gb_CI.yml
vicilliar: I agree, although our test suite now is quite confusing, and many tests that don't use vespa are interwoven with ones that do, with no concrete structure to differentiate them. We should have these tests manually tagged. It also adds overhead for future tests made, to always be marked with either `@pytest.mark.uses_vespa` or `@pytest.mark.no_vespa`

.github/workflows/unit_tests_with_shards_and_replicas.yml
vicilliar: This is the same as a normal unit test run, good point. If we intend to keep running standard unit tests I'll remove this

.github/workflows/unit_test_200gb_CI.yml
vicilliar: Ticket created: https://s2search.atlassian.net/jira/polaris/projects/MOSD/ideas/view/4863474?selectedIssue=MOSD-318&issueViewLayout=sidebar&issueViewSection=overview&atlOrigin=eyJpIjoiM2ZkNjRkYTUzZjljNGZhOGE3MmNmMDdmYzJhYzcwOTEiLCJwIjoiaiJ9

.github/workflows/unit_test_200gb_CI.yml
vicilliar: done

.github/workflows/unit_test_200gb_CI.yml
vicilliar: done

scripts/vespa_local/vespa_local.py
vicilliar: yaml is already a requirement here in marqo-base: https://github.com/marqo-ai/marqo-base/blob/main/requirements/amd64-gpu-requirements.txt#L323

vicilliar: docker is already a dependency of one of our reqs in requirements.dev.txt, but I'll add both docker and PyYaml with pinned versions to requirements.dev.txt just in case

scripts/vespa_local/vespa_local.py
vicilliar: done

.github/workflows/unit_test_200gb_CI.yml
vicilliar: done

papa99do (dismissed):
Look great. Thanks for adding the unit test to make sure the compose file and configs are generated correctly.

papa99do (approved):

.github/workflows/unit_test_200gb_CI.yml
```diff
@@ -16,13 +43,45 @@ on:
...
 concurrency:
-  group: unit-tests-${{ github.ref }}
+  group: unit-tests-${{ github.ref }}-${{ inputs.number_of_shards }}-${{ inputs.number_of_replicas }}
   cancel-in-progress: true
 
 permissions:
   contents: read
 
 jobs:
+  Determine-Vespa-Setup:
```
papa99do: This step should be run after `Check-Changes`, and should be run only if check-changes returns true:
`if: ${{ needs.Check-Changes.outputs.doc_only == 'false' }} # Run only if there are non-documentation changes`

.github/workflows/unit_test_200gb_CI.yml
```diff
@@ -224,7 +282,7 @@ jobs:
           cd marqo
           export PYTHONPATH="./tests:./src:."
           set -o pipefail
-          pytest --ignore=tests/test_documentation.py --ignore=tests/compatibility_tests \
+          pytest ${{ env.MULTINODE_TEST_ARGS }} --ignore=tests/test_documentation.py --ignore=tests/compatibility_tests \
```
papa99do: It seems MULTINODE_TEST_ARGS is not passed in correctly (or maybe is not populated correctly in the first place?)

Also, in the next line, we fail the build if `--cov-fail-under=69`, which does not make sense for these tests since they skip a lot of test cases. we should skip the coverage check in multi-shard/replica tests.

papa99do (changes requested):

.github/workflows/unit_test_200gb_CI.yml
vicilliar: It is passed in correctly for multinode runs. Please check this 2 shard 1 replica run: https://github.com/marqo-ai/marqo/actions/runs/13106217962/job/36561470973#step:9:15

MULTINODE_TEST_ARGS will be empty string for 1 shard and 0 replicas. Maybe that's the one you saw.

papa99do (dismissed):

wanliAlex (approved):

papa99do (approved):

PR #1075, opened 2024-12-18
Preserving document-process node in services.xml when bootstrapping Vespa app

papa99do:
Bug fix

document-process node is removed from container section of services.xml when bootstrapping Vespa app using Marqo 2.13+. This caused some failures when upgrading some old Cloud indexes. Please see [this slack thread](https://marqo-ai.slack.com/archives/C03L2FWTGLE/p1734439370330959) for more details

* document-process node is preserved (but the attributes and child nodes will be cleared) in container section of services.xml. This is to make sure the bootstrap process is compatible with indexes created with IndexCreateWorkflowV2, which uses a services.xml template containing an empty `document-processing` node. 

No

Yes, but the PR didn't trigger the build somehow. I manually run large model tests and unit tests pipeline.
https://github.com/marqo-ai/marqo/actions?query=branch%3Ayihan%2Fadd-doc-processing-to-services-xml

No

No

* **Other information**:

### DIFF
--- a/src/marqo/core/index_management/vespa_application_package.py
+++ b/src/marqo/core/index_management/vespa_application_package.py
@@ -86,7 +86,7 @@ class ServicesXml:
         """
         container_element = self._ensure_only_one('container')
         for child in container_element.findall('*'):
-            if child.tag in ['document-api', 'search']:
+            if child.tag in ['document-api', 'document-processing', 'search']:
                 child.clear()
             elif child.tag != 'nodes':
                 container_element.remove(child)
--- a/tests/core/index_management/test_services_xml.py
+++ b/tests/core/index_management/test_services_xml.py
@@ -201,6 +201,10 @@ class TestIndexSettingStore(unittest.TestCase):
         self._assertStringsEqualIgnoringWhitespace(expected_xml, service_xml.to_xml())
 
     def test_config_components_should_preserve_document_processing_nodes(self):
+        """
+        This test case tests that document-processing node in the `container` section is preserved when configuring
+        components since it is usually referenced in the `content` section, and we do not change content section.
+        """
         xml = """<?xml version="1.0" encoding="utf-8" ?>
                             <services version="1.0" xmlns:deploy="vespa" xmlns:preprocess="properties">
                                 <container id="default" version="1.0">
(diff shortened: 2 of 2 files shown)


tests/core/index_management/test_services_xml.py
```diff
@@ -200,6 +200,67 @@ def test_config_components(self):
                 """
         self._assertStringsEqualIgnoringWhitespace(expected_xml, service_xml.to_xml())
 
+    def test_config_components_should_preserve_document_processing_nodes(self):
```
farshidz, line 203: Add a comment as to why we're testing this

farshidz:
I suggest asking @Mehul-Porwal do double-check that the test accurately reflects the problematic services.xml

tests/core/index_management/test_services_xml.py
papa99do: Will do

papa99do: Just to confirm I should just add a generic comment instead of referring to the Cloud issue, right?

Mehul-Porwal (approved):

src/marqo/core/index_management/vespa_application_package.py
```diff
@@ -86,7 +86,7 @@ def _cleanup_container_config(self):
         """
         container_element = self._ensure_only_one('container')
         for child in container_element.findall('*'):
-            if child.tag in ['document-api', 'search']:
+            if child.tag in ['document-api', 'document-processing', 'search']:
                 child.clear()
```
Mehul-Porwal, line 90: Why clear the element if we are not regenerating it? Shouldn't we just keep it unchanged?

Mehul-Porwal (changes requested):

Mehul-Porwal (approved):

farshidz (approved):

src/marqo/core/index_management/vespa_application_package.py
papa99do: > Shouldn't we just leave it unchanged?

It seems we are doing destructive changes here. But the logic behind it is the same as deleting all other components, to make sure all config is controlled by Marqo. For Cloud, it should be the same since we don't config document-processing anyways. It wouldn't be more risky for OS users either since the previous logic is to delete this element.  On the other hand, clearing it makes it simple to change its config in the future since we do not need complicated logic to check the existing config (random) and decide what to keep and what to override.

PR #1086, opened 2025-01-03
Fix unstructured index infer modality logic

wanliAlex:
bug

  - In an unstructured index, even if a field is not a tensor field, Marqo still infers its modality.

  - In an unstructured index, even if the user sets `treat_urls_and_pointers_as_images=False`  and `treat_urls_and_pointers_as_media=False`, Marqo still tries to infer the modality based on the URL content instead of making it text directly.

  - In an unstructured index, Marqo will not infer the modality of a field if it's not a tensor field.

  - In an unstructured index, if the user sets `treat_urls_and_pointers_as_images=False`  and `treat_urls_and_pointers_as_media=False`, Marqo will treat the URL as a text now.

* **Other information**:


src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
```diff
@@ -65,30 +63,71 @@ def _validate_doc(self, doc):
             except api_errors.InvalidArgError as err:
                 raise AddDocumentsError(err.message, error_code=err.code, status_code=err.status_code) from err
 
+    def _collect_tensor_field_content(self, marqo_doc: dict, field_name: str, field_content: Any) -> Any:
+        """Collect the tensor field content. If the field need to be processed as a tensor field, the content will be
+        collected by the tensor_fields_container. Otherwise, the content will be returned as is.
+        """
+        if field_name in self.tensor_fields_container._tensor_fields or field_name in \
+            self.tensor_fields_container._multimodal_sub_field_reverse_map:
```
papa99do: `_tensor_fields` and `_multimodal_sub_field_reverse_map` are private fields of the `tensor_fields_container` object. And this check has been down already in `collect` method. Maybe we can pass in a function to the collect method to do type inference instead of the inferred type.

wanliAlex: Fixed

src/marqo/core/inference/tensor_fields_container.py
```diff
@@ -493,7 +493,18 @@ def populate_tensor_from_existing_doc(self, existing_marqo_doc: Dict[str, Any],
             tensor_content.populate_chunks_and_embeddings(existing_tensor[constants.MARQO_DOC_CHUNKS],
                                                           existing_tensor[constants.MARQO_DOC_EMBEDDINGS])
 
-    def collect(self, doc_id: str, field_name: str, field_content: Any, field_type: Optional[FieldType]) -> Any:
+    def collect(self, doc_id: str, field_name: str, field_content: Any, infer_field_type: Callable) -> Any:
```
papa99do: nit. we can be more specific about the callable's type: `Callable[[str, Any], FieldType]`, so the IDE could leverage this to do more static checks for us.

papa99do (dismissed):
LGTM

src/marqo/core/inference/tensor_fields_container.py
wanliAlex: done

papa99do (approved):
LGTM

PR #1091, opened 2025-01-16
Improve the performance of processing large search response

papa99do:
performance improvement

 Processing large search responses (such as 1000 large document) is suboptimal in Marqo. 

We did the following improvements to speed up the large response processing:
* skip pydantic validation when converting vespa docs to marqo docs. 
* use orjson for json serde of search response
* improved the performance of selecting fields post search for unstructured index

No

Yes

No

No

* **Other information**:

### DIFF
--- a/.github/workflows/largemodel_unit_test_CI.yml
+++ b/.github/workflows/largemodel_unit_test_CI.yml
@@ -87,6 +87,7 @@ jobs:
         run: |
           pip install -r marqo-base/requirements.txt
           # override base requirements with marqo requirements, if needed: 
+          pip install -r marqo/requirements.txt
           pip install -r marqo/requirements.dev.txt
           pip install pytest==7.4.0
 
--- a/.github/workflows/largemodel_unit_test_CI.yml
+++ b/.github/workflows/largemodel_unit_test_CI.yml
@@ -80,6 +80,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: marqo-ai/marqo-base
+          ref: 'releases/2.13'
           path: marqo-base
 
       - name: Install dependencies
--- a/.github/workflows/largemodel_unit_test_CI.yml
+++ b/.github/workflows/largemodel_unit_test_CI.yml
@@ -85,7 +85,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install -r marqo-base/requirements/amd64-gpu-requirements.txt
+          pip install -r marqo-base/requirements.txt
           # override base requirements with marqo requirements, if needed: 
           pip install -r marqo/requirements.dev.txt
           pip install pytest==7.4.0
--- a/src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
+++ b/src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
@@ -62,7 +62,7 @@ class SemiStructuredVespaDocument(MarqoBaseModel):
                 text_fields[field_name] = fields[field_name]
 
         return cls(id=document[cls._VESPA_DOC_ID],
-                   fixed_fields=SemiStructuredVespaDocumentFields(**fields),
+                   fixed_fields=SemiStructuredVespaDocumentFields.construct(**fields),
                    tensor_fields=tensor_fields,
                    text_fields=text_fields,
                    raw_tensor_score=cls.extract_field(fields, common.VESPA_DOC_HYBRID_RAW_TENSOR_SCORE, None),
--- a/src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
+++ b/src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
@@ -18,15 +18,13 @@ from marqo.core.unstructured_vespa_index.unstructured_vespa_index import Unstruc
 from marqo.core.vespa_index.add_documents_handler import AddDocumentsHandler, AddDocumentsError
 from marqo.s2_inference.errors import MediaDownloadError
 from marqo.s2_inference.multimodal_model_load import infer_modality, Modality
-
-from marqo.vespa.models import VespaDocument
-from marqo.vespa.models.get_document_response import Document
-from marqo.vespa.vespa_client import VespaClient
-
 # TODO deps to tensor_search needs to be removed
 from marqo.tensor_search.constants import ALLOWED_UNSTRUCTURED_FIELD_TYPES
 from marqo.tensor_search.validation import validate_custom_vector, \
     validate_map_numeric_field
+from marqo.vespa.models import VespaDocument
+from marqo.vespa.models.get_document_response import Document
+from marqo.vespa.vespa_client import VespaClient
 
 
 class UnstructuredAddDocumentsHandler(AddDocumentsHandler):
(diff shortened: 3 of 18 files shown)


.github/workflows/largemodel_unit_test_CI.yml
```diff
@@ -80,12 +80,14 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: marqo-ai/marqo-base
+          ref: 'releases/2.13'
```
papa99do, line 83: revert back to 2.13 snapshot of marqo-base for more consistent builds on releases/2.13 branch

src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
```diff
@@ -62,7 +62,7 @@ def from_vespa_document(cls, document: Dict, marqo_index: SemiStructuredMarqoInd
                 text_fields[field_name] = fields[field_name]
 
         return cls(id=document[cls._VESPA_DOC_ID],
-                   fixed_fields=SemiStructuredVespaDocumentFields(**fields),
+                   fixed_fields=SemiStructuredVespaDocumentFields.construct(**fields),
```
papa99do, line 65: Optimisation 1: skip validation when constructing the SemiStructuredVespaDocumentFields model from vespa fields dict. This saves about 100ms when dealing with large document set (1000 docs)

src/marqo/tensor_search/api.py
```diff
@@ -284,6 +290,7 @@ def search(search_query: SearchQuery, index_name: str, device: str = Depends(api
             text_query_prefix=search_query.textQueryPrefix,
             hybrid_parameters=search_query.hybridParameters
         )
+        return ORJSONResponse(result)
```
papa99do, line 287: [Optimisation]: use orjson to generate the search response

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1778,7 +1778,7 @@ def gather_documents_from_response(response: QueryResult, marqo_index: MarqoInde
     vespa_index = vespa_index_factory(marqo_index)
     hits = []
     for doc in response.hits:
-        marqo_doc = vespa_index.to_marqo_document(doc.dict(), return_highlights=highlights)
+        marqo_doc = vespa_index.to_marqo_document(dict(doc), return_highlights=highlights)
```
papa99do, line 1792: [Optimisation]: bypassing validation and alias handling when converting the vespa doc pydantic model to a dict. this saves about 150ms when converting a large doc set

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1797,12 +1797,13 @@ def gather_documents_from_response(response: QueryResult, marqo_index: MarqoInde
...
     # attributes_to_retrieve should already be validated at the start of search
-    attributes_to_retrieve = list(set(attributes_to_retrieve).union({"_id", "_score", "_highlights"}))
+    attributes_to_retrieve_as_prefixes = set([attribute + "." for attribute in attributes_to_retrieve])
+    attributes_to_retrieve = set(attributes_to_retrieve).union({"_id", "_score", "_highlights"})
     return {k: v for k, v in marqo_doc.items() if k in attributes_to_retrieve or
             # Please note that numeric map fields are flattened for unstructured or semi-structured indexes.
             # Therefore, when filtering on attributes_to_retrieve, we need to also include flattened map fields
             # with the specified attributes as prefixes. We keep this behaviour only for compatibility reasons.
-            any([k.startswith(attribute + ".") for attribute in attributes_to_retrieve])}
+            '.' in k and any([k.startswith(prefix) for prefix in attributes_to_retrieve_as_prefixes])}
```
papa99do: [Optimisation]: pre-construct the attributes_to_retrieve set and prefix set to reduce the time of `in` check. Also check if the field has `.` before checking prefix.

src/marqo/vespa/vespa_client.py
```diff
@@ -245,7 +246,7 @@ def query(self, yql: str, hits: int = 10, ranking: str = None, model_restrict: s
 
         self._query_raise_for_status(resp)
 
-        return QueryResult(**resp.json())
+        return QueryResult(**orjson.loads(resp.text))
```
papa99do, line 249: [Optimisation]: Use orjson to load the vespa response for query request

src/marqo/tensor_search/api.py
OwenElliottDev: Any reason not to use the provided one? https://fastapi.tiangolo.com/reference/responses/#fastapi.responses.ORJSONResponse

src/marqo/tensor_search/tensor_search.py
OwenElliottDev: Why not create these sets in the caller and pass them into `unstructured_index_attributes_to_retrieve`? Avoid repeating this work for every document.

src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
```diff
@@ -67,28 +65,52 @@ def _validate_doc(self, doc):
...
+            except MediaDownloadError as err:
+                raise AddDocumentsError(err.message) from err
+            if ((self.marqo_index.treat_urls_and_pointers_as_media is False) and modality in
+                    [Modality.AUDIO, Modality.VIDEO]):
                 modality = Modality.TEXT
-
             return MODALITY_FIELD_TYPE_MAP[modality]
-        except MediaDownloadError as err:
-            raise AddDocumentsError(err.message) from err
+        else:
```
OwenElliottDev, line 112: else not needed

wanliAlex (approved):

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1797,12 +1797,13 @@ def gather_documents_from_response(response: QueryResult, marqo_index: MarqoInde
 def unstructured_index_attributes_to_retrieve(marqo_doc: Dict[str, Any], attributes_to_retrieve: List[str]) -> Dict[
     str, Any]:
     # attributes_to_retrieve should already be validated at the start of search
-    attributes_to_retrieve = list(set(attributes_to_retrieve).union({"_id", "_score", "_highlights"}))
+    attributes_to_retrieve_as_prefixes = set([attribute + "." for attribute in attributes_to_retrieve])
```
farshidz: This method
* Needs a docstring as name isn't very intuitive 
* Must be unit tested

src/marqo/tensor_search/tensor_search.py
farshidz: I think we can do better here. Why don't we put those prefixes in a set to avoid a linear time search (which makes the loop polynomial right now)?

papa99do: OK, I got your point, prefixes are already in a set. I'll need to do split of the k and compare it with the prefix. I'll make this change.

OwenElliottDev: @papa99do something like this should work
```python
def unstructured_index_attributes_to_retrieve(marqo_doc: Dict[str, Any], attributes_to_retrieve: List[str]) -> Dict[str, Any]:
    required = set(attributes_to_retrieve).union({"_id", "_score", "_highlights"})
    result = {}
    for k, v in marqo_doc.items():
        if k in required:
            result[k] = v
        elif '.' in k and k.split('.', 1)[0] in required:
            result[k] = v
    return result
```

src/marqo/core/unstructured_vespa_index/unstructured_add_document_handler.py
papa99do: Why isn't else required?

OwenElliottDev: `else: return` is the same as just `return`. Anything after that `if`/`else` will be unreachable.

OwenElliottDev: It's just a tidiness/readability thing.

farshidz (approved):

PR #1095, opened 2025-01-22
Skip unit tests when only making documentation changes

wanliAlex:
workflow optimize

we need to run the unit tests even if it is just a documentation (`.md`) file change.

We still need to run the unit tests workflow as there is a check.
However, we skip the real run and return pass directly.

* **Other information**:

### DIFF
--- a/.github/workflows/unit_test_200gb_CI.yml
+++ b/.github/workflows/unit_test_200gb_CI.yml
@@ -46,7 +46,7 @@ jobs:
           echo "Head commit: $HEAD_COMMIT"
 
           # Perform the diff to check for non-documentation changes
-          if git diff --name-only $BASE_COMMIT $HEAD_COMMIT -- | grep -vE '\.(md|yml)$'; then
+          if git diff --name-only $BASE_COMMIT $HEAD_COMMIT -- | grep -vE '\.(md)$'; then
             echo "doc_only=false" >> $GITHUB_ENV
             echo "doc_only=false" >> $GITHUB_OUTPUT
           else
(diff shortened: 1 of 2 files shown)


.github/workflows/unit_test_200gb_CI.yml
```diff
@@ -54,9 +54,46 @@ jobs:
               {"Key": "PoloRole", "Value": "testing"}
             ]
 
+  Check-Changes:
+    runs-on: ubuntu-latest
```
papa99do: If this runs on `ubuntu-latest` runner, can we run this even before the start-ec2 job and skip all 3 jobs if only md files are changed?

wanliAlex: Brilliant idea! Done

.github/workflows/unit_test_200gb_CI.yml
```diff
@@ -21,9 +21,45 @@ permissions:
...
+          fetch-depth: 0
+          path: marqo
+
+      - name: Check for Documentation-Only Changes
+        id: set-output
+        run: |
+          cd marqo
+          set -x
+          # Get the base and head commits from the GitHub event
+          BASE_COMMIT=${{ github.event.pull_request.base.sha }}
```
papa99do, line 43: Did you try this on push event? triggering the test on the branch directly? Do you think we should skip unit test run for push event?  Maybe we can add the following to the push event trigger?
```
    paths-ignore:
      - '**.md'
```

wanliAlex: The push event only happens when a PR is merged into the mainline or release branches. Yeah, we can skip the test run. We are skipping other workflows right now.

papa99do (approved):
Looks good. Thanks @wanliAlex

PR #1108, opened 2025-01-30
[Releases 2.13] Output hybrid error message in marqo logs

vicilliar:
bug fix

any errors in tensor or lexical search in hybrid rrf search are swallowed by java code

output any tensor or lexical errors in marqo logs

no

* **Other information**:


vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -105,6 +106,8 @@ public Result search(Query query, Execution execution) {
                                 + e.toString());
             }
 
+            raiseErrorIfPresent(resultLexical, resultTensor);
```
papa99do: I was wondering if we should handle errors differently after reading Vespa docs. Instead of raising Runtime exception everywhere (which cause Vespa to always return 500). We should probably leverage HitGroup().addError method which is suggested in [this doc](https://docs.vespa.ai/en/searcher-development.html#error-handling).

For this specific scenario, we are collecting errors from both lexical result and tensor result. So we can do this:

```
HitGroup combinedErrors = new HitGroup();
combinedErrors.addErrorsFrom(resultTensor.hits());
combinedErrors.addErrorsFrom(resultLexical.hits());
if (combinedErrors.getError() != null) {
      return new Result(query, combinedErrors);
}
```

In this way, you should be able to collect all the errors from both result and allow search chain to handle the error properly, instead of just concatenating the first error found in each result.

papa99do (changes requested):
Please check with Vespa team about the best practices of handling errors in the search chain.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
vicilliar: Good find! Implemented this. However, vespa behavior seems to only raise the first error in the `ErrorHit`. So if our hit group has 2 errors (tensor was added first), the tensor error will always be raised in marqo.

vicilliar:
@papa99do Added a test with mock vespa result with 2 errors to show marqo handles this properly.

scripts/vespa_local/docker-compose.yml
```diff
@@ -0,0 +1,149 @@
+services:
```
papa99do: nit. are vespa config changes merged in from another PR?

vicilliar: Which vespa config changes are you referring to?

papa99do: These docker compose files and multinode configs

vicilliar: Ah yep! You're right, getting rid of these

tests/tensor_search/integ_tests/test_hybrid_search.py
```diff
@@ -1841,6 +1844,201 @@ def test_hybrid_search_unstructured_with_searchable_attributes_fails(self):
...
+                            {
+                                'code': 12,
+                                'summary': 'Timed out',
+                                'source': 'content_default',
+                                'message': "Error in execution of chain 'content_default': Chain timed out."
+                            }
+                        ]
+                    }
+                },
+                400,    # HTTP 400
```
papa99do, line 1982: not sure if this response code is expected if both errors are timeout.

tests/tensor_search/integ_tests/test_hybrid_search.py
```diff
@@ -1841,6 +1844,201 @@ def test_hybrid_search_unstructured_with_searchable_attributes_fails(self):
...
+                            {
+                                'code': 12,
+                                'summary': 'Timed out',
+                                'source': 'content_default',
+                                'message': "Error in execution of chain 'content_default': Chain timed out."
+                            }
+                        ]
+                    }
+                },
+                500,  # HTTP 500
```
papa99do: I think vespa would return a 400 in this case?

tests/tensor_search/integ_tests/test_hybrid_search.py
```diff
@@ -1841,6 +1844,201 @@ def test_hybrid_search_unstructured_with_searchable_attributes_fails(self):
...
+                                'summary': 'Invalid query parameter',
+                                'message': 'Could not create query from YQL.'
+                            }
+                        ]
+                    }
+                },
+                504,    # HTTP 504
+                False,  # Not a timeout, since 2nd error is not timeout
+            ),
+            # HTTP 504, second is Vespa 12
```
papa99do: nit. status code does not match that in test case.

tests/tensor_search/integ_tests/test_hybrid_search.py
vicilliar: good point

tests/tensor_search/integ_tests/test_hybrid_search.py
vicilliar: However, it may be possible for Vespa to return an error code 12 while not being a timeout (504). This is what I'm testing here (Marqo error should not end up as 504). The error messages themselves here are just placeholders. I can change it to something other than timeout.

papa99do (approved):

PR #1112, opened 2025-02-04
Enforce 95% unit test coverage for new code

farshidz:
## Change Summary
We currently have three workflows that cover all non-API tests:
* Integ tests except large model: All integ tests except large model, partitioned for a faster run so several .coverage files
* Largemodel tests: Large model tests, partitioned to run faster so several .coverage files
* Unit tests

There is no single coverage artifact that covers all integ tests or all integ and unit tests.

This PR introduced a parent workflow that runs all of the above, collects coverage files and creates combined integ test coverage and all tests coverage artifacts. We then require
* 75% combined test coverage
* 70% integ tests coverage
* 90% combined diff coverage
* 80% integ diff coverage
* 95% unit test diff coverage

## Related Jira Ticket
https://s2search.atlassian.net/browse/MOSD-180

## Checklist

- [N/A] Tests have been added for changes
- [ N/A] Breaking changes are clearly identified
- [ N/A] Python client changes linked or N/A

### For new field types:
- [N/A ] Test indexes updated to cover the new type for all APIs (add docs, search, partial update, etc.)


.github/workflows/unit_tests.yml
```diff
@@ -49,14 +50,49 @@ jobs:
...
+        continue-on-error: true
         run: |
           cd marqo
           export PYTHONPATH="./src"
-          pytest tests/unit_tests/ --durations=100 --cov=src --cov-branch --cov-context=test --cov-report=html:cov_html --cov-report=lcov:lcov.info
+          set -o pipefail
+          pytest --durations=100 --cov=src --cov-branch --cov-context=test \
+            --cov-report=html:cov_html --cov-report=xml:cov.xml --cov-report term:skip-covered \
+            --md-report --md-report-flavor gfm --md-report-output pytest_result_summary.md \
+            tests/unit_tests/ | tee pytest_output.txt
```
papa99do, line 54: What is the plan going forward? Will we set separate coverage threshold for unit tests and integration tests? Or do we combine the coverage report and set one threshold for all the tests.

.github/workflows/unit_tests.yml
```diff
@@ -49,14 +50,49 @@ jobs:
...
 
-      - name: Upload Test Report
+      - name: Check new code coverage
+        id: check_test_coverage
+        if: github.event_name == 'pull_request'
+        run: |
+          cd marqo
+          diff-cover cov.xml --compare-branch=origin/${{ github.base_ref }} --fail-under=95 \
```
papa99do: If unit test coverage is enforced to be above 95, do we still need to keep the same check in unit_test_200gb_CI workflow?

papa99do (dismissed):

.github/workflows/run_required_checks_coverage.yml
```diff
@@ -0,0 +1,255 @@
...
+          cd marqo
+          
+          export PYTHONPATH="."
+          export BASE_BRANCH="${{ github.event.pull_request.base.ref }}"
+          echo "Running diff-cover against branch $BASE_BRANCH"
+          git fetch origin $BASE_BRANCH:$BASE_BRANCH
+          
+          diff-cover ../coverage_data/coverage.xml --html-report ../coverage_data/diff_cov.html \
+            --markdown-report ../coverage_data/diff_cov.md \
+            --compare-branch $BASE_BRANCH --fail-under=90
```
papa99do: Is this check redundant? since unit_test coverage is already higher than 95%, so combined coverage will not be lower than 95%.

.github/workflows/unit_tests.yml
```diff
@@ -4,29 +4,21 @@ name: Run unit tests
...
-    branches:
-      - mainline
-      - releases/*
 
-concurrency:
-  group: unit-tests-${{ github.ref }}
-  cancel-in-progress: true
+# concurrency:
+#   group: unit-tests-${{ github.ref }}
+#   cancel-in-progress: true
```
papa99do: why do we remove the concurrency group here? Is it because unit test is cheaper to run, so we don't cancel the ongoing run when a new commit is pushed to a PR?

papa99do (commented):
It's much cleaner now. Nice work! left two questions.

papa99do:
nit. But the name of these tests should be `Integration Test with Shards and Replicas`
image

.github/workflows/run_required_checks_coverage.yml
farshidz: I guess so. I'll remove this

.github/workflows/unit_tests.yml
farshidz: That's an oversight. I removed these for easier debugging. I'll add this back

papa99do (dismissed):
LGTM

papa99do (approved):

PR #1124, opened 2025-02-26
add ability to change VESPA_DISK_USAGE_LIMIT

danyilq:
Ability to set `VESPA_DISK_USAGE_LIMIT` env variable (ranging 0 to 1).

Currently local disk usage limit is not configurable.

* **Other information**:


papa99do:
Looks good. This works when we run an external Vespa locally. It might not work for embedded Vespa. @farshidz , do you think we need to support embedded Vespa as well?

farshidz:
> Looks good. This works when we run an external Vespa locally. It might not work for embedded Vespa. @farshidz , do you think we need to support embedded Vespa as well?

Actually we just want it for the embedded Vespa. Sorry if I wasn't clear 
Basically if I do
```
docker run ... marqoai/marqo
```
Then I need to be able to pass an env var here to override the limit. This only needs to work with the built-in Vespa

Ignore if Vespa is external

run_marqo.sh
```diff
@@ -74,8 +74,10 @@ elif [ -z "$VESPA_QUERY_URL" ] && [ -z "$VESPA_DOCUMENT_URL" ] && [ -z "$VESPA_C
     if echo "$RESPONSE" | grep -q '"error-code":"NOT_FOUND"'; then
       echo "Marqo did not find an existing vector store. Setting up vector store..."
 
+      # Generate the application package
+      python3 /app/scripts/vespa_local/vespa_local.py generate-application-package
       # Deploy a dummy application package
-      vespa deploy /app/scripts/vespa_local --wait 300 >/dev/null 2>&1
+      vespa deploy vespa_dummy_application_package --wait 300 >/dev/null 2>&1
```
vicilliar: To be uniform, let's use the deploy command from vespa_local as well, since we have it. 
I'd prefer if we just make a new step in `vespa_local` like:
`python3 /app/scripts/vespa_local/vespa_local.py generate-and-deploy-application-package`

vicilliar: Having a combined generate and deploy step will also be useful for local users

scripts/vespa_local/vespa_local.py
```diff
@@ -587,6 +599,15 @@ def deploy_config(args):
     os.system(f'vespa deploy "{here}"')
 
 
+def generate_application_package(args):
+    # Create instance of VespaLocal
+    # vespa_local_instance is used for starting vespa & generating application package.
+    vespa_local_instance = VespaLocalSingleNode()
```
vicilliar: Instead of hardcoding it to be `VespaLocalSingleNode()`, can we use the `--Shards` and `--Replicas` argument here so it's dynamic as well? Might be confusing if some of the functions are hardcoded to single node and some are not.

scripts/vespa_local/vespa_local.py
```diff
@@ -587,6 +599,15 @@ def deploy_config(args):
...
+    # Generate the application package
+    file_path = vespa_local_instance.generate_application_package(zip=args.zip)
+    print(f"Application package generated at {file_path}")
```
vicilliar: We can have a function `generate_and_deploy_application_package` that 
1. calls `generate_application_package` (with shards and replicas argument)
2. calls `deploy_application_package`
3. cleans up

scripts/vespa_local/vespa_local.py
```diff
@@ -103,24 +107,28 @@ def generate_application_package_files(self):
                     content_for_hosts_xml = self.get_hosts_xml_content()
                     f.write(content_for_hosts_xml)
 
-    def generate_application_package(self) -> str:
+    def generate_application_package(self, zip=True) -> str:
```
vicilliar: Why does `zip` need to be configurable?

vicilliar (approved):

PR #1125, opened 2025-02-27
Improve integ tests execution time

danyilq:
### Enhancements in Integration Test Execution

This PR significantly improves integration test execution speed and simplifies instance management in GitHub Actions.

**Key Improvements**
- **Parallelized Time-Intensive Tests**
  - The 200GB test is now split across 2 instances, and the LargeModel test runs on 4 instances in parallel.
  - This reduces total test execution time from ~1-2 hours to ~33 minutes.
- **Refactored Instance Management**
  - Moved instance creation and teardown into a reusable GitHub Actions workflow.
  - Improves readability, maintainability, and reusability of CI workflows.
- **Enhanced Security**
  - Uses [GitHub OIDC integration](https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services) for better security and access management.
- **Improved Consistency**
  - Added Model retries with jitter for better consistency for model download attempts.
  - Set `HF_HUB_ENABLE_HF_TRANSFER=1` to use higher bandwidth hf models downloader.

These changes optimize test efficiency, reduce execution time, and improve CI/CD maintainability. 🚀


.github/workflows/largemodel_unit_test_CI.yml
```diff
@@ -23,46 +23,79 @@ concurrency:
 
 permissions:
   contents: read
+  id-token: write
 
 jobs:
-  Start-Runner:
-    name: Start self-hosted EC2 runner
+  Start-Runner1:
```
danyilq: Due to matrix output limitations we have to create runners in separate jobs

.github/workflows/unit_test_200gb_CI.yml
```diff
@@ -131,52 +82,58 @@ jobs:
             echo "MULTINODE_TEST_ARGS=--multinode --ignore=tests/integ_tests/core/index_management/test_index_management.py --ignore=tests/integ_tests/core/inference --ignore=tests/integ_tests/processing --ignore=tests/integ_tests/s2_inference" >> $GITHUB_OUTPUT
           fi
 
-  Start-Runner:
+  Start-Runner1:
```
danyilq, line 79: Due to matrix output limitations we have to create runners in separate jobs

.github/workflows/unit_test_200gb_CI.yml
```diff
@@ -256,55 +213,92 @@ jobs:
...
-            tests/integ_tests/ | tee pytest_output.txt
+          COVERAGE_FILE=.coverage.${{ matrix.part }} pytest ${{ env.MULTINODE_TEST_ARGS }} \
+            --ignore=tests/integ_tests/test_documentation.py --durations=100 \
+            --cov=src --cov-branch --cov-context=test --cov-report term:skip-covered \
+            tests/integ_tests/ --split-n 2 --split-part ${{ matrix.part }} --split-by=classes | tee pytest_output.log
+
+      - name: Upload Test Result
+        uses: actions/upload-artifact@v4
+        with:
+          name: marqo-test-report-${{ matrix.part }}
```
danyilq, line 213: Upload separete test results as artifacts to later run `coverage merge`

setup.cfg
```diff
@@ -0,0 +1,2 @@
+[coverage:run]
+relative_files = True
```
danyilq, line 2: necessary so as to mitigate difference between ec2-runner and github-runner's coverage results

tests/integ_tests/conftest.py
```diff
@@ -4,6 +4,10 @@
 def pytest_addoption(parser):
     parser.addoption("--largemodel", action="store_true", default=False)
     parser.addoption("--multinode", action="store_true", default=False, help="Run tests that have multiple Vespa nodes")
+    parser.addoption("--split-n", action="store", type=int, default=1, help="Split tests into N parts")
+    parser.addoption("--split-part", action="store", type=int, default=0, help="Run tests for part N (zero-based index)")
+    # add option to split either by classes or tests
+    parser.addoption("--split-by", action="store", type=str, default="tests", help="Split tests by classes or tests")
```
danyilq: It might be more efficient to split some test suites by tests or classes, therefore option for it.

.github/workflows/arm64_docker_marqo.yml
```diff
@@ -28,40 +28,18 @@ on:
...
-              {"Key": "WorkflowRunId", "Value": "${{ github.run_id }}"},
-              {"Key": "WorlflowURL", "Value": "${{ github.event.repository.html_url }}/actions/runs/${{ github.run_id }}"},
-              {"Key": "PoloRole", "Value": "testing"}
-            ]
+    uses: ./.github/workflows/start_ec2_runner.yml
+    with:
+      base-ami-id: ${{ vars.MARQO_CPU_ARM64_TESTS_INSTANCE_AMI }}
+      ec2-instance-type: r7g.large
+    secrets:
+      GH_PERSONAL_ACCESS_TOKEN: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
```
vicilliar: Maybe we could use `secrets: inherit` to even further avoid duplicated lines

.github/workflows/largemodel_unit_test_CI.yml
```diff
@@ -1,5 +1,5 @@
 name: largemodel_unit_test_CI
-# runs unit tests on CUDA machine with large model tests.
+# Runs unit tests on CUDA machine with large model tests.
```
vicilliar: At this point, since we're refactoring we might as well rename:
```
largemodel_unit_test_CI.yml -> largemodel_integ_test_CI.yml 
unit_test_200gb_CI.yml -> integ_test_CI.yml 
```

And the names can be:
Integration Tests with Shards and Replicas
Large Model Integration Tests

.github/workflows/largemodel_unit_test_CI.yml
```diff
@@ -23,46 +23,79 @@ concurrency:
...
+          INSTANCE_IDS='["${{ needs.Start-Runner1.outputs.ec2-instance-id }}", "${{ needs.Start-Runner2.outputs.ec2-instance-id }}", "${{ needs.Start-Runner3.outputs.ec2-instance-id }}", "${{ needs.Start-Runner4.outputs.ec2-instance-id }}"]'
+          
+          echo "labels=$LABELS" >> $GITHUB_OUTPUT
+          echo "ec2-instance-ids=$INSTANCE_IDS" >> $GITHUB_OUTPUT
 
   Test-Marqo:
-    name: Run Large Model Unit Tests
-    needs: Start-Runner # required to start the main job when the runner is ready
-    runs-on: ${{ needs.start-runner.outputs.label }} # run the job on the newly created runner
+    name: Run Large Model Unit Tests (Part ${{ matrix.part }}/3)
```
vicilliar: Rename to Run Large Model Integration Tests?

.github/workflows/largemodel_unit_test_CI.yml
```diff
@@ -23,46 +23,79 @@ concurrency:
...
+    needs: Create-Matrix # Ensures it starts only after the runners are ready
```
vicilliar: Maybe we can put a timeout on each workflow that generates many runners like this. We don't want a case where we accidentally have 4 g4dn.2xlarge hanging for hours

src/marqo/s2_inference/s2_inference.py
```diff
@@ -2,8 +2,11 @@
 The functions defined here would have endpoints, later on.
```
papa99do: What is the purpose of this change? Does this belong to a separate PR?

.github/workflows/arm64_docker_marqo.yml
```diff
@@ -28,40 +28,18 @@ on:
 
 permissions:
   contents: read
+  id-token: write
```
papa99do, line 31: Why do we need this permission?

.github/workflows/arm64_docker_marqo.yml
```diff
@@ -28,40 +28,18 @@ on:
...
-          aws-resource-tags: > # optional, requires additional permissions
-            [
-              {"Key": "Name", "Value": "marqo-github-runner-${{ github.run_id }}"},
-              {"Key": "GitHubRepo", "Value": "${{ github.repository }}"},
-              {"Key": "WorkflowName", "Value": "${{ github.workflow }}"},
-              {"Key": "WorkflowRunId", "Value": "${{ github.run_id }}"},
-              {"Key": "WorlflowURL", "Value": "${{ github.event.repository.html_url }}/actions/runs/${{ github.run_id }}"},
-              {"Key": "PoloRole", "Value": "testing"}
-            ]
+    uses: ./.github/workflows/start_ec2_runner.yml
```
papa99do, line 35: This refactoring is great, the pipelines are much easier to maintain now.

.github/workflows/unit_test_200gb_CI.yml
```diff
@@ -41,67 +41,20 @@ on:
...
 concurrency:
   group: unit-tests-${{ github.ref }}-${{ inputs.number_of_shards }}-${{ inputs.number_of_replicas }}
   cancel-in-progress: true
 
 permissions:
   contents: read
+  id-token: write
 
 jobs:
-  Check-Changes:
```
papa99do, line 53: We use this instead of `paths-ignore` for a reason. We still need this pipeline to run in order to merge our change.

src/marqo/s2_inference/s2_inference.py
```diff
@@ -519,47 +522,69 @@ def get_model_size(model_name: str, model_properties: dict) -> (int, float):
...
+
+            model.load()  # Load the model
+            print(f"✅ Model `{model_name}` loaded successfully on `{device}`.")
+            return model  # ✅ Success, return the model
+
+        except (ReadTimeoutError, requests.exceptions.Timeout, OSError, RuntimeError) as e:
+            print(f"⚠️ Error loading model `{model_name}` on `{device}`: {e}")
+            attempt += 1
+
+            if attempt >= max_retries:
```
vicilliar: Redundant check if we have `while attempt < max_retries` at the start of loop. Also what if we get an exception not in this list? We will still error out then as normal?

One worry I have also is that this transient error in model loading usually gives us a 400 (user error). How do we differentiate this from an actual user error? If user gave the wrong input, we shouldn't retry at all.

vicilliar: Tagging @wanliAlex to doublecheck my logic here

src/marqo/s2_inference/s2_inference.py
```diff
@@ -519,47 +522,69 @@ def get_model_size(model_name: str, model_properties: dict) -> (int, float):
...
-            device=device,
-            embedding_dim=model_properties['dimensions'],
-            model_properties=model_properties,
-            model_auth=model_auth,
-            max_seq_length=model_properties.get('tokens', get_default_seq_length())
-        )
-    model.load()
-    return model
+    attempt = 0
+    while attempt < max_retries:
```
vicilliar: We should have a unit + integration test for this. Maybe we could put this whole thing in another PR since it's a src code change, and there might be more comments. We don't want it to block the other changes you made

papa99do (commented):
Great job @danyilq , I have some questions in the comments above.

.github/workflows/arm64_docker_marqo.yml
danyilq: To get rid of using pregenerated credentials, OIDC configuration linked in comment requires id-token: write permission

src/marqo/s2_inference/s2_inference.py
danyilq: ahh, if being fair, yes it should have been in different PR. This change simply adds retries for attempts to download model. The reason why it is in this PR is because due to hf hub timeouts flakiness has been too high to test any changes.

tests/integ_tests/conftest.py
```diff
@@ -14,21 +18,57 @@ def pytest_configure(config):
...
+    parts = config.getoption("--split-n")
+    part = config.getoption("--split-part")
+    split_by = config.getoption("--split-by")
+
+    filtered_items = []
+
+    # Step 1: **Pre-filter tests that would be skipped**
+    for item in items:
+        # Skip tests that are cpu_only if --largemodel is set
+        if config.getoption("--largemodel") and (
```
vicilliar, line 40: We should probably separate the `"skip" in item.keywords` check, since it will skip these tests regardless of the flag

tests/integ_tests/conftest.py
```diff
@@ -14,21 +18,57 @@ def pytest_configure(config):
...
+            continue  # Skip adding this test to filtered_items
+
+        filtered_items.append(item)
+
+    # Step 1.1: **Split by classes if split-by==classes**
+    if split_by == "classes":
+        class_to_tests = {}
+        for item in filtered_items:
+            class_name = item.parent.name
+            if class_name not in class_to_tests:
```
vicilliar: nit: should this be `classes_to_test` ?

.github/workflows/unit_test_200gb_CI.yml
danyilq: I'm slightly confused, why do we need it to run? Isn't it technically the same?
with paths-ignore: just never run it
with check job: check for changes and still never run anything else

Isn't it the same for us?

vicilliar (changes requested):

tests/integ_tests/conftest.py
```diff
@@ -14,21 +18,57 @@ def pytest_configure(config):
...
+        chunk_size = max(1, len(filtered_items) // parts)
+    print(f"Total tests: {len(filtered_items)}, Splitting into {parts} parts, each with {chunk_size} {split_by}")
+
+    # Step 3: **Determine the range of tests/classes to keep in this partition**
+    start_idx = part * chunk_size
+    end_idx = start_idx + chunk_size
+
+    if part + 1 == parts:
+        # Include all remaining classes/tests in the last partition
+        end_idx = len(filtered_items) if not split_by == "classes" else len(sorted_classes)
```
vicilliar: A bit confusing to keep evaluating `split_by == "classes"` inline. Maybe just put all logic in blocks:
```
if split_by == "classes":
    # Do all logic for class split
elif split_by == "another_option":
    # Do something else
else:
    # Default logic
```

src/marqo/s2_inference/s2_inference.py
danyilq: Yes, any other scenario it will just raise the same error without retrying. This check is not redundant, it is there to capture error and propagate correct error when we run out of retries. As to while loop, I'd prefer check there being redundant rather than having loop that could end up running indefinitely 🙂. I'm not sure about what error user scenario would end up with but I'd assume something like NotFound? which shouldn't be retried here

tests/integ_tests/conftest.py
danyilq: the reason it is here is to make sure we split tests efficiently. i.e. let's say scenario where there is a total of 100 tests and 4 runners, tests from 25-50 could be skippable and runner would be spun up for no reason. Getting rid of those tests here would mean that each runner is actually spun up for a reason

src/marqo/s2_inference/s2_inference.py
danyilq: ahh, I agree

src/marqo/s2_inference/s2_inference.py
papa99do: Could this be caused by downloading hf models concurrently in these parallel test cases, leading to us being throttled?

.github/workflows/unit_test_200gb_CI.yml
papa99do: with paths-ignore: just never run it
This is the problem we try to solve. This pipeline is required to pass to merge a PR, if it's never run, we will not be able to merge it.

src/marqo/s2_inference/s2_inference.py
danyilq: it shouldn't be different to us running several tests suites on several PRs sequentially, besides Li's PRs were hitting the same issue. It might have been some incident on hf side. It was a ReadTimeoutError too, not a 429.

.github/workflows/unit_test_200gb_CI.yml
danyilq: It sounds hacky. Why do we need to enforce this pipeline run for a merge?
If we don't have any documentation changes - it runs regardless and is required.
If we do - we can just not run it?

tests/integ_tests/conftest.py
vicilliar: Yep I agree, I meant in this skipping logic, we could just have the if statement:
```
if "skip" in item.keywords:
    continue
```

at the start, so we don't have to check for it in the other if statements right?

vicilliar (changes requested):

tests/integ_tests/conftest.py
danyilq: ah, well I wanted our non-keyword tests to still skip them so we can keep track of how many we skipped and why in the future. If you think it's not needed we can do it this way.

vicilliar (approved):

PR #1132, opened 2025-03-03
Fixing the score modifier not being updated bug. 

adityabharadwaj198:
* bug fix

* in cases where the partial update request is not changing the entire map, score modifiers were not getting updated. So for example 
* ```mq.index('my-index').add_documents(
    documents=[
        {
            '_id': '1',
            'text': 'This is a test document',
            'rank': 100,
            'metadata': {
                'key1': 1.5,
            }
        }
    ],
    tensor_fields=['text'],)

```mq.index('my-index').update_documents(
        [{'_id': '1', 'metadata': {'key1':300.2}}]
    )
```

Score modifiers were not getting updated for this scenario. 

This is fixed now. 

* No

* Manual testing has been done 

* **Other information**:

### DIFF
--- a/src/marqo/core/document/document.py
+++ b/src/marqo/core/document/document.py
@@ -20,6 +20,7 @@ from marqo.core.structured_vespa_index.structured_add_document_handler import St
 from marqo.core.unstructured_vespa_index.unstructured_add_document_handler import UnstructuredAddDocumentsHandler
 from marqo.core.vespa_index.vespa_index import for_marqo_index as vespa_index_factory
 from marqo.logging import get_logger
+from marqo.marqo_docs import update_documents_response
 from marqo.tensor_search.telemetry import RequestMetricsStore
 from marqo.vespa.models import UpdateDocumentsBatchResponse, VespaDocument
 from marqo.vespa.models.delete_document_response import DeleteAllDocumentsResponse
@@ -128,10 +129,14 @@ class Document:
         unsuccessful_docs: List[Tuple[int, MarqoUpdateDocumentsItem]] = []
 
         # Remove duplicated documents based on _id
-        partial_documents, doc_ids, documents_that_contain_maps = self.process_documents(partial_documents,
+        partial_documents, documents_that_contain_maps = self.process_documents(partial_documents,
                                                                                          unsuccessful_docs, is_index_semi_structured=marqo_index.type is IndexType.SemiStructured)
+        documents_that_contain_maps_but_dont_exist_in_vespa = set() # Set to keep track of documents that contain maps but don't exist in Vespa. This will remain unpopulated for structured indexes.
         existing_vespa_documents = {}
 
+        # This block will only execute for SemiStructured Indexes and in case the request has documents that contain maps.
+        # 1. Retrieve those documents, which contain maps in the update request, from Vespa.
+        # 2. If there's any documents that dont' exist in Vespa, we will append them to unsuccessful_docs
         if marqo_index.type is IndexType.SemiStructured and documents_that_contain_maps: # Only retrieve the document back if the partial update request contains maps and the index is semi-structured
             get_batch_response = self.vespa_client.get_batch(ids = list(documents_that_contain_maps), fields = [
                 VESPA_FIELD_ID, INT_FIELDS, FLOAT_FIELDS, VESPA_DOC_FIELD_TYPES, VESPA_DOC_CREATE_TIMESTAMP], schema = marqo_index.schema_name)
@@ -139,8 +144,18 @@ class Document:
             for resp in responses:
                 if resp.document:
                     existing_vespa_documents[resp.document.fields[VESPA_FIELD_ID]] = resp.document.dict()
+                else: 
+                    id = self.extract_document_id_from_vespa_id(resp) # Extract the document id from the Vespa response. Vespa response will contain the document id even though the document was not found.
+                    unsuccessful_docs.append((documents_that_contain_maps.get(id), MarqoAddDocumentsItem(id = id, 
+                                                                                                         status = int(api_exceptions.BadRequestError.status_code),
+                                                                                                         error = "Marqo vector store couldn't update the document. Please see: " + update_documents_response() + " for more details")))
+                    documents_that_contain_maps_but_dont_exist_in_vespa.add(id)
 
         for index, doc in enumerate(partial_documents):
+            if (marqo_index.type is IndexType.SemiStructured # Only have this check in place for SemiStructured indexes
+                    and documents_that_contain_maps_but_dont_exist_in_vespa
+                    and doc.get(MARQO_DOC_ID, '') in documents_that_contain_maps_but_dont_exist_in_vespa):
+                continue
             try:
                 vespa_document = VespaDocument(**vespa_index.to_vespa_partial_document(doc, existing_vespa_documents.get(doc.get(MARQO_DOC_ID, ''), None)))
                 vespa_documents.append(vespa_document)
@@ -182,7 +197,7 @@ class Document:
 
         if responses is not None:
             for resp in responses.responses:
-                doc_id = resp.id.split('::')[-1] if resp.id else None
+                doc_id = self.extract_document_id_from_vespa_id(resp)
                 status, message = self.vespa_client.translate_vespa_document_response(resp.status, None)
                 new_item = MarqoUpdateDocumentsItem(id=doc_id, status=status, message=message, error=message)
                 items.append(new_item)
@@ -195,7 +210,7 @@ class Document:
                                             processingTimeMs=(timer() - start_time) * 1000)
 
     def process_documents(self, documents: List[Dict], unsuccessful_docs: List[Tuple[int, MarqoUpdateDocumentsItem]],
-                          is_index_semi_structured = False) -> Tuple[List, set, set]:
+                          is_index_semi_structured = False) -> Tuple[List[Dict], Dict]:
         """Process documents to remove duplicates and identify documents containing maps.
         
         This method combines duplicate removal and map detection into a single pass through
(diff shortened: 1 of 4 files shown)


src/marqo/core/document/document.py
```diff
@@ -137,7 +137,8 @@ def partial_update_documents(self, partial_documents: List[Dict], marqo_index) \
                 VESPA_FIELD_ID, INT_FIELDS, FLOAT_FIELDS, VESPA_DOC_FIELD_TYPES, VESPA_DOC_CREATE_TIMESTAMP], schema = marqo_index.schema_name)
             responses = get_batch_response.responses
             for resp in responses:
-                existing_vespa_documents[resp.document.fields[VESPA_FIELD_ID]] = resp.document.dict()
+                if resp.document:
```
farshidz, line 145: Why did we add this line? Also note these checks tend to be risky as it evals to false if `resp.document` is None or empty list/dict. Can we do a more specific condition if it's required?

tests/api_tests/v1/tests/api_tests/unstructured_index/test_update_documents_unstructured_index.py
```diff
@@ -154,4 +154,78 @@ def test_update_document_with_ids_change_field_type(self):
 
         self.assertEqual(update_docs_response['items'][0]['status'], 400)
         self.assertIn("Marqo vector store couldn't update the document. Please see", update_docs_response['items'][0]['message'])
-        self.assertIn("reference/api/documents/update-documents/#response", update_docs_response['items'][0]['message'])
\ No newline at end of file
+        self.assertIn("reference/api/documents/update-documents/#response", update_docs_response['items'][0]['message'])
+
+    def test_update_document_with_changes_in_score_modifiers(self):
+        # Test updating a document with new fields and updating existing fields
```
farshidz: Make this line a docstring so it shows up in test runs

tests/api_tests/v1/tests/api_tests/unstructured_index/test_update_documents_unstructured_index.py
```diff
@@ -154,4 +154,78 @@ def test_update_document_with_ids_change_field_type(self):
...
+        update_docs_response = self.client.index(self.text_index_name).update_documents(
+            [{
+                '_id': '1',
```
farshidz, line 208: Where is the doc first added?

farshidz (changes requested):
Integ tests were also missing this issue. Update integ tests as well

src/marqo/core/document/document.py
papa99do, line 145: why do we need this check?

adityabharadwaj198: I've added it because when debugging I noticed that there was a case when resp.document was None so resp.document.dict() throwing the error `[error 'NoneType' object has no attribute '__dict__']`

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
```diff
@@ -326,17 +326,22 @@ def _update_score_modifiers(self, original_doc: Optional[SemiStructuredVespaDocu
             ]
 
 
-        if len(score_modifier_to_be_removed) > 0 or len(score_modifier_to_be_removed) > 0:
```
papa99do, line 329: I missed this in the review, and I thought the integration test covers this. However, it was not covered since in the test case, we always remove a map key. That said, I think it makes sense to change the integration test to split the test case to multiple sub test cases or multiple method, which covers following cases
1. only add and change fields (int, float, and map) 
2. only remove keys from maps
3. both 1 and 2
So the test needs to be defensive. So if we revert this fix, some test case should fail. Your thoughts? @adityabharadwaj198

src/marqo/core/document/document.py
adityabharadwaj198: I've added it because when debugging I noticed that there was a case when resp.document was None so resp.document.dict() throwing the error `[error 'NoneType' object has no attribute '__dict__']`

tests/api_tests/v1/tests/api_tests/unstructured_index/test_update_documents_unstructured_index.py
adityabharadwaj198: moved the document adding logic to setupClass() method as I realised that previously wasn't the case

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
adityabharadwaj198: Had a call offline for this and we decided that this is required for maps and score modifiers specifically. (we also discussed string arrays but later I realised that you can't add new string arrays and only replace existing string arrays which is already covered). I have added a test for maps & score modifiers where we only add a new map (thereby adding a new score modifier as well) - this check is done in two distinct test cases though. I also have a test case for changing fields in a map which by default takes cares of the `only remove keys from maps`

tests/integ_tests/core/document/test_partial_update_semi_structured.py
```diff
@@ -388,6 +401,67 @@ def test_partial_update_should_update_score_modifiers(self):
...
+        res = self.config.vespa_client.get_document('2',
+                                                    self.config.index_management.get_index(self.index.name).schema_name)
+        doc = res.document.dict().get('fields')
+        # Verify original fields are preserved
+        self.assertEqual(doc['marqo__score_modifiers']['cells']['int_field'], 123.0)
+        self.assertEqual(doc['marqo__score_modifiers']['cells']['float_field'], 123.0)
+        self.assertEqual(doc['marqo__score_modifiers']['cells']['int_map.a'], 3.0)
+        self.assertEqual(doc['marqo__score_modifiers']['cells']['int_map.b'], 4.0)
+        self.assertEqual(doc['marqo__score_modifiers']['cells']['float_map.c'], 3.0)
+        self.assertEqual(doc['marqo__score_modifiers']['cells']['float_map.d'], 4.0)
```
papa99do, line 465: Lacking a test case for removing map keys from score modifiers? and probably a combined one (1 add, 1 replace and 1 removal)?

src/marqo/core/document/document.py
papa99do: Let's track in this thread. Do we know why it returns None? should we raise an error in this case instead?

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
papa99do: what is the name of that test case that `only remove keys from maps`?

adityabharadwaj198: We don't need a seperate test case for that as any test which does key updation (i.e either completely replaces the map or even a few keys inside the map) will remove keys from the map. 
Only two different scenarios are 
1. exclusively adding new maps => adding new keys inside the score modifier tensor
2. replacing entire map / replacing few keys in map => adding new keys inside the tensor + removing keys from the tensor

tests/integ_tests/core/document/test_partial_update_semi_structured.py
adityabharadwaj198: We don't need a seperate test case for just removal. Any test case where entire map is being replaced or even a few keys are being replaced will lead to removal from score modifier tensor

src/marqo/core/document/document.py
adityabharadwaj198: It returns none when a user sends a update request for a doc that doesn't exist but the update request contains a map. 
Since the request contains a map we will try to get docs which contains maps but since the doc doesn't exist in the index resp.document will be none. 

I think we can raise an error in this case (similar to how we handle other errors i.e by adding to unsuccessful_docs & sending it in response).

papa99do: yes, if a user tries to update an non-existing doc, I think we should return 404 early, instead of updating Vespa and expect a 412.

tests/integ_tests/core/document/test_partial_update_semi_structured.py
papa99do: but do we still have any assertion for removing a key from the score modifier?

src/marqo/core/document/document.py
adityabharadwaj198: to maintain consistency we will throw a 400

tests/integ_tests/core/document/test_partial_update_semi_structured.py
papa99do: OK, I think the original case covers this. I thought we deleted that one. All good.

farshidz (dismissed):

src/marqo/core/document/document.py
```diff
@@ -128,18 +129,33 @@ def partial_update_documents(self, partial_documents: List[Dict], marqo_index) \
...
             get_batch_response = self.vespa_client.get_batch(ids = list(documents_that_contain_maps), fields = [
                 VESPA_FIELD_ID, INT_FIELDS, FLOAT_FIELDS, VESPA_DOC_FIELD_TYPES, VESPA_DOC_CREATE_TIMESTAMP], schema = marqo_index.schema_name)
             responses = get_batch_response.responses
             for resp in responses:
-                existing_vespa_documents[resp.document.fields[VESPA_FIELD_ID]] = resp.document.dict()
+                if resp.document:
+                    existing_vespa_documents[resp.document.fields[VESPA_FIELD_ID]] = resp.document.dict()
+                else: 
+                    id = self.extract_document_id_from_vespa_id(resp) # Extract the document id from the Vespa response. Vespa response will contain the document id even though the document was not found.
+                    unsuccessful_docs.append((documents_that_contain_maps.get(id), MarqoAddDocumentsItem(id = id,
```
papa99do: This should be `MarqoUpdateDocumentsItem`

adityabharadwaj198: fixed

papa99do (approved):

farshidz (approved):

PR #1138, opened 2025-03-06
Implement rerankDepth feature

danyilq:
Implements rerankDepth feature. It will be only available for recommend, tensor and hybrid search with minimal always equal to `limit + offset`. It also adds first unit tests for search, with base class that shall be reused.

`rerankDepthTensor` for hybrid. existing `rerankDepth` will serve two purposes now: rerank top k results for hybrid, collect k hits for tensor (vespa's targetHits).

Currently `rerankDepth` is not exposed to the user, which results in queries returning inconsistent number of results (edge case behaviour).

Users will be able to set `rerankDepth` or `rerankDepthTensor` when creating a query, which will allow them to get consistent number of results in some edge cases. This is solution recommended by Vespa team.

https://github.com/marqo-ai/py-marqo/pull/278

https://github.com/marqo-ai/marqodocs/pull/418

* **Other information**:


tests/unit_tests/marqo/base_test_case.py
```diff
@@ -0,0 +1,118 @@
...
+        cls.model = Model(name="hf/all_datasets_v4_MiniLM-L6")
+
+        # Structured index with multimodal fields
+        cls.structured_index = StructuredMarqoIndex(
+            name="index_name", schema_name="test_schema", type=IndexType.Structured, model=cls.model,
+            normalize_embeddings=True,
+            text_preprocessing=TextPreProcessing(split_length=5, split_overlap=2, split_method="word"),
+            image_preprocessing=ImagePreProcessing(patch_method=None), distance_metric=DistanceMetric.Euclidean,
+            vector_numeric_type=VectorNumericType.Float, hnsw_config=HnswConfig(ef_construction=200, m=16),
+            marqo_version="2.16.0", created_at=1234567890, updated_at=1234567890, fields=[Field(
```
adityabharadwaj198: This should be picked up from the actual marqo version right?

src/marqo/core/models/marqo_query.py
```diff
@@ -23,6 +23,7 @@ class Config(StrictBaseModel.Config):
     filter: Optional[SearchFilter] = None
     score_modifiers: Optional[List[ScoreModifier]] = None
     expose_facets: bool = False
+    target_hits: Optional[int] = None
```
adityabharadwaj198: can you add a doc to highlight what this parameter does?

tests/unit_tests/marqo/core/search/test_search.py
```diff
@@ -0,0 +1,55 @@
...
+        self.assertEqual(call_args['yql'], self.get_expected_tensor_yql())
+        self.assertEqual(call_args['model_restrict'], 'test_schema')
+        self.assertEqual(call_args['hits'], 3)
+        self.assertEqual(call_args['offset'], 0)
+
+    def test_tensor_search_with_target_hits(self):
+        tensor_search.search(self.config, "index_name", "query", search_method="tensor", target_hits=5)
+        self.vespa_client_mock.query.assert_called_once()
+        call_args = self.vespa_client_mock.query.call_args[1]
+        self.assertEqual(call_args['yql'], self.get_expected_tensor_yql(target_hits=5))
```
adityabharadwaj198: should we   compare call_args['hits'] and call_args['offset'] here as well?

src/marqo/core/search/hybrid_search.py
```diff
@@ -61,6 +62,7 @@ def search(
                 text_query_prefix: prefix for text queries (for vectorisation only)
                 hybrid_parameters: HybridParameters object to specify all parameters for hybrid search. If not provided,
                     default values will be used.
+                target_hits: target number of hits to return. If None, the default value will be used.
```
adityabharadwaj198: nit: Can we mention what default value it is?

tests/unit_tests/marqo/core/search/test_search.py
```diff
@@ -0,0 +1,55 @@
...
+            self.get_expected_tensor_yql()
+        )
+        self.assertEqual(
+            call_args['marqo__yql.lexical'], self.get_expected_lexical_yql("query")
+        )
+
+    def test_hybrid_search_with_target_hits(self):
+        tensor_search.search(self.config, "index_name", "query", search_method="hybrid", target_hits=5)
+        self.vespa_client_mock.query.assert_called_once()
+        call_args = self.vespa_client_mock.query.call_args[1]
```
adityabharadwaj198, line 203: ^

adityabharadwaj198 (changes requested):
Can you also update the PR description with the current behavior & the new behavior? Will help me understand the feature better

tests/unit_tests/marqo/base_test_case.py
danyilq: it is a unit test, it doesn't use marqo directly, it'd be more fair to replace it with empty string so it is less confusing

tests/unit_tests/marqo/core/search/test_search.py
danyilq: it is a unit test so those values are not actually real, as vespa client is mocked. What would be the point of testing those values here?

src/marqo/core/models/marqo_query.py
danyilq: what do you mean by that? where do you expect that doc?

tests/unit_tests/marqo/core/search/test_search.py
adityabharadwaj198: sorry, I don't get why we're testing them here then: https://github.com/marqo-ai/marqo/pull/1138/files#diff-cc3043acbc9be7314b2862b3dd8a4f2db244f4e0f7ba4c8d82f60658d4b703c9R12

src/marqo/core/models/marqo_query.py
adityabharadwaj198: I mean a short description on what this parameter is used for. To the right or the top of the actual parameter definition.

tests/unit_tests/marqo/core/search/test_search.py
danyilq: unfortunately `targetHits` do not affect any of those values returned, we can't actually integ test it. @papa99do was not able to reproduce the reason we are exposing that parameter. It is a very unique edge case that might be a solution for some large index query results inconsistencies. Therefore we can only really unit test that we are actually passing this parameter.

src/marqo/core/search/hybrid_search.py
```diff
@@ -37,7 +37,8 @@ def search(
             media_download_headers: Optional[Dict] = None, context: Optional[SearchContext] = None,
             score_modifiers: Optional[ScoreModifierLists] = None, model_auth: Optional[ModelAuth] = None,
             highlights: bool = False, text_query_prefix: Optional[str] = None,
-            hybrid_parameters: HybridParameters = None) -> Dict:
+            hybrid_parameters: HybridParameters = None, target_hits: Optional[int] = None
```
farshidz: This may not be the best name for this parameter. Have you seen this document? https://www.notion.so/Marqo-reranking-configuration-18175d43da4c809492adf29e0273e8a5

target hits is rerank depth in certain situations

danyilq: Thanks for sharing this document, if I understand it correctly `targetHits == rerankCount`. Are you saying we should rename it to `rerankCount` instead @farshidz?

src/marqo/core/models/marqo_query.py
```diff
@@ -44,6 +44,7 @@ class MarqoTensorQuery(MarqoQuery):
     vector_query: List[float]
     ef_search: Optional[int] = None
     approximate: bool = True
+    rerank_depth: Optional[int] = None
```
farshidz: There's an existing `rerank_depth` in the parent class `MarqoQuery`. So technically you don't need this here. But this is all a bit confusing because we're attaching different semantics to the same field (can mean global depth for one query, and tensor depth for another). Plus `rerank_depth` in parent `MarqoQuery` is already questionable because not all queries support it (e.g. lexical).

I wonder if internally we should use more descriptive names for these in the code (use aliases if needed so the API is what we agreed). One idea is to have `global_rerank_depth` in `MarqoHybridQuery`, `rerank_depth_tensor` is `MarqoTensorQuery` (`MarqoHybridQuery` will need validation to disallow this for hybrid) and `rerank_depth_tensor` within `HybridParameters` for hybrid.

src/marqo/core/models/marqo_query.py
```diff
@@ -62,6 +63,7 @@ class MarqoHybridQuery(MarqoTensorQuery, MarqoLexicalQuery):
     # Core module will use these fields instead of the score_modifiers_lexical and score_modifiers_tensor inside the HybridParameters
     score_modifiers_lexical: Optional[List[ScoreModifier]] = None
     score_modifiers_tensor: Optional[List[ScoreModifier]] = None
+    rerank_depth_tensor: Optional[int] = None
```
farshidz: This should be under `hybridParameters`

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -490,7 +493,11 @@ def _to_vespa_hybrid_query(self, marqo_query: MarqoHybridQuery) -> Dict[str, Any
         fields_to_search_tensor = self._get_tensor_fields_to_search(
             searchable_attributes=marqo_query.hybrid_parameters.searchableAttributesTensor
         )
-        tensor_term = self._get_tensor_search_term(marqo_query) if fields_to_search_tensor else "False"
+        tensor_term = "False"
+        if fields_to_search_tensor:
+            tensor_marqo_query = copy.deepcopy(marqo_query)
+            tensor_marqo_query.rerank_depth = marqo_query.hybrid_parameters.rerankDepthTensor
```
farshidz: This seems a bit weird. You're moving the value within the query object because `_get_tensor_search_term` is looking somewhere else? The other refactoring for the pydantic models may help with this code

src/marqo/tensor_search/helper.py
```diff
@@ -0,0 +1,18 @@
+from marqo.core.models import MarqoTensorQuery
+
+
+def get_rerank_depth_and_additional_hits_from_query(query: MarqoTensorQuery):
```
farshidz: Let's have this in the class that needs it, or a more specific new class if we really have to. Helper-style utility classes have in the past been a source of bad coupling and poor code organization, we've eliminated a lot of them and don't want to add new ones. Over time they make us lazy as anything is a *helper* technically so things keep getting dumped here.

I suspect there will be an existing good place for this.

tests/unit_tests/marqo/base_test_case.py
```diff
@@ -0,0 +1,119 @@
...
+
+from marqo.core.models.marqo_index import (
+    StructuredMarqoIndex, Model, TextPreProcessing, ImagePreProcessing,
+    DistanceMetric, VectorNumericType, HnswConfig, FieldType, FieldFeature, IndexType, Field, TensorField
+)
+from marqo.config import Config
+from marqo.tensor_search.telemetry import RequestMetricsStore
+from marqo.version import get_version
+
+class BaseUnitTest(unittest.TestCase):
```
farshidz: I'm not a fan of this base class. It will be unnecessary for most unit test classes. I suggest mocking what you need in your test class. Unit tests should be very light and generally self-contained.

tests/unit_tests/marqo/base_test_case.py
```diff
@@ -0,0 +1,119 @@
...
+
+    @classmethod
+    def tearDownClass(cls):
+        # Stop the patchers
+        cls.get_index_patcher.stop()
+        cls.logger_patcher.stop()
+        cls.metrics_store_patcher.stop()
+
+    def get_expected_tensor_yql(self, rerank_depth=3):
+        yql = f"select * from {self.current_index.schema_name} where ("
```
farshidz: These don't seem reusable across the board. Again, the whole base class is dubious IMO.

farshidz (changes requested):
Integ tests and API tests missing

tests/unit_tests/marqo/base_test_case.py
danyilq: imo it is something that could be expanded as we add more tests, currently it is bare minimum introduced specifically for my tests

src/marqo/core/models/hybrid_parameters.py
```diff
@@ -33,6 +33,8 @@ class HybridParameters(StrictBaseModel):
     scoreModifiersLexical: Optional[ScoreModifierLists] = None
     scoreModifiersTensor: Optional[ScoreModifierLists] = None
 
+    rerankDepthTensor: Optional[int] = None
```
farshidz, line 36: Not directly related to this PR, but I'm confused by the arbitrary switching to/from camelCase in these models. I guess the only way around it would be to have duplicate API models that are camel case, which is extra code. Perhaps pydantic aliases could help?

src/marqo/core/models/marqo_query.py
```diff
@@ -62,6 +62,7 @@ class MarqoHybridQuery(MarqoTensorQuery, MarqoLexicalQuery):
     # Core module will use these fields instead of the score_modifiers_lexical and score_modifiers_tensor inside the HybridParameters
     score_modifiers_lexical: Optional[List[ScoreModifier]] = None
     score_modifiers_tensor: Optional[List[ScoreModifier]] = None
+    global_rerank_depth: Optional[int] = None
```
farshidz, line 66: Because you're inheriting `MarqoTensorQuery`, you need variation to error out if `rerank_depth_tensor` is populated

farshidz: Then again I guess that validation would break `structured_vrspa_index.py` so maybe we can leave it as is

src/marqo/core/vespa_index/vespa_index.py
```diff
@@ -211,6 +213,17 @@ def _get_hybrid_score_modifiers(self, hybrid_query: MarqoHybridQuery) -> \
 
         return result
 
+    @staticmethod
```
farshidz: I'm generally of the opinion that static methods in a case like this (base class) add little value and will force extra refactoring if later on the function evolves to access instance variables (assuming they're called as static methods somewhere).

src/marqo/core/vespa_index/vespa_index.py
```diff
@@ -211,6 +213,17 @@ def _get_hybrid_score_modifiers(self, hybrid_query: MarqoHybridQuery) -> \
...
+    def _get_rerank_depth_and_additional_hits_from_query(query: Union[MarqoTensorQuery, MarqoHybridQuery]) -> Tuple[int, int]:
+        if query.ef_search is not None:
+            base_rerank_depth = min(query.limit + query.offset, query.ef_search)
+            additional_hits = max(query.ef_search - (query.limit + query.offset), 0)
+        else:
+            base_rerank_depth = query.limit + query.offset
+            additional_hits = 0
+
+        return query.rerank_depth_tensor if query.rerank_depth_tensor else base_rerank_depth, additional_hits
```
farshidz: I don't think this makes sense. If I provide efSearch=1000 and rerankDpeth=2000, we're doing additional_hits=1000, targethits=2000 which is efSearch=3000. efSearch should be the maximum of both, in this case 2000. So targetHits=2000 and no additional hits. Similar to the approach we've taken with pagination before.

But I'm also realizing now that users could break pagination with this and it's not clear to them why. If you have offset=1000, limit=100, rerankDepthTensor=200, you're not going to get 1100 results back (maybe with a lot of content nodes, but generally you won't). Might need to check with Jesse on this one

tests/api_tests/v1/tests/api_tests/unstructured_index/test_search.py
```diff
@@ -453,3 +453,26 @@ def test_searchable_attributes(self):
                                                                       searchable_attributes=["title"])
         self.assertEqual(len(search_res["hits"]), 1)
         self.assertEqual(search_res["hits"][0]["_id"], "1")
+
+    def test_rerank_depth(self):
```
farshidz: Looking at the implementation I'm guessing you're unintentionally enabling this for structured as well so would be good to test that too. We have `test_search_common.py` for tests like this where everything except index type is the same. You could move it there + subtests

tests/integ_tests/core/search/test_recommender.py
```diff
@@ -647,3 +647,38 @@ def test_recommend_searchCallValid_success(self):
...
+        for index in [self.unstructured_text_index, self.structured_text_index]:
+            with self.subTest(type=index.type):
+                self.add_documents(
+                    self.config, add_docs_params=AddDocsParams(
+                        index_name=index.name, docs=docs, tensor_fields=["title", "content"] if isinstance(index, UnstructuredMarqoIndex) else None
+                    )
+                )
+
+                res = self.recommender.recommend(
+                    documents=['1', '2'], result_count=10,
```
farshidz: Can you add more cases to cover edge cases and invalid values (e.g., negative), as well as how this interacts with efSearch, limit, offset?

tests/integ_tests/tensor_search/integ_tests/test_hybrid_search.py
```diff
@@ -2716,4 +2716,62 @@ def test_correct_error_is_raised_if_rerank_depth_is_provided_on_old_marqo_versio
             search_method="HYBRID",
             # rerank_depth=3, # This should not raise an error
             result_count=3
-        )
\ No newline at end of file
+        )
+
+    def test_rerank_depth_tensor_hybrid_search(self):
```
farshidz, line 2721: Can you add more cases to cover edge cases and invalid values (e.g., negative), as well as how this interacts with efSearch, limit, offset?

tests/api_tests/v1/tests/api_tests/unstructured_index/test_search.py
```diff
@@ -453,3 +453,26 @@ def test_searchable_attributes(self):
...
+            docs.append({
+                "title": f"Doc {i}",
+                "content": "some extra info",
+                "_id": str(i)
+            })
+
+        self.client.index(self.text_index_name).add_documents(docs, tensor_fields=["title", "content"])
+
+        # search for "Doc" with rerank_depth=5
+        search_res = self.client.index(self.text_index_name).search(q="Doc", rerank_depth=5, limit=10, search_method="TENSOR")
```
farshidz: Hybrid API test missing

farshidz (changes requested):

src/marqo/core/vespa_index/vespa_index.py
danyilq: but it is not something new. It is a behaviour we had previously, and I just moved it to a new place, accounting for possibly provided targetHits.

tests/api_tests/v1/tests/api_tests/unstructured_index/test_search.py
danyilq: thanks, I didn't realise that file existed

tests/api_tests/v1/tests/api_tests/test_search_common.py
```diff
@@ -165,4 +165,72 @@ def test_proper_error_if_both_imageDownloadHeaders_and_mediaDownloadHeaders_are_
...
+                    "content": "some extra info",
+                    "_id": str(i)
+                })
+
+            self.client.index(index_name).add_documents(docs, tensor_fields=["title", "content"] if index_name == self.unstructured_image_index_name else None)
+
+            # search for "Doc" with rerank_depth=5
+            search_res = self.client.index(index_name).search(q="Doc", rerank_depth=5, limit=10, search_method="TENSOR")
+            # due to additional hits sometimes the result may not be 5
+            self.assertNotEqual(len(search_res["hits"]), 10)
```
farshidz: These are separate test cases. One problem with putting too many things under one test is if the first assert fails, the second one is never evaluated. Sometimes we get around this with subtests, so you could also do that

tests/api_tests/v1/tests/api_tests/test_search_common.py
```diff
@@ -165,4 +165,72 @@ def test_proper_error_if_both_imageDownloadHeaders_and_mediaDownloadHeaders_are_
                     media_download_headers={"marqo_media_header": "media_header_test_key"}
                 )
                 self.assertIn("Cannot set both imageDownloadHeaders and mediaDownloadHeaders.",
-                              str(cm.exception.message))
\ No newline at end of file
+                              str(cm.exception.message))
+
+    def test_rerank_depth(self):
+        for index_name in [self.unstructured_image_index_name, self.structured_image_index_name]:
```
farshidz, line 172: When you loop through indexes, you want to use subtests so all cases run even if one fails

tests/api_tests/v1/tests/api_tests/test_search_common.py
```diff
@@ -165,4 +165,72 @@ def test_proper_error_if_both_imageDownloadHeaders_and_mediaDownloadHeaders_are_
...
+            search_res = self.client.index(index_name).search(q="Doc", rerank_depth=5, limit=10, search_method="TENSOR")
+            # due to additional hits sometimes the result may not be 5
+            self.assertNotEqual(len(search_res["hits"]), 10)
+
+            # search for "Doc" without rerank_depth
+            search_res = self.client.index(index_name).search(q="Doc", limit=10, search_method="TENSOR")
+            self.assertEqual(len(search_res["hits"]), 10)
+
+    def test_rerank_depth_hybrid(self):
+        for index_name in [self.unstructured_image_index_name, self.structured_image_index_name]:
```
farshidz, line 208: Same issue with combining cases without subtests

tests/api_tests/v1/tests/api_tests/test_search_common.py
```diff
@@ -165,4 +165,72 @@ def test_proper_error_if_both_imageDownloadHeaders_and_mediaDownloadHeaders_are_
...
+            # search for "Doc" with rerank_depth=5
+            search_res = self.client.index(index_name).search(
+                q="Doc", limit=10, search_method="HYBRID", hybrid_parameters={
+                    "rankingMethod": "tensor",
+                    "retrievalMethod": "tensor",
+                    "rerankDepthTensor": 5
+                }
+            )
+            # due to additional hits sometimes the result may not be 5
+            self.assertNotEqual(len(search_res["hits"]), 10)
```
farshidz: You're just testing that it works, not that the results are correct. I suggest you look at https://github.com/marqo-ai/marqo/blob/6b3a028dc34a1e82e912b39a0f0cf987ba17835b/tests/api_tests/v1/tests/api_tests/test_score_modifiers_search.py#L137 and other global score modifier tests

farshidz (changes requested):

src/marqo/core/vespa_index/vespa_index.py
farshidz: What I propose is 

```
if efSearch is not None:
	targetHits=min(max(limit+offset, rerankDepth), efSearch)
	additionalHits=min(efSearch-targetHits, 0)
else:
	targetHits=max(limit+offset, rerankDepth)
	additionalHits = 0
```

farshidz (dismissed):

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -663,15 +665,21 @@ def _get_individual_field_tensor_search_terms(self, marqo_query: MarqoTensorQuer
...
         else:
             rerank_depth = marqo_query.limit + marqo_query.offset
-        additional_hits = 0
 
         if marqo_query.ef_search is not None:
             rerank_depth = min(rerank_depth, marqo_query.ef_search)
-            additional_hits = max(marqo_query.ef_search - rerank_depth, 0)
+        else:
+            # efSearch must be min result_count + offset
+            marqo_query.ef_search = max(
```
farshidz, line 685: nit I find it confusing that you're mutating efsearch, especially because in the line above ef_search overrides rerank depth, but here it doesn't. So I would've put this value into a new variable

farshidz (dismissed):

farshidz (approved):

PR #1139, opened 2025-03-06
Fixing the bug where map was not getting updated if the map wasn't present when the document was added using the add_documents call. 

adityabharadwaj198:
Fixing the bug where if the original document didn't contain any map fields, adding new maps with docs resulted in map type issues & later didn't allow us to properly update maps in subsequent  calls. 

Also added more robust testing where we are now testing the field_type metadata & running every test through 3 different docs

* Bug fix

* In case a user didn't send in a map in an add_docs request, subsequent addition of a map field in an update docs request will lead to the map not getting replaced 

* Even if a user's original documents didn't have a map to begin with, updates to maps will work as expected. 

No

* Added more unit tests. Now every unit test checks if the marqo_field_types metadata field is having the correct value for a field. This is checked after both updation and addition of docs 
* Manually tested the test case for which this was failing:
* Screenshot 2025-03-06 at 4 43 02 PM

* NA

* NA

* **Other information**:
* NA

### DIFF
--- a/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
+++ b/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
@@ -216,8 +216,10 @@ class SemiStructuredVespaIndex(StructuredVespaIndex, UnstructuredVespaIndex):
         Returns:
             Dict containing the Vespa partial document format with:
             - 'id': Document ID
-            - 'field_types': Field name to type mapping
-            - 'fields': Field values
+            - 'field_types': Field name to type mapping. Later used to create pre-conditions
+            - 'fields': Field values. Each field is represented as an update statement, for the actual field, the field type metadata, and the score modifiers if applicable. Example:
+                - 'marqo__bool_fields{active}': {"assign": 1}
+                - 'marqo__field_type{active}': {"assign": "bool"}
             - 'create_timestamp': Original document timestamp if it exists
 
         Raises:
@@ -238,8 +240,6 @@ class SemiStructuredVespaIndex(StructuredVespaIndex, UnstructuredVespaIndex):
         vespa_fields = {}
         vespa_field_types = {}
 
-        # Initialize dictionary to be later used for updating score modifiers. 
-        numeric_fields = {}
 
         numeric_field_map: Dict[str, Any] = dict() # This map is used to store the numeric fields in the document. It is used to update the numeric fields & score modifiers later
         if original_doc:
@@ -254,17 +254,10 @@ class SemiStructuredVespaIndex(StructuredVespaIndex, UnstructuredVespaIndex):
             validate_field_name(field_name)
 
             # This method broadly processes the field based on its type and updates the vespa_fields,
-            # vespa_field_types, numeric_fields, numeric_field_map dictionaries. Numeric fields and numeric field maps
+            # vespa_field_types, numeric_field_map dictionaries. Numeric fields and numeric field maps
             # are special cases and are processed later.
-            self._process_field(
-                field_name=field_name,
-                value=value,
-                fields=vespa_fields,
-                field_types=vespa_field_types,
-                numeric_fields=numeric_fields,
-                numeric_field_map=numeric_field_map,
-                doc_id=doc_id
-            )
+            self._process_field(field_name=field_name, value=value, fields=vespa_fields, field_types=vespa_field_types,
+                                numeric_field_map=numeric_field_map, doc_id=doc_id)
 
         # This method creates the update statement for updating int fields / int map fields.
         int_fields_changed = self._create_update_statement_for_updating_numeric_and_numeric_map_field(
@@ -342,16 +335,8 @@ class SemiStructuredVespaIndex(StructuredVespaIndex, UnstructuredVespaIndex):
         if len(score_modifiers) > 0:
             vespa_fields[common.SCORE_MODIFIERS] = score_modifiers
 
-    def _process_field(
-        self,
-        field_name: str,
-        value: Any,
-        fields: Dict[str, Any],
-        field_types: Dict[str, Any],
-        numeric_fields: Dict[str, Any],
-        numeric_field_map: Dict[str, Any],
-        doc_id: str
-    ) -> None:
+    def _process_field(self, field_name: str, value: Any, fields: Dict[str, Any], field_types: Dict[str, Any],
+                       numeric_field_map: Dict[str, Any], doc_id: str) -> None:
         """Process a single field from a document based on its type.
 
         This method determines the type of the field value and delegates processing to the appropriate handler method.
@@ -361,8 +346,7 @@ class SemiStructuredVespaIndex(StructuredVespaIndex, UnstructuredVespaIndex):
             field_name: The name of the field being processed
             value: The value of the field, can be of the type bool, dict, int, float, list, or str
             fields: Dictionary to store the update statements corresponding to the processed fields 
-            field_types: Dictionary mapping field names to their Marqo field types
-            numeric_fields: Dictionary storing numeric field values for being later used to update score modifier 
+            field_types: Dictionary mapping field names to their Marqo field types. Later used to create pre-conditions.
             doc_id: The ID of the document containing this field
 
         Raises:
@@ -538,8 +522,10 @@ class SemiStructuredVespaIndex(StructuredVespaIndex, UnstructuredVespaIndex):
             # Set the appropriate field type based on the value type
             if isinstance(v, int):
                 field_types[f'{field_name}.{k}'] = MarqoFieldTypes.INT_MAP.value
+                field_types[f'{field_name}'] = MarqoFieldTypes.INT_MAP.value
             else:  # Must be float based on the earlier check
                 field_types[f'{field_name}.{k}'] = MarqoFieldTypes.FLOAT_MAP.value
+                field_types[f'{field_name}'] = MarqoFieldTypes.FLOAT_MAP.value
 
     def _handle_string_array_field(
         self,
--- a/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
+++ b/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
@@ -1,4 +1,3 @@
-import json
 from typing import Dict, Any, List, Optional, Type, Union, cast
 
 from marqo.core.constants import MARQO_DOC_HIGHLIGHTS, MARQO_DOC_ID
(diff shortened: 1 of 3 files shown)


tests/integ_tests/tensor_search/integ_tests/test_add_documents_semi_structured.py
```diff
@@ -781,4 +782,102 @@ def test_a_text_index_will_treat_a_url_as_text(self):
...
+                self.assertEqual(vespa_fields['marqo__field_types']['float_map.c'], MarqoFieldTypes.FLOAT_MAP.value, f"Expected float_map.c to have type {MarqoFieldTypes.FLOAT_MAP.value} for document {id}")
+                self.assertEqual(vespa_fields['marqo__field_types']['float_map.d'], MarqoFieldTypes.FLOAT_MAP.value, f"Expected float_map.d to have type {MarqoFieldTypes.FLOAT_MAP.value} for document {id}")
+                self.assertEqual(vespa_fields['marqo__field_types']['bool_field'], MarqoFieldTypes.BOOL.value, f"Expected bool_field to have type 'bool' for document {id}")
+                self.assertEqual(vespa_fields['marqo__field_types']['short_string_field'], MarqoFieldTypes.STRING.value, f"Expected short_string_field to have type 'string' for document {id}")
+                self.assertEqual(vespa_fields['marqo__field_types']['long_string_field'], MarqoFieldTypes.STRING.value, f"Expected long_string_field to have type 'string' for document {id}")
+                self.assertEqual(vespa_fields['marqo__field_types']['int_field'], MarqoFieldTypes.INT.value, f"Expected int_field to have type {MarqoFieldTypes.INT_MAP.value} for document {id}")
```
papa99do: nit. type in the assertion message does not match expected type.

tests/integ_tests/tensor_search/integ_tests/test_add_documents_semi_structured.py
```diff
@@ -781,4 +782,102 @@ def test_a_text_index_will_treat_a_url_as_text(self):
...
+                self.assertEqual(vespa_fields['marqo__field_types']['float_field'], MarqoFieldTypes.FLOAT.value, f"Expected float_field to have type 'float' for document {id}")
+                self.assertEqual(vespa_fields['marqo__field_types']['bool_field2'], MarqoFieldTypes.BOOL.value, f"Expected bool_field2 to have type 'bool' for document {id}")
+                self.assertEqual(vespa_fields['marqo__field_types']['custom_vector_field'], MarqoFieldTypes.TENSOR.value, f"Expected custom_vector_field to have type 'custom_vector' for document {id}")
```
papa99do: Do we have a `custom_vector` type? (in assertion message) should it be `tensor`?

tests/integ_tests/tensor_search/integ_tests/test_add_documents_semi_structured.py
```diff
@@ -781,4 +782,102 @@ def test_a_text_index_will_treat_a_url_as_text(self):
...
+                "content": "abcd",
+                "vector": [1.0] * 32
+            }
+        }
+        self.id_to_doc = {
+            '1': self.doc,
+            '2': self.doc2,
+            '3': self.doc3
+        }
+        self.add_documents(self.config, add_docs_params=AddDocsParams(
```
papa99do, line 842: Suggestion for an edge case:  add self.doc3 in another batch and do not specify tensor_field and multimodal_combo_field as tensor field. In this case. We expect both `tensor_field` and `tensor_subfield` to be just string. This verified that different batch can have different type for the same field.

tests/integ_tests/core/document/test_partial_update_semi_structured.py
```diff
@@ -133,6 +136,11 @@ def test_partial_update_should_update_bool_field(self):
                 self.assertFalse(updated_doc['bool_field'], f"Expected bool_field to be False for document {id}")
                 self._assert_fields_unchanged(updated_doc, ['bool_field'])
 
+                # Verify field type
+                raw_vespa_doc = self.config.vespa_client.get_document(id, self.config.index_management.get_index(self.index.name).schema_name)
```
papa99do: we can use `self.index.schema_name`

tests/integ_tests/tensor_search/integ_tests/test_add_documents_semi_structured.py
adityabharadwaj198: thanks for catching that!

tests/integ_tests/tensor_search/integ_tests/test_add_documents_semi_structured.py
adityabharadwaj198: we don't have that. Thanks for catching that!

tests/integ_tests/core/document/test_partial_update_semi_structured.py
papa99do: Also, this section of code seems to be duplicated a lot of times in this test, consider extract a method? something like `_assert_field_type(id, [(field_name, field_type)])`?

tests/integ_tests/core/document/test_partial_update_semi_structured.py
```diff
@@ -156,400 +164,640 @@ def test_partial_update_should_update_int_field_to_int(self):
...
+                self.assertEqual(doc['int_map.d'], 2, f"Expected int_map.d to be 2 for document {id}")
+                
+                # Verify field type
+                raw_vespa_doc = self.config.vespa_client.get_document(id, self.config.index_management.get_index(self.index.name).schema_name)
+                vespa_fields = raw_vespa_doc.document.dict().get('fields')
+                self.assertEqual(vespa_fields['marqo__field_types']['int_map.d'], MarqoFieldTypes.INT_MAP.value, 
+                              f"Expected int_map.d to have type {MarqoFieldTypes.INT_MAP.value} for document {id}")
 
+                #Verify that the int_map.a and int_map.b is removed from the field types
+                self.assertEqual(vespa_fields['marqo__field_types'].get('int_map.a', None), None)
```
papa99do: nit. this is equivalent to `self.assertNotIn('int_map.a', vespa_fields['marqo__field_types'])`

tests/integ_tests/core/document/test_partial_update_semi_structured.py
papa99do, line 1: We have two test cases `test_partial_update_should_reject_updating_dict_to_int_field` and `test_updating_int_map_to_int` to cover updating numeric_map to int. Do we have any test case to cover updating numeric fields to numeric_map fields?

Also, can we write a test case just to cover adding a new map and then try to update it (with some fields removed)?

tests/integ_tests/tensor_search/integ_tests/test_add_documents_semi_structured.py
adityabharadwaj198: thanks, i'll add that

tests/integ_tests/core/document/test_partial_update_semi_structured.py
adityabharadwaj198: good idea! I have added a test case to check this `test_updating_int_to_int_map`

tests/integ_tests/core/document/test_partial_update_semi_structured.py
adityabharadwaj198: have created a method to do this as suggested by you above, so have removed this statement.

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
```diff
@@ -538,8 +522,10 @@ def _handle_dict_field(
             # Set the appropriate field type based on the value type
             if isinstance(v, int):
                 field_types[f'{field_name}.{k}'] = MarqoFieldTypes.INT_MAP.value
+                field_types[f'{field_name}'] = MarqoFieldTypes.INT_MAP.value
```
papa99do, line 525: would this become a problem if the map contains both float and int? We probably should have a `numeric_map` field type for the map itself.
For example, in the original doc,  `'map1':  {'a': 1}`, and when we update it to `'map1': {'b': 2.0}`, its type stays at `INT_MAP`, which is not very accurate.

tests/integ_tests/core/document/test_partial_update_semi_structured.py
```diff
@@ -112,6 +112,15 @@ def _assert_fields_unchanged(self, doc: Dict[str, Any], excluded_fields: List[st
             else:
                 self.assertEqual(value, doc.get(field, None), f'{field} is changed.')
 
+    def _assert_field_types(self, field_names, field_types, id):
+        raw_vespa_doc = self.config.vespa_client.get_document(id, self.index.schema_name)
+        vespa_fields = raw_vespa_doc.document.dict().get('fields')
+
+        for field_name, field_type in zip(field_names, field_types):
```
papa99do: nit. instead of zipping two lists, it might be more readable to do Tuple[(str, type)] .So when we pass them in, we can group the field name and type together. For example:
```
self._assert_field_types(id, [
  ('field1', BOOL), 
  ('field2', INT),
  ('field3', None)
])
```

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
adityabharadwaj198: currently we don't support a map with both int & float keys within it. And we also don't support type changes of keys from float to int etc. If we want to have that in place it will require a bigger change I believe.

tests/integ_tests/core/document/test_partial_update_semi_structured.py
adityabharadwaj198: done

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
```diff
@@ -424,37 +417,31 @@ def _update_numeric_field(
...
-                
-                # Set appropriate field type based on numeric_type and whether it's a map
-                field_type = (MarqoFieldTypes.INT_MAP if numeric_type is int else MarqoFieldTypes.FLOAT_MAP) if is_map_type else (MarqoFieldTypes.INT if numeric_type is int else MarqoFieldTypes.FLOAT)
+            if not field_exists_in_original_doc or field_value_changed:
+                vespa_fields[vespa_field_name] = {"assign": field_value}
 
-                # Set field type metadata by creating a assigned statement
-                vespa_fields[vespa_field_types_field_name] = {"assign": field_type.value}
+                # Set field type metadata by creating an assign statement
+                vespa_fields[vespa_field_types_field_name] = {"assign": field_type}
```
papa99do, line 436: I noticed this change here. What is the type of `vespa_field_types`? is field_type a `str` or `MarqoFieldTypes`?

adityabharadwaj198: It's supposed to be str itself. In the [original code](https://github.com/marqo-ai/marqo/pull/1139/files/82a994ab1e96b09f7114901c0b52a788259b2848#diff-3d90f53e58a621604f7dcf97912190162a5239578b0ca47cbf9ac704536aa91dL447) you'll see that field_type was the MarqoFieldTypes enum, that's why we needed to do field_type.value. But in this PR I am populating it with the actual string value of that enum (https://github.com/marqo-ai/marqo/pull/1139/files/82a994ab1e96b09f7114901c0b52a788259b2848#diff-3d90f53e58a621604f7dcf97912190162a5239578b0ca47cbf9ac704536aa91dL447) so we don't need to do a .value again

papa99do (approved):

PR #1140, opened 2025-03-07
Change create_timestamp to version_uuid 

adityabharadwaj198:
* Changes the using a create_timestamp value for consistency token to a version_uuid value which is a uuid. 
* Everytime a get and set operation happens for partial update (which happens when the document contains a map), the version_uuid will be updated 

* Currently on a get & set op, the create_timestamp is checked. It's not updated

* on a get & set op, the version_uuid will be checked. Post the check passing, version_uuid will also be updated with a new uuid. This acts as a consistency token in a optimistic locking type mechanism to avoid race conditions. 

* No

* Yes. Ran parallel update operations to see if the update is happening properly: https://www.notion.so/Analysis-after-updating-map-fields-with-parallel-threads-1ae75d43da4c80dd8ceade5bbcd91771?pvs=4. This indicates that the version uuid gets updated properly. 
* Existing unit tests for partial updates pass

* **Other information**:


src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_schema_template_2_16.sd.jinja2
```diff
@@ -7,11 +7,9 @@ schema {{ index.schema_name }} {
             rank: filter
         }
 
-        field marqo__create_timestamp type double {
-                    indexing: attribute | summary
-                    attribute: fast-search
-                    rank: filter
-                }
+        field marqo__version_uuid type string {
+                indexing: summary
```
papa99do: Can we check if define this as `indexing: attribute | summary` boosts update performance?

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
```diff
@@ -276,11 +277,16 @@ def to_vespa_partial_document(self, marqo_document: Dict[str, Any], existing_ves
                 vespa_fields=vespa_fields
             )
 
+        # Add version_uuid to the vespa_fields to update the document's version_uuid,
+        # only if this is a type of update that requires updating version_uuid (i.e a partial update with map fields)
+        if original_doc is not None and original_doc.fixed_fields.version_uuid:
```
papa99do, line 283: * In what circumstance will we get a doc without the version_uuid?
* Also add a test case for this to verify that uuid is updated if map field is change.
* Is it possible to simulate the race condition in a integ test? I guess it might be tricky, right?

papa99do: Also add a test case for this to verify that uuid is updated

adityabharadwaj198: > In what circumstance will we get a doc without the version_uuid?

We won't ever get a document without a version_uuid. However there could be a case where we don't get the document at all (i.e for update requests that don't contain maps). In which case this if block won't execute. 

> Also add a test case for this to verify that uuid is updated if map field is change.

Added this test in score modifiers test case since we're testing the fields that are not returned in a get_docs call in that test case and also because version_uuid always gets changed when we process an update request that contains maps (similar to score modifiers which also get changed in this case)

adityabharadwaj198: > Is it possible to simulate the race condition in a integ test? I guess it might be tricky, right?

Yes it was tricky to do as a integ test. But I have simulated it as a API test here: https://github.com/marqo-ai/marqo/pull/1140/files#diff-d727aa22399621a1e57beb4a1791871a899a35d06c248fe4827564cad179beacR164.

src/marqo/core/document/document.py
```diff
@@ -13,7 +13,7 @@
     UnstructuredMarqoIndex
 from marqo.core.models.marqo_update_documents_response import MarqoUpdateDocumentsResponse, MarqoUpdateDocumentsItem
 from marqo.core.semi_structured_vespa_index.common import SEMISTRUCTURED_INDEX_PARTIAL_UPDATE_SUPPORT_VERSION, \
-    VESPA_FIELD_ID, INT_FIELDS, FLOAT_FIELDS, VESPA_DOC_FIELD_TYPES, VESPA_DOC_CREATE_TIMESTAMP
+    VESPA_FIELD_ID, INT_FIELDS, FLOAT_FIELDS, VESPA_DOC_FIELD_TYPES, VESPA_DOC_VERSION_UUID
```
farshidz, line 16: See my other comment about naming this. I think calling it document version makes it easier to understand

src/marqo/core/semi_structured_vespa_index/common.py
```diff
@@ -24,7 +24,7 @@
 
 # A metadata field that's used to store a dictionary of key-value pairs where key is the field name and value is a string denoting the field type
 VESPA_DOC_FIELD_TYPES = "marqo__field_types"
-VESPA_DOC_CREATE_TIMESTAMP = "marqo__create_timestamp"
+VESPA_DOC_VERSION_UUID = "marqo__version_uuid"
```
farshidz, line 27: I don't think having uuid in the name makes sense. How about `marqo__document_version`?

src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
```diff
@@ -8,17 +8,20 @@
...
-from marqo.core.models.marqo_index import SemiStructuredMarqoIndex, logger
+from marqo.core.models.marqo_index import SemiStructuredMarqoIndex
 from marqo.core.semi_structured_vespa_index import common
 from marqo.core.semi_structured_vespa_index.common import VESPA_DOC_FIELD_TYPES, STRING_ARRAY
 from marqo.core.semi_structured_vespa_index.marqo_field_types import MarqoFieldTypes
 from marqo.core.unstructured_vespa_index.common import MARQO_DOC_MULTIMODAL_PARAMS, MARQO_DOC_MULTIMODAL_PARAMS_WEIGHTS
 
 
+def generate_uuid_str() -> str:
+    return str(uuid.uuid4())
```
farshidz: We should at least remove the hyphens to make it smaller (reduces the size by 4 bytes). However, have you looked at storing it in a byte field in Vespa? I'm guessing a byte field won't work for our update condition, but worth looking into. In its byte form the UUID takes up 16 bytes, vs 32 bytes with hyphens removed, or 36 byte including hyphens.

src/marqo/core/semi_structured_vespa_index/common.py
adityabharadwaj198: me and @papa99do discussed on the naming before finalising this. marqo__document_version may be confusing since 
1. It’s not an actual version number (e.g., 1, 2, 3, ...).
2. It doesn’t track document history, just prevents conflicting updates.
3. UUIDs are random, not sequential, so calling it a "version" might confuse people expecting version increments.

@papa99do what do you think? I'm open to changing this, however if having uuid in the name is a concern, we could do something like 'concurrency_token' or 'optimistic_lock_token' or 'update_token' to abstract away the uuid related low level implementation detail.

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_schema_template_2_16.sd.jinja2
adityabharadwaj198: yes it does, added back `attribute | summary`

src/marqo/core/semi_structured_vespa_index/common.py
papa99do: It's an internal field, I'm not too concerned about naming since user will not see it.  `update_token` makes sense to me, and is closer to the intention of this field.

farshidz: Makes sense. Let's leave it as is then

src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
adityabharadwaj198: I have removed hyphens. Discussed offline re: storing uuid as bytes

tests/api_tests/v1/tests/api_tests/unstructured_index/test_update_documents_unstructured_index.py
```diff
@@ -158,6 +161,107 @@ def test_update_document_with_ids_change_field_type(self):
         self.assertIn("Marqo vector store couldn't update the document. Please see", update_docs_response['items'][0]['message'])
         self.assertIn("reference/api/documents/update-documents/#response", update_docs_response['items'][0]['message'])
 
+    def test_concurrent_partial_update_requests(self):
```
papa99do, line 164: If we comment out the version_uuid update code, can this test capture concurrency issue?

tests/integ_tests/core/document/test_partial_update_semi_structured.py
```diff
@@ -523,13 +523,24 @@ def test_partial_update_should_update_short_string_to_long_string(self):
...
+        """
         test_docs = [self.doc2, self.doc3]
-        
+
+        version_uuid = {}
+
         for doc in test_docs:
             with self.subTest(f"Updating score modifiers for document with ID {doc['_id']}"):
                 id = doc['_id']
+                raw_vespa_doc = self.config.vespa_client.get_document(id, self.config.index_management.get_index( # Doing a get to set the version_uuid, which we'll check later to make sure it has changed after
```
papa99do: nit. the comment seems to be out of place, can you reformat it? 
also `self.config.index_management.get_index(self.index.name).schema_name` -> `self.index.schema_name`?

adityabharadwaj198: The comment makes sense. I have added a bit more details to it to make it like `# Doing a get to set the version_uuid in the version_uuid hashmap, which we'll check later to make sure it has changed after
                # processing an update request that contains maps`
                
Changed self.config.index_management.get_index(self.index.name).schema_name -> self.index.schema_name

tests/api_tests/v1/tests/api_tests/unstructured_index/test_update_documents_unstructured_index.py
adityabharadwaj198: yes, here's a screenshot: 
image

farshidz (approved):

papa99do (approved):

PR #1144, opened 2025-03-12
Fix map field's prefix not being set in marqo__field_types when partial updating + add api tests 

adityabharadwaj198:
1. Fixed a bug where when we insert a map using partial updates, the map prefix in a flattened map field name was not getting set in marqo__field_types
2. Adds some API tests in correspondence with the set of manual tests that we were running

* wrt the bug - currently if a map is inserted via a partial update (it must not exist from before), the marqo__field_types will consist of int_map.key1, int_map.key2 but no int_map. This doesn't happen when inserting maps via add_docs and is critical to prevent type conversions.

* if a map is inserted via a partial update (it must not exist from before), the marqo__field_types will consist of int_map.key1: int_map_entry, int_map.key2: int_map_entry & int_map: int_map_entry. 

* No

* Yes

* NA

* **Other information**:


src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
```diff
@@ -437,12 +437,18 @@ def _create_update_statement_for_updating_numeric_and_numeric_map_field(
...
+                vespa_fields[vespa_field_name] = {"assign": field_value} # assign statement for adding / updating the field value in marqo__int_fields / marqo__float_fields
 
                 # Set field type metadata by creating an assign statement
-                vespa_fields[vespa_field_types_field_name] = {"assign": field_type}
+                vespa_fields[vespa_field_types_field_name] = {"assign": field_type} # assign statement for adding / updating the field type in marqo__field_types
+                
+                # Handle creating update statements for the map field name. 
+                if "." in field_name:
+                    # For fields like "map1.key1", extract the map name "map1"
+                    map_name = field_name.split(".", 1)[0]
```
farshidz: Did we have this logic in a function that we could reuse?

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
```diff
@@ -453,12 +459,17 @@ def _create_update_statement_for_updating_numeric_and_numeric_map_field(
             if (original_field_name not in numeric_field_map and
                 original_doc.fixed_fields.field_types.get(original_field_name) in (MarqoFieldTypes.INT_MAP.value, MarqoFieldTypes.FLOAT_MAP.value)):
 
+                map_name = original_field_name.split(".", 1)[0]
```
farshidz: Should have a func rather than repeat the split logic. Probably even a func to detect if it's a map (contains `.`)

tests/api_tests/v1/tests/api_tests/unstructured_index/test_update_documents_unstructured_index.py
```diff
@@ -102,8 +102,131 @@ def test_update_document_with_ids(self):
...
+        self.assertEqual(get_docs_response['new_int_field'], 42)
+        self.assertEqual(get_docs_response['new_float_field'], 3.14159)
+        self.assertEqual(get_docs_response['new_bool_field'], True)
+        self.assertEqual(get_docs_response['new_int_map.key1'], 100)
+        self.assertEqual(get_docs_response['new_int_map.key2'], 200)
+        self.assertEqual(get_docs_response['new_float_map.pi'], 3.14)
+        self.assertEqual(get_docs_response['new_float_map.e'], 2.718)
+
+    def test_add_new_tensor_fields_with_update_documents_api(self):
+        """Test that adding new tensor fields, string arrays, and strings via update_documents fails.
```
farshidz: I think in general you're putting too many things in your tests. This one for example looks like at least three separate tests. In general if you have a lot of asserts, you're probably testing too many things at once.

Also print statement in test is unusual and not good practice usually. If there's an issue, raise an error and have the string as its message

farshidz (changes requested):
I don't think there's a test that covers the bug you're fixing in this PR

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
adityabharadwaj198: introduced a method for this.

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
adityabharadwaj198: introduced a method for this.

tests/api_tests/v1/tests/api_tests/unstructured_index/test_update_documents_unstructured_index.py
adityabharadwaj198: I've broken down some tests that I could see fit. Please let me know if you think any other tests can be split also. 

Deleted all print statements. Some are still there in the tests with threads and can help in debugging. let me know if you want me to delete them.

farshidz (changes requested):
One issue with _assert_field_types. It verifies expected fields are found, but doesn't check for extra fields in the Vespa doc. You want to make sure the Vespa doc has all required field types and only those fields not more

adityabharadwaj198:
Releases/2.16 branch version of this PR is approved and will be merged into mainline. Closing this PR in favor of that one: https://github.com/marqo-ai/marqo/pull/1146

PR #1156, opened 2025-03-19
Use the new native inference implementation for add_docs and search

papa99do:
* Refactoring

  * add_docs and search calls s2_inference.vectorise to generate embedding
  * chunking logic when add docs is handled in tensor_fields_container
  * 3 vectorisation batching mode is implemented in add_documents_handler
  * for unstructured index, modality is inferred twice, one in add_documents_handler, one in threaded download. 
  * for structured index, modality is inferred in threaded donwload, and checked with the field type
  * for search, modality is inferred 3 times, 
     -  1st: when we add_prefix to queries
     -  2nd: when we create_vector_jobs
     -  3rd: before we send the request to inference

  * add_docs and search (run_vectorising_jobs) calls the new Inference.vectorise interface to generate embedding
  * chunking logic is removed from tensor_fields_container. Now chunking and preprocessing is done in Inference
  * when add_docs, the tensor fields will be collected based on modality, and sent to Inference in a batch, per modality
  * we consolidated the modality infer logic to be in add_documents_handler for add_docs, before sending to Inference. For structured index, we will check the field type and do not send to inference if there's a mismatch.
  * for search, we removed the 3rd modality inference, and use the value from the 2nd.

  * No

Yes

No

No

* **Other information**:


src/marqo/core/inference/dummy_inference.py
```diff
@@ -0,0 +1,32 @@
+from typing import Optional
+
+import numpy as np
+
+from marqo.core.inference.api import Inference, InferenceResult, InferenceRequest, InferenceError, InferenceErrorModel
+
+
+class DummyInference(Inference):
```
wanliAlex: Is this still needed?

src/marqo/core/inference/tensor_fields_container.py
```diff
@@ -1,359 +1,170 @@
...
+            return not self.embeddings
+        else:
+            # text and image fields stores extra embedding for subfield without chunking
+            # in multimodal_subfield_embedding field
+            return not self.multimodal_subfield_embedding
 
-class Vectoriser(ABC):
-    @abstractmethod
-    def vectorise(self, content_chunks: List[ContentChunkType], key_prefix: str = None) -> List[List[float]]:
+    def populate_chunks_and_embeddings(self, chunks: List[str], embeddings: List[List[float]],
```
wanliAlex, line 61: Why do we need to populate the embeddings later? Is it because we need to instantiate the object first, do some processes, and populate embeddings later?

src/marqo/core/structured_vespa_index/structured_add_document_handler.py
```diff
@@ -58,12 +69,25 @@ def _handle_field(self, marqo_doc, field_name, field_content):
...
 
-    def _infer_field_type(self, field_name:str, field_content: Any) -> FieldType:
-        return self.marqo_index.field_map[field_name].type
+    def _infer_modality(self, tensor_field: TensorField) -> Modality:
+        if tensor_field.field_type == FieldType.Text:
+            return Modality.TEXT
+
+        url = tensor_field.field_content
+        try:
+            modality = infer_modality(url, self.add_docs_params.media_download_headers)
```
wanliAlex, line 82: Could this lead to inconsistent behaviour? I think previously we just process the content based on the given modality, and rely on the preprocessor to raise an error.

src/marqo/core/vespa_index/add_documents_handler.py
```diff
@@ -266,162 +270,130 @@ def _validate_doc(self, doc) -> None:
...
+                               f'{r.error_message}')
+                erroneous_doc_ids.add(doc_id)
+                self.tensor_fields_container.remove_doc(doc_id)
+                self.add_docs_response_collector.collect_error_response(
+                    doc_id,
+                    AddDocumentsError(error_message=r.error_message, error_code=r.error_code, status_code=r.status_code)
+                )
+            else:
+                field.populate_chunks_and_embeddings(
+                    chunks=[chunk for chunk, _ in r],
```
wanliAlex: Can we do this in one loop over `r`?

wanliAlex (changes requested):

src/marqo/core/inference/dummy_inference.py
papa99do: Not really, not used in any place. will remove

src/marqo/core/inference/modality_utils.py
```diff
@@ -0,0 +1,122 @@
+import io
```
wanliAlex, line 1: This is just the code we have before right?

src/marqo/core/inference/tensor_fields_container.py
wanliAlex: Ah I see, the difficulty is that a tensor_field might have one text field, and one image field and the embeddings can't be collected at the same time.

papa99do: These tensor fields are collected to the container when parsing documents. And then the existing tensor will also be populated. Those that need vectorisation will then be collected to send to Inference. This method is used in various stages to populate chunks and embeddings for different types of tensor fields. As mentioned in the comment

src/marqo/core/structured_vespa_index/structured_add_document_handler.py
papa99do: That's correct, and we cannot do that now since Inference does not know about the field type. The good thing about doing it earlier is that if there's any mismatch, we fail early and do not send to Inference.

src/marqo/core/inference/modality_utils.py
papa99do: yes, I copied it here, and will delete the other one once all deps are removed

src/marqo/core/vespa_index/add_documents_handler.py
papa99do: Not sure if this is better: 

```
# unzip the result for each content. format of r is [(chunk1, embedding1), (chunk2, embedding2)]
# after the unzipping, chunks is (chunk1, chunk2), embeddings is (embedding1, embedding2)
chunks, embeddings = zip(*r)
field.populate_chunks_and_embeddings(
    chunks=list(chunks),
    embeddings=[embedding.tolist() for embedding in embeddings],
    for_top_level_field=for_top_level_field
)
```

src/marqo/config.py
```diff
@@ -50,8 +51,9 @@ def __init__(
                                                 enable_index_operations=True,
                                                 deployment_lock_timeout_seconds=deployment_lock_timeout)
 
+        self.inference = NativeInferenceLocal()
```
farshidz, line 54: Will another PR make this configurable to remote inference?

src/marqo/core/inference/api/preprocessing_config.py
```diff
@@ -52,7 +52,7 @@ def validate_chunk_config(cls, values):
 class ImagePreprocessingConfig(PreprocessingConfig):
     """Preprocessing config for image modality"""
     modality: Literal[Modality.IMAGE] = Modality.IMAGE
-    download_timeout_ms: Optional[int] = pydantic.Field(default=None, alias='downloadTimeoutMs')
+    download_timeout_ms: int = pydantic.Field(default=3000, alias='downloadTimeoutMs')  # default to 3000ms
```
farshidz, line 55: Is this default limit current behavior?

src/marqo/core/inference/dummy_inference.py
papa99do: removed

src/marqo/tensor_search/tensor_search.py
```diff
@@ -752,43 +756,66 @@ def create_vector_jobs(queries: List[BulkSearchQueryEntity], config: Config, dev
...
+                    model_properties=v.model_properties,
                     model_auth=v.model_auth,
-                    enable_cache=True,
-                    modality=modality
-                )
-                result[v.groupby_key()] = dict(zip(v.content, vectors))
+                    normalize_embeddings=v.normalize_embeddings,
+                ),
+                device=v.device,
+                use_inference_cache=True,
```
farshidz, line 789: Are hard-coding cache enabled?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -752,43 +756,66 @@ def create_vector_jobs(queries: List[BulkSearchQueryEntity], config: Config, dev
...
+                preprocessing_config=_get_preprocessing_config(v.modality, v.media_download_headers),
+            )
+
+            inference_result = inference.vectorise(inference_request)
+
+            # Sanity check the response from Inference
+            if len(inference_result.result) != len(v.content):
+                raise InternalError(f'Inference result contains embeddings for {len(inference_result.result)} '
+                                    f'query items, but {len(v.content)} is expected')
+            if any([isinstance(r, InferenceErrorModel) for r in inference_result.result]):
```
farshidz: Aren't we losing the individual errors?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -752,43 +756,66 @@ def create_vector_jobs(queries: List[BulkSearchQueryEntity], config: Config, dev
...
+                raise InternalError(f'Individual errors returned when vectorising query: {v.content}')
+            if any([len(chunks) > 1 for chunks in inference_result.result]):
+                raise InternalError(f'Some query items have multiple chunks: {v.content}')
```
farshidz: Again not sure if there will be enough data to debug effectively

src/marqo/core/inference/api/preprocessing_config.py
papa99do: yes, the current default is 3000ms. I added this here so its easier to expose this config to user in the future.
Current default: https://github.com/marqo-ai/marqo/blob/yihan/mw-cloud-perf-0320/src/marqo/inference/media_download_and_preprocess/image_download.py#L146

src/marqo/tensor_search/tensor_search.py
papa99do: Yes, just to keep the previous logic. There's another control point in inference where we enable the cache. Inference caching is not in the new implementation yet. It's covered in this ticket: https://s2search.atlassian.net/browse/MQ-317

src/marqo/tensor_search/tensor_search.py
papa99do: The purpose is to keep the logic same as before. For search, we don't keep individual inference errors

src/marqo/tensor_search/tensor_search.py
papa99do: I added some test cases. These checks are just sanity checks to make sure inference works in expected ways for search. Such as no individual errors, and no chunking. Do you mean there's not enough data about whether search should support chunking?

src/marqo/tensor_search/tensor_search.py
farshidz: Can we log them?

src/marqo/tensor_search/tensor_search.py
farshidz: I mean the specific errors aren't logged and are lost

src/marqo/tensor_search/tensor_search.py
papa99do: you mean the individual errors? they'll be logged in inference. I'm not sure if we should log them here since it's not supposed to happen. But yes, I can add logs if that helps

src/marqo/tensor_search/tensor_search.py
papa99do: OK, I can log them

src/marqo/tensor_search/tensor_search.py
```diff
@@ -797,10 +797,16 @@ def vectorise_jobs(inference: Inference, jobs: List[VectorisedJobs]) -> Dict[JHa
             if len(inference_result.result) != len(v.content):
                 raise InternalError(f'Inference result contains embeddings for {len(inference_result.result)} '
                                     f'query items, but {len(v.content)} is expected')
-            if any([isinstance(r, InferenceErrorModel) for r in inference_result.result]):
-                raise InternalError(f'Individual errors returned when vectorising query: {v.content}')
-            if any([len(chunks) > 1 for chunks in inference_result.result]):
-                raise InternalError(f'Some query items have multiple chunks: {v.content}')
+            individual_errors = [f'{v.content[index]}: {r.error_message}'
```
farshidz, line 800: How expensive is this loop? Perhaps we could have an errors field in the response to avoid unnecessarily looping through things to check for errors?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -797,10 +797,16 @@ def vectorise_jobs(inference: Inference, jobs: List[VectorisedJobs]) -> Dict[JHa
...
+                                 for index, r in enumerate(inference_result.result)
+                                 if isinstance(r, InferenceErrorModel)]
+            if individual_errors:
+                raise InternalError(f'Individual errors returned when vectorising query string: {individual_errors}')
+            chunked_contents = [(v.content[index], len(chunks)) for index, chunks in enumerate(inference_result.result)
+                                if len(chunks) > 1]
+            if chunked_contents:
+                raise InternalError(f'Tensor query string should not be chunked but some '
+                                    f'query items have multiple chunks: {chunked_contents}')
```
farshidz, line 809: I'm having second thoughts for this one. How large could the chunks data be? Also it's user data so perhaps not a good idea to print out to logs

src/marqo/tensor_search/tensor_search.py
papa99do: very unlikely to be expensive. please keep in mind that this is the search endpoint, and user wouldn't provide a lot of query items in a search.  (we don't support bulk search anyways, right?)

src/marqo/tensor_search/tensor_search.py
papa99do: I'm not logging out the actual chunks, just the query term and the chunk size, for example ('query1', 3) . Not sure if it's a security concern to log out query, but we log it out anyway now.

farshidz (approved):

wanliAlex (approved):

wanliAlex (approved):

PR #1157, opened 2025-03-19
Lexical and Filter String Sanitization

vicilliar:
Bug Fix

Lexical query error: In lexical queries, backslashes are not treated as special characters, so it they cannot escape themselves. This leads to stray unescaped `"` characters when they are followed by escaped backslashes. 

Example bad query:
Vespa 400 because backslash is not escaped, meaning a stray backslash is interpreted literally by vespa, causing Marqo 500
```
q = r'\\"test\\"' 
```
Error:
```
"Could not create query from YQL: query:L1:94 no viable alternative at input '(weakAnd(default contains \"\\\\\"test'"
```

Filter string error: double quotes are not escaped when they are :
1. In a field name (for unstructured indexes)

Example bad query:
```
filter_string=r'fiel\"d_1:test'
```
Error:
```
marqo.vespa.exceptions.VespaStatusError: 400: {"root":{"id":"toplevel","relevance":1.0,"fields":{"totalCount":0},"errors":[{"code":4,"summary":"Invalid query parameter","message":"Could not create query from YQL: query:L1:251 no viable alternative at input '(({targetHits:10, approximate:True, hnsw.exploreAdditionalHits:1990}nearestNeighbor(marqo__embeddings_field_1, marqo__query_embedding))) AND (((marqo__short_string_fields contains sameElement(key contains \"fiel\"d_1'"}]}}
```

2. In an IN statement (for structured indexes)
```
filter_string=r"field_1 in (\"hello,goodbye)"
```
Error:
```
marqo.vespa.exceptions.VespaStatusError: 400: {"root":{"id":"toplevel","relevance":1.0,"fields":{"totalCount":0},"errors":[{"code":4,"summary":"Invalid query parameter","message":"Could not create query from YQL: query:L1:222 no viable alternative at input '(({targetHits:10, approximate:True, hnsw.exploreAdditionalHits:1990}nearestNeighbor(marqo__embeddings_field_1, marqo__query_embedding))) AND marqo__filter_field_1 in (\"\"hello'"}]}}
```

Updated lexical query parser to properly treat backslash as an escape character
Updated filter string logic to always escape the `\` and `"` characters in a filter string

No

In progress

* **Other information**:


tests/integ_tests/tensor_search/integ_tests/test_search_combined.py
```diff
@@ -659,6 +659,136 @@ def test_filter_spaced_fields(self):
...
+                # Define test parameters as tuples expected_ids)
+                eq_test_cases = [
+                    ('text_field_1:(some text)', ["doc1", "doc2"]),
+                    ('text_field_1:(som\\e text)', ["doc1", "doc2"]),
+                    ('text_field_1:(som\\\\e text)', ["doc5"]),
+                    ('text_field_1:(som"e text)', ["doc4"]),
+                    ('text_field_1:(som\\"e text)', ["doc4"]),
+                ]
+
+                unstructured_eq_test_cases = [
```
farshidz, line 691: Why do we need special tests for an index type?

tests/integ_tests/tensor_search/integ_tests/test_search_combined.py
```diff
@@ -659,6 +659,136 @@ def test_filter_spaced_fields(self):
...
+        """
+        for index in [self.unstructured_default_text_index, self.structured_default_text_index]:
+            with self.subTest(index=index.type):
+                docs = [
+                    {"_id": "doc1", "text_field_1": "some text", "float_field_1": 0.5},
+                    {"_id": "doc2", "text_field_1": "some text", "float_field_1": 2},
+                    {"_id": "doc3", "text_field_1": "another text", "float_field_1": 2},
+
+                    # Docs with vespa special chars in field content
+                    {"_id": "doc4", "text_field_1": "som\"e text"},
```
farshidz, line 678: Coverage for characters known to cause issues is good, but we're ignoring other special characters. Also cases focus on special chars in the middle, but not start or end. I suggest:
* Expanding test cases based on the above
* Additionally, using fuzz testing as this is a prime use case

tests/integ_tests/tensor_search/integ_tests/test_search_combined.py
```diff
@@ -978,30 +1108,45 @@ def test_search_query_CanAcceptDifferentSearchMethods(self):
...
-            ('', [])                            # Empty string should return no results
+            ('hello\\normal char', []),
+            ('1\\"2', ['doc1']),
+            ('"exact match"', ['doc2']),
+            ('\\"escaped\\"', ['doc4', 'red_herring_4']),   # Vespa tokenizer removes " so both docs are retrieved
+            ('escaped', ['doc4', 'red_herring_4']),         # Vespa tokenizer removes " so both docs are retrieved
+            ('"exacto" wrong"', ['doc3']),
+            ('""', []),
+            ('"', []),
+            ('back\\\\slash', ['doc5']),    # escaped backslash
```
farshidz, line 1144: Coverage for characters known to cause issues is good, but we're ignoring other special characters. I suggest:
* Expanding test cases based on the above
* Additionally, using fuzz testing as this is a prime use case

tests/integ_tests/tensor_search/test_utils.py
```diff
@@ -267,7 +262,43 @@ def test_parse_lexical_query(self):
             ('sa"ma" hello!', ([], ['sa', 'ma', 'hello!'])),    # Bad opening, good closing
             ('"sam"?', ([], ['sam', '?'])),
             ('"朋友"你好', ([], ['朋友', '你好'])),
+        ]
+        for input, expected_output in cases:
```
farshidz, line 243: Was this test running without asserts before?

src/marqo/core/vespa_index/vespa_index.py
```diff
@@ -211,6 +211,9 @@ def _get_hybrid_score_modifiers(self, hybrid_query: MarqoHybridQuery) -> \
 
         return result
 
+    def escape(self, s: str) -> str:
+        return s.replace('\\', '\\\\').replace('"', '\\"')
```
farshidz: You could do this in a single pass rather than two. Not sure if the difference will be measurable but it would be easy to compare the two in isolation and find out

src/marqo/tensor_search/utils.py
```diff
@@ -214,16 +218,21 @@ def parse_lexical_query(text: str) -> Tuple[List[str], List[str]]:
...
 
-    Users need to escape the backslash itself. (Single \ get ignored) -> q='dwayne \\"the rock\\" johnson'
+    PYTHON NOTE:
+    Users using python need to escape the backslash itself. (Single \ get ignored) -> q='dwayne \\"the rock\\" johnson'
+    Unneeded if using CLI or raw string.
 
     Return:
         2-tuple of <required terms> (for "must" clause) <optional terms> (for "should" clause)
     """
+    CHARACTERS_TO_BE_ESCAPED_IN_VESPA = ('"', '\\')
```
farshidz: Can you add a link to your Vespa doc page for these characters? Are we sure there's nothing else to escape?

src/marqo/tensor_search/utils.py
```diff
@@ -232,14 +241,17 @@ def parse_lexical_query(text: str) -> Tuple[List[str], List[str]]:
...
-                pass
-
+        # Every character immediately after a \\ should be read literally
+        if escape:
+            escape = False
+        elif text[i] == "\\":
+            escape = True
+            # Stray backslashes should be removed (not followed by special char, or is last char)
+            if i == len(text) - 1 or text[i + 1] not in CHARACTERS_TO_BE_ESCAPED_IN_VESPA:
+                blob = blob[:-1]
```
farshidz: I'm guessing this op is a linear time full pass over the list. It would be more efficient if you only added and never removed

src/marqo/tensor_search/utils.py
```diff
@@ -232,14 +241,17 @@ def parse_lexical_query(text: str) -> Tuple[List[str], List[str]]:
...
-                # Read quote literally. Backslash should be passed directly to Vespa.
-                pass
-
+        # Every character immediately after a \\ should be read literally
+        if escape:
+            escape = False
+        elif text[i] == "\\":
+            escape = True
+            # Stray backslashes should be removed (not followed by special char, or is last char)
+            if i == len(text) - 1 or text[i + 1] not in CHARACTERS_TO_BE_ESCAPED_IN_VESPA:
```
farshidz: `i+1` could be out of range?

src/marqo/tensor_search/utils.py
```diff
@@ -232,14 +241,17 @@ def parse_lexical_query(text: str) -> Tuple[List[str], List[str]]:
...
-        if text[i] == '"':
-            # Check if ESCAPED
-            if i > 0 and text[i - 1] == '\\':
-                # Read quote literally. Backslash should be passed directly to Vespa.
-                pass
-
+        # Every character immediately after a \\ should be read literally
+        if escape:
+            escape = False
+        elif text[i] == "\\":
```
farshidz, line 210: This code needs to be heavily unit tested

src/marqo/core/vespa_index/vespa_index.py
```diff
@@ -211,6 +211,9 @@ def _get_hybrid_score_modifiers(self, hybrid_query: MarqoHybridQuery) -> \
 
         return result
 
+    def escape(self, s: str) -> str:
```
farshidz, line 216: This code needs to be heavily unit tested

farshidz (changes requested):

tests/integ_tests/tensor_search/integ_tests/test_search_combined.py
vicilliar: For structured indexes, it's impossible to test for `\` or `"` in the field name, since its impossible to make a doc that has a field name with those characters, so if you attempt a search with a filter string like: `text_f\\ield_1:(some text)`, you'll just get the error `"field 'text_f\\ield_1' does not exist in index`.

Thus we only test this case on unstructured indexes.

tests/integ_tests/tensor_search/integ_tests/test_search_combined.py
vicilliar: The reason I'm using the `"` and `\` characters specifically is because it is what's mentioned in vespa documentation here: https://docs.vespa.ai/en/reference/query-language-reference.html?mode=cloud
```
Quote " and backslash \ characters in text values must be escaped by a backslash, also see [how does backslash escapes work](https://docs.vespa.ai/en/faq.html#how-does-backslash-escapes-work).
```

Which other special characters are you referring to? All our issues that need this PR as a fix are caused by 500s from Vespa that are cause by one of these 2 characters not being escaped.

tests/integ_tests/tensor_search/test_utils.py
vicilliar: No, I just split the test into `test_parse_lexical_query` and `test_parse_lexical_query_escaped_characters`

src/marqo/tensor_search/utils.py
vicilliar: Yep it's here: https://docs.vespa.ai/en/reference/query-language-reference.html?mode=cloud

I'll add it as a comment

src/marqo/core/vespa_index/vespa_index.py
vicilliar: can do, should be an easy fix

src/marqo/tensor_search/utils.py
vicilliar: It will never hit that because when `i` is the last character, the first statement `i == len(text) - 1` will be `true`, so the second statement won't be evaluated.

src/marqo/tensor_search/utils.py
vicilliar: It already is, here: https://github.com/marqo-ai/marqo/blob/joshua/double-backslash-fix/tests/integ_tests/tensor_search/test_utils.py#L247

src/marqo/core/vespa_index/vespa_index.py
vicilliar: done, added unit tests for filtering for both semistructured and structured indexes

tests/integ_tests/tensor_search/integ_tests/test_search_combined.py
vicilliar: see above reply: https://github.com/marqo-ai/marqo/pull/1157#discussion_r2011410032

vicilliar:
Fuzz testing generates strings where each character has 30% chance of being `"` or `\`.
This should clear many edge cases for broken vespa queries.
Example cases:
```
Trying example: test_fuzz_lexical_search_all_characters(
    self=<integ_tests.tensor_search.integ_tests.test_search_combined.TestSearchFuzz testMethod=test_fuzz_lexical_search_all_characters>,
    query='\U000c642e\U000c642e\\""""\\\\\\\\""ľÎ"\\"\U000c642e\\',
)

Trying example: test_fuzz_lexical_search_all_characters(
    self=<integ_tests.tensor_search.integ_tests.test_search_combined.TestSearchFuzz testMethod=test_fuzz_lexical_search_all_characters>,
    query='\\\U0006e13aRËJ\\\\\\',
)
```

src/marqo/core/vespa_index/vespa_index.py
```diff
@@ -213,6 +213,19 @@ def _get_hybrid_score_modifiers(self, hybrid_query: MarqoHybridQuery) -> \
...
+        """
```
wanliAlex, line 217: Is this method used for generating a lexical search query, a filter string, or both?

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
```diff
@@ -156,6 +158,9 @@ def generate_equality_filter_string(node: search_filter.EqualityTerm) -> str:
             return final_filter_string
 
         def generate_range_filter_string(node: search_filter.RangeTerm) -> str:
+            # Escape special characters in field name
+            node.field = self.escape(node.field)
```
wanliAlex, line 162: why do we need to escape the special characters in field names when field name should not contain any special character?

src/marqo/tensor_search/utils.py
```diff
@@ -167,6 +168,10 @@ def parse_lexical_query(text: str) -> Tuple[List[str], List[str]]:
     All other terms go into blob, split by whitespace. blob starts as a string then splits into
     a list by space.
 
+    Backslash will be used to escape " or \, but if it is not followed by " or \,
```
wanliAlex, line 171: Is there any shared logic between how we parse a lexical search query and a filter string? If not, I think what you have is fine. If there is common logic, can we combine them?

src/marqo/core/vespa_index/vespa_index.py
vicilliar: Filter string only. We don't do this for lexical search, the mechanism is different.

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
vicilliar: For structured indexes yes, we don't need it technically, but for queries on an unstructured (semi-structured) index, it's still possible to filter on a field that doesn't technically exist in the index. So query will still be sent to vespa

src/marqo/tensor_search/utils.py
vicilliar: The issue with combining them is they have separate special characters, rules, DSLs, and parsers. The way we generate tokens for both is also slightly different.

Lexical query parsing is done here: https://github.com/marqo-ai/marqo/blob/18149c76bcb81089a0210158f2a6704aaa0f4498/src/marqo/tensor_search/utils.py#L163

Filter string parsing is done here: https://github.com/marqo-ai/marqo/blob/18149c76bcb81089a0210158f2a6704aaa0f4498/src/marqo/core/search/search_filter.py#L284

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
wanliAlex: So we choose to send the query and return 0 hits, instead of erroring out here, right?

src/marqo/core/vespa_index/vespa_index.py
wanliAlex: Got it. Can you add this to the doc string?

src/marqo/tensor_search/utils.py
wanliAlex: Makes sense

wanliAlex (changes requested):

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
vicilliar: Yep exactly, it's the same behavior in unstructured when you try to filter on a field that doesn't exist

src/marqo/core/vespa_index/vespa_index.py
vicilliar: done!

wanliAlex (approved):

farshidz (approved):

PR #1161, opened 2025-03-23
Run API tests with separate inference

farshidz:
PR to test running split API tests


tests/api_tests/v1/docker-compose.yml
```diff
@@ -0,0 +1,150 @@
+services:
```
papa99do, line 1: should we have separate docker-compose files for each test setup? could be less confusing

.github/workflows/cuda_docker_marqo_split.yml
```diff
@@ -0,0 +1,106 @@
...
+    branches:
+      - mainline
+      - releases/*
+    paths-ignore:
+      - '**.md'
+  pull_request:
+    branches:
+      - mainline
+      - releases/*
+      - yihan/*
```
farshidz: Remove

tests/api_tests/v1/docker-compose.yml
```diff
@@ -0,0 +1,151 @@
...
+  # Default profile services
+  marqo:
```
farshidz: Either remove base services or extend these in profiles

wanliAlex: Removed base services and local services

wanliAlex (approved):

PR #1168, opened 2025-03-31
Implement facets feature

danyilq:
# Introduce Facets Feature for Hybrid Search
This PR adds support for faceted search powered by **Vespa's grouping language**. It enables users to retrieve field-level aggregations (facets) alongside search results, such as counts per color or price statistics.

### New facets search parameter
Accepts a dictionary specifying fields to facet on. Example:

```
{
  "fields": {
    "color": {"type": "string"},
    "price": {"type": "number"}
  }
}
```
Each field must explicitly declare a type: "string", "number", or "array".

### Advanced Parameters
- **maxResults**:
  - Controls the maximum number of unique facet values returned per field.
  - Can be set:
    - Globally (at the top level of facets)
    - Per field (overrides global setting)
- **maxDepth** (top-level only):
  - Limits the number of documents considered during facet aggregation.
  - Useful for optimizing performance.
- **order** (per field):
  - Controls sorting order of string/array facets.
  - Accepts "asc" or "desc" based on count (e.g., show most or least frequent values first).
- **ranges** (for number fields):
  - Enables bucketing of numeric values.
  - Each range supports:
    - from / to — either can be omitted (defaults to -Infinity/Infinity)
    - name — optional label for the bucket
  - If name is not provided, ranges are labeled as <from>:<to>, substituting -Infinity/Infinity where needed.
  - Ranges must not overlap.

### Hybrid-only support
Facets are only supported for hybrid search on unstructured indexes.

### Retrieval behavior

For retrievalMethod=Disjunction, a disjunctive combination of lexical and tensor queries is used (OR logic).

### Facet query execution

Facets are retrieved via a separate hybrid search parameter: `marqo__yql.facets`.

Multiple facet queries are passed to hybrid searcher separate by unique delimiter:

```
\n---MARQO-YQL-QUERY-DELIMITER---\n
```
This is required when using the excludeTerms feature, where each unique exclusion set spawns a distinct facet query.

## Response format

Facets are included at the top level of the search response:

```
{
  "hits": [...],
  "facets": {...}
}
```
## Examples Use Case
Search query with color facet:

```
{
  "text": "shirt",
  "facets": {
    "fields": {
      "color": {"type": "string"}
    }
  }
}
```
Expected response structure:

```
{
  "hits": [...],
  "facets": {
    "color": {
      "red": {"count": 3},
      "blue": {"count": 2}
    }
  }
}
```


src/marqo/core/search/search_filter.py
```diff
@@ -69,7 +69,7 @@ def __eq__(self, other):
         return type(self) == type(other) and self.modified == other.modified and self.raw == other.raw
 
     def __str__(self):
-        return f'{self.raw} ({str(self.modified)})'
+        return f'{self.raw} {str(self.modified)}'
```
farshidz: Why this change?

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
```diff
@@ -92,12 +93,112 @@ def to_vespa_query(self, marqo_query: MarqoQuery) -> Dict[str, Any]:
         else:
             raise InternalError(f'Unknown query type {type(marqo_query)}')
 
+    def _get_facets_term(self, facets_parameters: FacetsParameters, exclusion_terms: List[str] = None) -> str:
```
farshidz, line 97: This seems very unit testable but is't unit tested

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -556,9 +556,47 @@ def _to_vespa_hybrid_query(self, marqo_query: MarqoHybridQuery) -> Dict[str, Any
         if hybrid_score_modifiers[constants.MARQO_GLOBAL_SCORE_MODIFIERS]:
             query_inputs.update(hybrid_score_modifiers[constants.MARQO_GLOBAL_SCORE_MODIFIERS])
 
-        tensor_yql = f'select {select_attributes} from {self._marqo_index.schema_name} where {tensor_term}{filter_term}'
-        lexical_yql = f'select {select_attributes} from {self._marqo_index.schema_name} where {lexical_term}{filter_term}'
+        tensor_yql_no_filter = f'select {select_attributes} from {self._marqo_index.schema_name} where {tensor_term}'
```
farshidz: The dependency of semistructured on structured is problematic (I know you didn't introduce it). Let's be 100% sure we're not breaking structured indexes

farshidz: Maybe this is a good time to move these into the semistructured class?

src/marqo/tensor_search/models/api_models.py
```diff
@@ -172,6 +175,38 @@ def validate_search_method(cls, value):
             case_sensitive=False
         )
 
+    @root_validator(pre=False)
+    def validate_facets_only_for_hybrid_search(cls, values):
```
farshidz, line 180: All validation should be unit tested, including other models in core you've added

tests/api_tests/v1/tests/api_tests/structured_index/test_hybrid_search.py
```diff
@@ -670,5 +670,194 @@ def test_hybrid_search_default_parameters(self):
...
+                            ]
+                        }
+                    }
+                }
+            )
+            self.assertIn("facets", res)
+            self.assertIn("price", res["facets"])
+            self.assertEqual(len(res["facets"]["price"]), 2)
+
+        with self.subTest("Try getting number facet with overlapping ranges raises an error"):
```
farshidz: nit "try getting" seems redundant in these descriptions. I think "Number facet with overlapping ranges raises an error" would suffice here for example

tests/api_tests/v1/tests/api_tests/structured_index/test_hybrid_search.py
```diff
@@ -670,5 +670,194 @@ def test_hybrid_search_default_parameters(self):
...
+                                        "ranges": [
+                                            {"from": 0, "to": 50}
+                                        ]
+                                    }
+                                }
+                            }
+                        )
+                    self.assertIn("Ranges can only be used for 'number' facets", str(e.exception))
+
+        with self.subTest("Try getting facets with filter terms exclusions"):
```
farshidz: I feel like there are two many sub tests here. You could at least have negative cases in a separate test case/functions or even more granular. In fact, I think you could have a `test_facets.py` and a bunch of functions there. This module is approaching 1000 lines which is often not a good thing.

tests/integ_tests/tensor_search/integ_tests/test_hybrid_search.py
```diff
@@ -3328,3 +3330,322 @@ def test_lexical_query_provided_for_tensor_retrieval_tensor_ranking_raises_error
...
+    def test_single_facet(self):
+        """
+        Test that single facet requests work as expected.
+        """
+        def assert_dict_almost_equal(self, d1, d2, places=1):
+            """Compare two dictionaries with numeric values allowing for small differences."""
+            self.assertEqual(d1.keys(), d2.keys(), "Dictionaries have different keys")
+            for key in d1:
+                if isinstance(d1[key], dict):
+                    assert_dict_almost_equal(self, d1[key], d2[key], places)
```
farshidz: Comments I left on API tests re test organization largely apply here as well

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -86,11 +86,35 @@ public Result search(Query query, Execution execution) {
         logIfVerbose(String.format("Base Query is: "), verbose);
         logIfVerbose(query.toDetailString(), verbose);
 
-        // Validation for limit and rerank count
+        // Validation for limit
```
farshidz, line 89: This class has existing unit tests which you haven't updated for the new code

src/marqo/core/search/search_filter.py
danyilq: parenthness representation was redundant here. in the _get_filter_term they are already wrapped. It was breaking exclude_terms due to user passing `NOT field:something`, but it being represented as `NOT (field:something)`. It didn't break any existing tests so I assume it was not needed indeed

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
danyilq: I agree, will add tests for it.

farshidz:
Issues found in testing:
* Handling of missing fields, or no missing value for all matches. Currently returns 
```
 'facets': {'fied_name': {'': {'count': 3}}},
```
* Above response also when type is wrong. e.g., field is string, request says numeric (so no values of that type)
* Range overlap error not detected in this example, which returns empty facets
```
 'popularity': {
                                         'type': 'number',
                                         'ranges': [
                                             {
                                                    'from': 0,
                                                    'name': 'low',
                                                },
                                                {
                                                    'from': 14,
                                                    'name': 'high',
                                             }
                                         ]
                                     },
```
* `exclusionTerm` is intolerant of whitespace (at least). This fails
```
filter_string='age:[100   TO    1000]',
...
 'style': {'type': 'string',
       'maxResults': 10,
       'excludeTerms': ['age:[100 TO 1000]']
       },
```
* Handling of `exclusionTerm` for complex filter strings has a number of issues.
  * ~~`filter_string='age:[100 TO 1000] OR style:modern AND style:classic'` and ` 'excludeTerms': ['style:modern']` seems to use `filter_string='age:[100 TO 1000] AND style:classic'` for facets (i.e. always removing the operator before)~~ This was my misunderstanding
  * Handling of parentheses. In the example above, errors out if `filter_string='(age:[100 TO 1000] OR style:modern) AND style:classic'`
* Stepping through the code I can see you're using weakAnd. I think you should switch to OR but also this will require repeating perf tests. The problem is weakAnd will never get all matches, I think there's a default targetHits
```
select * from marqo__my_01index2 where weakAnd(default contains "nice", default contains "bag") AND ...
```
Is one example facets query I'm seeing. We ran into issues with score modifiers and switched to OR when there are score modifiers.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -621,6 +624,8 @@ def gather_documents_from_response(response: QueryResult, marqo_index: MarqoInde
     vespa_index = vespa_index_factory(marqo_index)
     hits = []
     for doc in response.hits:
+        if doc.id.startswith("group:facet:"):
```
farshidz: Is there anything stopping users from having this prefix in a doc ID?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -635,6 +640,152 @@ def gather_documents_from_response(response: QueryResult, marqo_index: MarqoInde
...
+                        continue
+                    if facets.fields[field_name].ranges is None:
+                        if any(value in [-9223372036854775808, 'NaN'] for value in processed_stats.values()):
+                            # Vespa's value for null
+                            continue
+                        # aggregate statistic between int and float
+                        facets_response[field_name] = _combine_number_stats(
+                            facets_response.get(field_name, {}), processed_stats
+                        )
+                    elif group_type != "null": # the case when there is no values in range
```
farshidz: What's the `else` here? If these are expected to be exhaustive, you can error out for `else` as an invariant violation (bug). This seems like it could fail silently

farshidz: i.e. what happens if `ranges is None and group_type == null and type not in ["string", "array"]`

src/marqo/tensor_search/tensor_search.py
```diff
@@ -635,6 +640,152 @@ def gather_documents_from_response(response: QueryResult, marqo_index: MarqoInde
...
+def _build_field_map(facets: FacetsParameters) -> Dict[str, Any]:
+    """Build a mapping from field names to their facet parameters."""
+    return {
+        facet_field[0]: facet_field[1]
+        for facet_field in facets.fields.items()
+    }
+
+def _extract_facet_field_name(field_id: str, facets: FacetsParameters) -> str:
+    """Extract the facet field name from a Vespa field ID."""
+    if "marqo__string_array_" in field_id:
```
farshidz: There's a lot of logic specific to index type here which makes me wonder if this logic should really be in `SemistructuredVesapIndex` class. Search logic in this module has so far been agnostic of index type.

farshidz: There's a constant `FIELD_STRING_ARRAY_PREFIX` for this

src/marqo/tensor_search/tensor_search.py
```diff
@@ -635,6 +640,152 @@ def gather_documents_from_response(response: QueryResult, marqo_index: MarqoInde
...
+        # strip group:marqo__string_array_ part of name
+        return field_id[30:]
+    if "marqo" not in field_id:
```
farshidz: This seems fragile. A user could have `marqo` in their field name. The only thing we don't allow in field names is the `marqo__` prefix

src/marqo/tensor_search/tensor_search.py
```diff
@@ -635,6 +640,152 @@ def gather_documents_from_response(response: QueryResult, marqo_index: MarqoInde
...
+        if "neg" in field_id:  # :neg(n) - when combining
```
farshidz: Also risky. User could have `neg` in field name

src/marqo/tensor_search/tensor_search.py
```diff
@@ -635,6 +640,152 @@ def gather_documents_from_response(response: QueryResult, marqo_index: MarqoInde
...
+    """Get the name for a range facet."""
+    if facet_range.name is not None:
+        if facet_range.to_ is not None:
+            if float(value_key) == facet_range.to_:
+                return facet_range.name
+        elif value_key == "Infinity":
+            return facet_range.name
+        return None
+
+    from_val = "-Infinity" if facet_range.from_ is None else str(facet_range.from_)
```
farshidz: I think `Inf` is more commonly used

src/marqo/tensor_search/tensor_search.py
danyilq: I'm gonna add comment here. This is relevant only to number fields which in custom searcher we set to `n` where `n` is id of facet.

src/marqo/tensor_search/tensor_search.py
danyilq: this is a custom id defined by us. neg is for integer fields. Floats and Ints are assigned same id, but one is with negative. Only relevant for 'number' fields

src/marqo/tensor_search/tensor_search.py
danyilq: this is what vespa returns. I am not opinionated, if you want it will be `Inf`

src/marqo/tensor_search/tensor_search.py
danyilq: it is not actually doc's id. It is vespa's content group id

src/marqo/core/search/search_filter.py
farshidz: Okay so you've definitely broken this. The reason nothing is failing is `str(node)` isn't actually used except by your code. Here's an example. Filter string `NOT (age:[100 TO 1000] AND style:modern)`
Your str method returns `NOT age:[100 TO 1000] AND style:modern`. Which really means `(NOT age:[100 TO 1000]) AND style:modern` which is not the same as `NOT (age:[100 TO 1000] AND style:modern)`

tests/unit_tests/marqo/core/search/test_facets_term.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+
+class TestFacetsTerm(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        # Create mock objects
+        cls.marqo_index = MagicMock(spec=SemiStructuredMarqoIndex)
+        cls.marqo_index.parsed_marqo_version.return_value = get_version()
+        cls.index = SemiStructuredVespaIndex(cls.marqo_index)
```
farshidz: With unit tests, you generally structure them just like the code. So tests for `SemiStructuredVespaIndex` will be in `unit_tests.marqo.core.semi_structured_vespa_index.semi_structured_vespa_index.py`

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
```diff
@@ -682,3 +811,153 @@ def _create_update_statement_for_updating_field(self, fields, key, val):
...
+                            continue
+                        if facets.fields[field_name].ranges is None:
+                            if any(value in [-9223372036854775808, 'NaN'] for value in processed_stats.values()):
+                                # Vespa's value for null
+                                continue
+                            # aggregate statistic between int and float
+                            facets_response[field_name] = self._combine_number_stats(
+                                facets_response.get(field_name, {}), processed_stats
+                            )
+                        elif group_type != "null": # the case when there is no values in range
```
farshidz, line 873: What's the `else` here? If these are expected to be exhaustive, you can error out for `else` as an invariant violation (bug). This seems like it could fail silently 

i.e. what happens if `ranges is None and group_type == null and type not in ["string", "array"]`

src/marqo/tensor_search/tensor_search.py
farshidz: Moved comments to new location

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
```diff
@@ -682,3 +811,153 @@ def _create_update_statement_for_updating_field(self, fields, key, val):
...
+            for facet_field in facets.fields.items()
+        }
+
+    def _extract_facet_field_name(self, field_id: str, facets: FacetsParameters) -> str:
+        """Extract the facet field name from a Vespa field ID."""
+        if SemiStructuredVespaSchema.FIELD_STRING_ARRAY_PREFIX in field_id:
+            # strip group:marqo__string_array_ part of name
+            return field_id[30:]
+        # this is only possible if field query was a number without ranges. So group name is n for float and negative n for int
+        if "marqo" not in field_id:
```
farshidz: This seems fragile. A user could have `marqo` in their field name. The only thing we don't allow in field names is the `marqo__` prefix.

I'm gonna add comment here. This is relevant only to number fields which in custom searcher we set to `n` where `n` is id of facet.

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
```diff
@@ -682,3 +811,153 @@ def _create_update_statement_for_updating_field(self, fields, key, val):
...
+            if "neg" in field_id:  # :neg(n) - when combining
```
farshidz: Also risky. User could have `neg` in field name.

this is a custom id defined by us. neg is for integer fields. Floats and Ints are assigned same id, but one is with negative. Only relevant for 'number' fields

danyilq: It is impossible for a field that includes user's field name to reach here. "marqo" is always present in any field, at least in name of an attriubute. This code's part execution is ONLY possible if it was a number field without specified range. So neg(n) is defined by us only.

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
danyilq: and it is okay if they do, it doesn't interfere with logic anyhow, logic relies on "marqo" not being present.

src/marqo/tensor_search/models/api_models.py
```diff
@@ -172,6 +176,63 @@ def validate_search_method(cls, value):
...
+                        term = term[1:]
+                    elif term.count(')') > term.count('('):
+                        term = term[:-1]
+                # Append actual value
+                filter_str_terms.append(term.strip())
+
+            for facet_field in facets.fields.items():
+                field_name, field_parameters = facet_field
+                if field_parameters.exclude_terms:
+                    missing_exclusions = [ex for ex in field_parameters.exclude_terms
```
farshidz: Just noting the time complexity on this is really bad. It's `number_of_exclusion_terms * length of filter string`. Let's sort it out soon after this PR

src/marqo/core/search/hybrid_search.py
```diff
@@ -125,6 +129,22 @@ def search(
...
+        if facets is not None and not isinstance(marqo_index, SemiStructuredMarqoIndex):
+            raise core_exceptions.UnsupportedFeatureError(
+                f"Facets are only supported for unstructured indexes"
+            )
+        if track_total_hits is not None and not isinstance(marqo_index, SemiStructuredMarqoIndex):
+            raise core_exceptions.UnsupportedFeatureError(
+                f"trackTotalHits is only supported for unstructured indexes"
+            )
+        if isinstance(marqo_index, SemiStructuredMarqoIndex):
+            string_array_field_names = [string_array_field.name for string_array_field in marqo_index.string_array_fields]
```
farshidz: We have a pattern in `marqo_index.py` where we pre-compute and cache these things. In this case I think there's an existing map `SemiStructuredMarqoIndex.name_to_string_array_field_map` and you can use the keyset. 

Regardless, you shouldn't be using a list here but a set (the map gives you a key set anyway, but just pointing out the poor choice of data structure that is consequential).

src/marqo/core/search/hybrid_search.py
```diff
@@ -125,6 +129,22 @@ def search(
...
+            for facet_field_name, facet_field_params in facets.fields.items():
+                if facet_field_params.type == "array" and facet_field_name not in string_array_field_names:
+                    raise core_exceptions.InvalidArgumentError(
```
farshidz: nit convention is to not have a trailing period in exception messages.

src/marqo/tensor_search/models/api_models.py
```diff
@@ -207,7 +207,7 @@ def validate_facet_exclude_terms_in_filter(cls, values):
                 # Handle range queries and clean up any remaining spaces
                 term = term.strip()
                 while term.startswith('(') or term.count('(') != term.count(')'):
```
farshidz: I know we're leaving this for now, but just keep in mind count operations are linear in string length. If you look at the filter string parser, we don't do any count, find etc partly for this reason. Generally you want to parse with one pass (or close to that).

src/marqo/tensor_search/models/api_models.py
```diff
@@ -207,7 +207,7 @@ def validate_facet_exclude_terms_in_filter(cls, values):
...
-                    if term.count('(') > term.count(')'):
+                    if term.count('(') > term.count(')') or term.startswith('('):
                         term = term[1:]
                     elif term.count(')') > term.count('('):
```
farshidz: You have a while loop looking at term, `if...elif` but no `else`. If neither of the two match, you'll loop indefinitely because `term` remains unchanged. If `else` isn't expected to happen, you should either make the second `elif` an `else`, or have `else: error out`.

farshidz (changes requested):

src/marqo/tensor_search/models/api_models.py
farshidz: Looking again the second one is clearly your `else`

src/marqo/core/search/hybrid_search.py
```diff
@@ -137,10 +137,9 @@ def search(
             raise core_exceptions.UnsupportedFeatureError(
                 f"trackTotalHits is only supported for unstructured indexes"
             )
-        if isinstance(marqo_index, SemiStructuredMarqoIndex):
-            string_array_field_names = [string_array_field.name for string_array_field in marqo_index.string_array_fields]
+        if isinstance(marqo_index, SemiStructuredMarqoIndex) and marqo_index.name_to_string_array_field_map and facets is not None:
```
farshidz: Are we erroring out when array field doesn't exist?

I think we agreed to behave like other fields and not error out? But also it only errors out if `marqo_index.name_to_string_array_field_map` is non-empty. So if there aren't any array fields and you specify an array facet that doesn't exist no error. The behavior seems quite inconsistent

farshidz (approved):

PR #1169, opened 2025-03-31
Fix backwards Compatibility Tests

wanliAlex:
bug fix

  - We add the suffix `-cloud` when we pull Docker images. From 2.17.0, marqo's docker image is tagged with a `-cloud` suffix.
  - Change the log level from debug to info.

* **Other information**:


tests/compatibility_tests/docker_manager.py
```diff
@@ -498,17 +498,30 @@ def pull_image_from_dockerhub(self, image_name: str):
...
-        try:
-            self.logger.debug(f"Pulling image: {image_name}")
-            self.docker_client.images.pull(image_name)
-            self.logger.info(f"Successfully pulled image: {image_name}")
-        except ImageNotFound:
-            raise RuntimeError(f"Image not found: {image_name}")
-        except APIError as e:
-            raise RuntimeError(f"Failed to pull image {image_name}: {str(e)}")
+
+        variants = [
```
farshidz, line 523: What happens if there is `x.y.z` as well as `x.y.z-cloud`? I think we should use `-cloud` only if the main tag doesn't exist. Perhaps log a warning also if both exist. This seems to prefer `-cloud` over just `x.y.z`

farshidz (changes requested):

tests/compatibility_tests/docker_manager.py
wanliAlex: The main tag is pulled first because it is the first variant

wanliAlex: Add check and log warning if there are multiple variants.

farshidz (approved):

PR #1174, opened 2025-04-07
[releases/2.16]Fix `version_uuid` backwards compatibility issue

wanliAlex:
## Change Summary
Fix a bug that Marqo incorrectly sends the `marqo__version_uuid` field to Vespa, even if the index does not support this field. This field was added in 2.16.0. Only indexes created after 2.16.0 have this field. However, Marqo is sending this field to vespa for indexes created before 2.16.0

## Related Jira Ticket
[MOSD-372](https://s2search.atlassian.net/browse/MOSD-372)

## Checklist

### For new field types:

[MOSD-372]: https://s2search.atlassian.net/browse/MOSD-372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/marqo/core/document/document.py
```diff
@@ -117,7 +117,7 @@ def partial_update_documents(self, partial_documents: List[Dict], marqo_index) \
         elif marqo_index.type is IndexType.Structured:
             pass
         elif marqo_index.type is IndexType.SemiStructured:
-            if marqo_index.parsed_marqo_version() < SEMISTRUCTURED_INDEX_PARTIAL_UPDATE_SUPPORT_VERSION: # Partial updates for semi-structured indexes are only supported for Marqo version >= 2.16.0
+            if not marqo_index.index_supports_partial_updates: # Partial updates for semi-structured indexes are only supported for Marqo version >= 2.16.0
                 raise UnsupportedFeatureError("Partial document update is not supported for this index version. "
```
farshidz: Add min required version here. I think it's fine to hard-code in the message as this will never change.

src/marqo/core/models/marqo_index.py
```diff
@@ -611,6 +611,13 @@ def generate():
 
         return self._cache_or_get('tensor_subfield_map', generate)
 
+    @property
+    def index_supports_partial_updates(self) -> bool:
+        """
+        Check if the index supports partial updates.
+        """
+        return self.parsed_marqo_version() >= semver.VersionInfo.parse("2.16.0")
```
farshidz: Add a class variable so you parse this only once

farshidz (changes requested):

src/marqo/core/document/document.py
wanliAlex: done

src/marqo/core/models/marqo_index.py
wanliAlex: done

farshidz (approved):

PR #1175, opened 2025-04-07
Fix backwards compatibility add docs tests

farshidz:
## Change Summary
Add docs tests don't actually call add docs.

This PR fixes add docs tests that run for >=2.0.0. It also improves logging and addresses an issue where test runs are started even if image build fails due to network etc. (all runs will fail eventually as there's no image to test).

## Checklist

- N/A Tests have been added for changes
- N/A Documentation has been updated

### For new field types:
- N/A Tests cover score modifier usage of this new type
- N/A Test indexes updated to cover the new type for all APIs (add docs, search, partial update, etc.)


wanliAlex (dismissed):
LGTM

wanliAlex (approved):

PR #1179, opened 2025-04-09
Fix infer modality

wanliAlex:
## Change Summary

When inferring the modality of a URL, we now strip query parameters before performing the inference.
This improves modality detection accuracy by focusing on the actual file extension.

Potential **breaking change**:

- Previously, if a file extension appeared only in the query parameters, Marqo would fall back to MIME type inference.

- Now, Marqo uses the URL extension (after removing the query) directly, without considering the MIME type if the extension is missing.

## Checklist

- [x] Documentation has been updated

### For new field types:


src/marqo/core/inference/modality_utils.py
```diff
@@ -34,56 +35,90 @@ def fetch_content_sample(url: str, media_download_headers: Optional[dict] = None
...
+    """
+    Infer the modality based on the file extension. Is it is not a known extension, return None.
+
+    Args:
+        extension: A string representing the file extension (e.g., 'jpg', 'mp4', etc.)
+
+    Returns:
+        Modality: The inferred modality (IMAGE, VIDEO, AUDIO, or None if unknown)
+    """
+    if extension in ['jpg', 'jpeg', 'png', 'gif', 'webp']:
```
vicilliar, line 53: what about capitals? Maybe we should use `extension.lower()`

src/marqo/core/inference/modality_utils.py
```diff
@@ -120,3 +155,22 @@ def validate_url(url: str) -> bool:
         return validators.url(url) or validators.url(encode_url(url))
     else:
         return False
+
+
+def get_url_file_extension(url: str) -> Optional[str]:
+    """Get the file extension from a URL.
+
+    Note that we should clean the url before passing it to this function, e.g., remove query parameters.
```
vicilliar: Why do we need to remove query parameters if getting `parsed_url.path` already excludes query parameters?

src/marqo/tensor_search/models/preprocessors_model.py
```diff
@@ -1,7 +1,7 @@
 from typing import Optional, Any
 
 from marqo.base_model import MarqoBaseModel
-from marqo.s2_inference.multimodal_model_load import Modality
+from marqo.core.inference.modality_utils import Modality
```
vicilliar: Shouldn't we be importing from the source? Which is:
```
marqo.core.inference.api
```
based on the import in `marqo.core.inference.modality_utils`

tests/unit_tests/marqo/core/inference/test_infer_modality.py
```diff
@@ -88,3 +89,29 @@ def test_infer_modality_list_of_strings(self):
...
+            ("https://example.com/file.unknown?param=test", None, "Unknown extension should return None"),
+            ("https://example.com/no_extension?query=data", None, "URL with no extension should return None"),
+            ("https://example.com/imagejpg?query=string", None,
+             "URL with incorrect extension format (missing dot) should return None"),
+        ]
+
+        for url, expected_modality, message in test_cases:
+            with self.subTest(msg=message, url=url):
+                inferred_modality = _infer_modality_based_on_extension(get_url_file_extension(url))
+                self.assertEqual(expected_modality, inferred_modality)
```
vicilliar: Can we test these codepaths:
1. `infer_modality` where `content` is a `str`, modality is correctly inferred from extension
2. `infer_modality` where `content` is a `str`, no proper extension so we must fetch content sample and use `_infer_modality_based_on_mime_type`
3. `infer_modality` where `content` is `bytes` so we use `_infer_modality_based_on_mime_type`

vicilliar (changes requested):

src/marqo/core/inference/modality_utils.py
wanliAlex: Make sense. Added extension.lower()

src/marqo/core/inference/modality_utils.py
wanliAlex: Update the doc strings

src/marqo/tensor_search/models/preprocessors_model.py
wanliAlex: done

tests/unit_tests/marqo/core/inference/test_infer_modality.py
wanliAlex: added tests

tests/integ_tests/tensor_search/integ_tests/test_search_combined.py
```diff
@@ -1046,12 +1046,12 @@ def test_search_private_image_return_proper_error(self):
...
         test_queries_list = [
-            ("https://a-dummy-image-url.jpg", "A invalid image"),
-            ({"https://a-dummy-image-url.jpg": 1, "test": 1},
+            ("https://a-dummy-image-url.com/image1.jpg", "A invalid image"),
+            ({"https://a-dummy-image-url.com/image2.jpg": 1, "test": 1},
              "A invalid image in the dictionary")
         ]
 
-        for index_name in [self.structured_default_image_index, self.unstructured_default_image_index]:
+        for index_name in [self.unstructured_default_image_index]:
```
vicilliar: why aren't we testing structured here?

vicilliar (approved):

PR #1190, opened 2025-04-22
Add the `POST indexes/{index_name}/documents/get-batch` API to retrieve batch documents

wanliAlex:
## Change Summary

Add the `POST indexes/{index_name}/documents/get-batch` API to retrieve multiple documents by IDs. 

## Checklist

### For new field types:

### DIFF
--- a/src/marqo/api/models/get_batch_documents_request.py
+++ b/src/marqo/api/models/get_batch_documents_request.py
@@ -1,0 +1,10 @@
+from pydantic import Field, conlist
+
+from marqo.base_model import MarqoBaseModel
+
+
+class GetBatchDocumentsRequest(MarqoBaseModel):
+    """
+    A request model for getting batch documents by their ids from Marqo.
+    """
+    document_ids: conlist(str, min_items=1) = Field(alias="documentIds")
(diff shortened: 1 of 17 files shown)


src/marqo/api/models/get_batch_documents_request.py
```diff
@@ -0,0 +1,10 @@
+from pydantic import Field, conlist
+
+from marqo.base_model import MarqoBaseModel
+
+
+class GetBatchDocumentsRequest(MarqoBaseModel):
+    """
+    A request model for getting batch documents by their ids from Marqo.
+    """
+    document_ids: conlist(str, min_items=1) = Field(alias="documentIds")
```
farshidz, line 10: Should we permit empty list and return empty? What does ES do?

wanliAlex: We have this check [here](https://github.com/marqo-ai/marqo/blob/09b34350ad32c5e9579e09b5f71e5e3319ce0a96/src/marqo/tensor_search/tensor_search.py#L165-L166). So I added the `min_items=1` for an earlier stop.

wanliAlex: But for ES, they return an empty list.

farshidz (approved):

PR #1191, opened 2025-04-22
Upgrade pydantic to v2 (PR1)

papa99do:
## Change Summary
Our perf test identified pydantic validation logic to be one of the major perf bottleneck of the search response parsing process. the core logic of Pydantic v2 is rewritten in rust and is much more performant than Pydantic v1. Upgrading to pydantic v2 helps address this perf bottleneck. In addition, after upgrading Pydantic, we can also upgrade Starlette and FastAPI to address some security issues.

Instead of a big bang change which touches all the Pydantic model class in Marqo repo. We will take baby steps and gradually migrate the model classes to v2. This is less risky. Please see [This guide](https://medium.com/codex/migrating-to-pydantic-v2-5a4b864621c3) for more details. 

In this PR
* Upgrade Pydantic dep to the latest v2 version
* Upgrade FastAPI, Starlette, Uvicorn dep to the latest version
* Replace all dep to pydantic package to pydantic.v1 so we still use the embedded v1 implementation in Pydantic v2 package. 
* Replace API method parameter with Pydantic model type with dictionary and handle the conversion from dictionary to Pydantic v1 model manually. This is due to that the latest version of FastAPI/Starllete uses Pydantic v2 and do not take v1 models as API method parameter. 
* * Please check `api.py` file and `test_api.py` file specifically for this change.
* * Please check 

In a following PR, we will
* Upgrade QueryResult to use Pydantic v2
* Upgrade SemiStructuredVespaDocument to use Pydantic v2

## Related Jira Ticket
[MOSD-365](https://s2search.atlassian.net/browse/MOSD-365)

## Checklist

### For new field types:

[MOSD-365]: https://s2search.atlassian.net/browse/MOSD-365?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/run_required_checks.yml
+++ b/.github/workflows/run_required_checks.yml
@@ -24,10 +24,17 @@ jobs:
       - name: Check for Non-Markdown Changes
         id: check_changes
         run: |
-          if git diff --name-only ${{ github.event.before }} ${{ github.sha }} | grep -vE '\.md$'; then
-            echo "non_documentation_changes_found=true" >> $GITHUB_OUTPUT
+          BEFORE="${{ github.event.before }}"
+          AFTER="${{ github.sha }}"
+          
+          if [ -z "$BEFORE" ] || \
+             git diff --name-only "$BEFORE" "$AFTER" | grep -qvE '\.md$'
+          then
+            # either BEFORE was empty, or we found at least one non‑.md change
+            echo "non_documentation_changes_found=true" >> "$GITHUB_OUTPUT"
           else
-            echo "non_documentation_changes_found=false" >> $GITHUB_OUTPUT
+            # BEFORE was set and all changed files end in .md
+            echo "non_documentation_changes_found=false" >> "$GITHUB_OUTPUT"
           fi
 
   Run-integration-tests:
--- a/src/marqo/api/models/add_docs_objects.py
+++ b/src/marqo/api/models/add_docs_objects.py
@@ -1,8 +1,8 @@
 from typing import List, Dict
 from typing import Optional, Any, Sequence
 
-from pydantic import BaseModel, root_validator
-from pydantic import Field
+from pydantic.v1 import BaseModel, root_validator
+from pydantic.v1 import Field
 
 from marqo.tensor_search.enums import EnvVars
 from marqo.tensor_search.models.private_models import ModelAuth
@@ -25,7 +25,7 @@ class AddDocsBodyParams(BaseModel):
     mappings: Optional[dict] = None
     documents: Sequence[Dict[str, Any]]
     imageDownloadThreadCount: int = Field(default_factory=lambda: read_env_vars_and_defaults_ints(EnvVars.MARQO_IMAGE_DOWNLOAD_THREAD_COUNT_PER_REQUEST))
-    mediaDownloadThreadCount: Optional[int]
+    mediaDownloadThreadCount: Optional[int] = None
     textChunkPrefix: Optional[str] = None
 
     @root_validator
--- a/src/marqo/core/inference/api/preprocessing_config.py
+++ b/src/marqo/core/inference/api/preprocessing_config.py
@@ -2,7 +2,7 @@ from abc import ABC, abstractmethod
 from typing import Optional, Dict, Literal, List, Set, Union
 
 import pydantic
-from pydantic import root_validator
+from pydantic.v1 import root_validator
 
 from marqo.base_model import ImmutableBaseModel
 from marqo.core.inference.api.modality import Modality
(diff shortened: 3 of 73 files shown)


.github/workflows/run_required_checks.yml
```diff
@@ -24,10 +24,17 @@ jobs:
       - name: Check for Non-Markdown Changes
         id: check_changes
         run: |
-          if git diff --name-only ${{ github.event.before }} ${{ github.sha }} | grep -vE '\.md$'; then
-            echo "non_documentation_changes_found=true" >> $GITHUB_OUTPUT
+          BEFORE="${{ github.event.before }}"
```
papa99do, line 27: This is to make sure when `github.event.before` is not present, the build will still run integration tests. This happened in one of my PR commit.

src/marqo/api/models/add_docs_objects.py
```diff
@@ -25,7 +25,7 @@ class Config:
     mappings: Optional[dict] = None
     documents: Sequence[Dict[str, Any]]
     imageDownloadThreadCount: int = Field(default_factory=lambda: read_env_vars_and_defaults_ints(EnvVars.MARQO_IMAGE_DOWNLOAD_THREAD_COUNT_PER_REQUEST))
-    mediaDownloadThreadCount: Optional[int]
+    mediaDownloadThreadCount: Optional[int] = None
```
farshidz, line 28: Why this change?

src/marqo/core/inference/api/preprocessing_config.py
```diff
@@ -11,12 +11,12 @@
...
     """Parent class of preprocessing config for all modality types"""
     modality: str
-    should_chunk: bool = pydantic.Field(default=False, alias='shouldChunk')
+    should_chunk: bool = pydantic.v1.Field(default=False, alias='shouldChunk')
 
 
 class ChunkConfig(ImmutableBaseModel):
-    split_length: int = pydantic.Field(gt=0, alias='splitLength')
-    split_overlap: int = pydantic.Field(ge=0, alias='splitOverlap')
+    split_length: int = pydantic.v1.Field(gt=0, alias='splitLength')
```
farshidz: Couldn't we have done `import pydantic.v1 as pydantic` for all of these?

src/marqo/vespa/models/query_result.py
```diff
@@ -1,23 +1,23 @@
 from typing import List, Dict, Any, Optional
 
-from pydantic import BaseModel, Field
+from pydantic.v1 import BaseModel, Field
 
 
 # See https://docs.vespa.ai/en/reference/default-result-format.html
 class RootFields(BaseModel):
-    total_count: Optional[int] = Field(alias='totalCount')
+    total_count: Optional[int] = Field(None, alias='totalCount')
```
farshidz, line 8: Why do we need all these default `None`s now?

tests/integ_tests/tensor_search/test_prefix.py
```diff
@@ -213,19 +210,13 @@ def test_prefix_text_chunks_e5(self):
 
                 embed_res_document_prefix = embed(
                     marqo_config=self.config, index_name=index.name,
-                    embedding_request=EmbedRequest(
-                        content=["hello"],
-                        content_type="document"
-                    ),
+                    embedding_request_dict={"content": ["hello"], "content_type": "document"},
```
farshidz, line 213: How much do we gain from this change?

papa99do: we don't gain from this change. We can no longer inject pydantic v1 model as the parameter of an API method since fastAPI now use pydantic v2. We'll get an error mixing v1 and v2 usage.

src/marqo/api/models/add_docs_objects.py
papa99do: This might be an accidental change when I'm trying the migration. Pydantic v2 requires `None` to be specified as a default value of an Optional field. It does no harm to pydantic v1 model to specify a default as well. Since we will need to add it eventually, can we keep it in this PR?

src/marqo/vespa/models/query_result.py
papa99do: This might be an accidental change when I'm trying the migration. Pydantic v2 requires `None` to be specified as a default value of an Optional field. It does no harm to pydantic v1 model to specify a default as well. Since we will need to add it eventually, can we keep it in this PR?

src/marqo/core/inference/api/preprocessing_config.py
papa99do: good point. Will change this.

farshidz (approved):

PR #1194, opened 2025-04-23
exclude int/float values when they are not present due to a race condition

danyilq:
### Main change
This PR updates `to_marqo_document` function to check type of `int`/`float` fields before updating response dict. If type is a `list` it means one or more of the fields has no value, which happens due to a race condition when adding and updating the same document. New behaviour will exclude fields without values from response rather than erroring out.
During testing I wasn't able to achieve the same behaviour for `array` fields, it seemed to be only reproducible for `int` and `float` fields.
### Side change
- removed `logger.error` print in custom route handler due to logs being printed out twice.

## Checklist

### For new field types:


src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
wanliAlex, line 1: Add some unit tests on this?

wanliAlex (changes requested):

src/marqo/api/route.py
```diff
@@ -19,7 +19,6 @@ async def marqo_custom_route_handler(request: Request):
             try:
                 return await original_route_handler(request)
             except Exception as exc:
-                logger.error(str(exc), exc_info=True)
```
farshidz: Why are you removing this line? Pretty sure it's needed to log the stack trace

src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
```diff
@@ -394,8 +394,22 @@ def to_marqo_document(self, marqo_index: SemiStructuredMarqoIndex) -> Dict[str,
         # Add int and float fields back
         # Please note that int-map and float-map fields are flattened in the result. The correct behaviour is to convert
         # them back to the format when they are indexed. We will keep the behaviour as is to avoid breaking changes.
-        marqo_document.update(self.fixed_fields.int_fields)
-        marqo_document.update(self.fixed_fields.float_fields)
+        # Due to a race condition, the int_fields and float_fields can be a list of dicts, we want to format them
+        # and exclude fields with empty values
+        if isinstance(self.fixed_fields.int_fields, list):
```
farshidz: Definitely need an integ test that actually calls Vespa but modifies the response through mocking

farshidz (changes requested):

src/marqo/api/route.py
danyilq: > - removed `logger.error` print in custom route handler due to logs being printed out twice.
We actually don't uvicorn already prints out error, this just makes it double print.

wanliAlex (dismissed):
LGTM

tests/integ_tests/tensor_search/integ_tests/test_search_unstructured.py
```diff
@@ -1564,3 +1564,33 @@ def test_special_characters_in_map_score_modifiers(self):
...
+                index_name=self.default_text_index, docs=docs, tensor_fields=["text_field_1"]
+            )
+        )
+        with mock.patch.object(self.config.vespa_client, "query") as mock_query:
+        # mock the VespaClient.query method to return real response and modify the response
+            def wrapper(*args, **kwargs):
+                # Call the real method
+                response = real_query(*args, **kwargs)
+                # Modify the response — e.g., remove a field
+                del response.hits[0].fields['marqo__int_fields']
```
farshidz: This isn't the race condition response, is it? You're deleting the whole dict

src/marqo/api/route.py
farshidz: Are you sure about this? @wanliAlex implemented this. Let's make sure he checks this out as well if he hasn't

farshidz (changes requested):

tests/integ_tests/tensor_search/integ_tests/test_search_semi_structured.py
```diff
@@ -1503,4 +1503,34 @@ def test_special_characters_in_map_score_modifiers(self):
...
+                index_name=self.default_text_index.name, docs=docs, tensor_fields=["text_field_1"]
+            )
+        )
+        with mock.patch.object(self.config.vespa_client, "query") as mock_query:
+        # mock the VespaClient.query method to return real response and modify the response
+            def wrapper(*args, **kwargs):
+                # Call the real method
+                response = real_query(*args, **kwargs)
+                # Modify the response — e.g., remove a field
+                response.hits[0].fields['marqo__int_fields'] = [{"key": "int_field_1"}]
```
farshidz: This is more of an edge case. There could be a mix of valid and invalid entries

farshidz (changes requested):

farshidz (dismissed):

farshidz (approved):

PR #1196, opened 2025-04-24
Pydantic V2 upgrade PR2

papa99do:
## Change Summary
The following improvements are done to improve the performance of Vespa search result parsing
* QueryResult is upgraded to PydanticV2
* SemiStructuredDocument is upgraded to PydanticV2
* fields in SemiStructuredDocument are rearranged so we can construct most static fields from vespa doc in one go
* string array field handling logic is simplified
* field map properties in SemiStructuredMarqoIndex are evaluated per document instead of per field

Now that we are using pydantic v2 model, we will need to properly handle ValidationError in pydantic v2.
* Add an exception handler to handle pydantic.ValidationError error

## Checklist

### For new field types:


src/marqo/base_model.py
```diff
@@ -1,3 +1,5 @@
+import pydantic
+from pydantic import ConfigDict
 from pydantic.v1 import BaseModel
```
farshidz, line 3: Shouldn't all v1 imports be removed now?

src/marqo/core/models/marqo_index.py
```diff
@@ -624,7 +624,9 @@ def index_supports_partial_updates(self) -> bool:
         """
         Check if the index supports partial updates.
         """
-        return self.parsed_marqo_version() >= self._PARTIAL_UPDATE_SUPPORTED_VERSION
+        return self._cache_or_get(
```
farshidz, line 628: Was this check actually slow? Curious

farshidz: Also feels like these versions checks are all over the place and we need a uniform pattern for them.

src/marqo/core/models/marqo_index.py
```diff
@@ -624,7 +624,9 @@ def index_supports_partial_updates(self) -> bool:
...
+            'index_supports_partial_updates',
+            lambda: self.parsed_marqo_version() >= self._PARTIAL_UPDATE_SUPPORTED_VERSION)
```
farshidz, line 630: I believe you said you found this cache was slow. What change did you make to address that?

src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
```diff
@@ -2,39 +2,50 @@
...
     score_modifiers_fields: Dict[str, Any] = Field(default_factory=dict, alias=common.SCORE_MODIFIERS)
     vespa_multimodal_params: Dict[str, str] = Field(default_factory=dict, alias=common.VESPA_DOC_MULTIMODAL_PARAMS)
+
+    # metadata fields
+    version_uuid: Optional[str] = Field(default=None, alias=common.VESPA_DOC_VERSION_UUID)
     field_types: Dict[str, str] = Field(default_factory=dict, alias=common.VESPA_DOC_FIELD_TYPES)
+    vector_counts: int = Field(default=0, alias=common.FIELD_VECTOR_COUNT)
+
+    # Only in search result
+    match_features: Dict[str, Any] = Field(default_factory=dict, alias=common.VESPA_DOC_MATCH_FEATURES)
```
farshidz, line 43: Are you moving these so it's closer to the Vespa response?

src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
```diff
@@ -43,119 +54,54 @@ class SemiStructuredVespaDocument(MarqoBaseModel):
...
         text_fields = {}
         string_arrays_dict = {}
-        string_arrays_list = []
 
-        def process_field(field_name: str, fields: Dict) -> None:
-            """Helper function to process individual tensor fields, lexical fields and populate the appropriate dictionaries"""
-            if field_name in marqo_index.tensor_subfield_map:
-                tensor_fields[field_name] = fields[field_name]
-            elif field_name in marqo_index.lexical_field_map:
+        tensor_subfield_map = marqo_index.tensor_subfield_map
```
farshidz, line 71: Assuming `marqo_index.tensor_subfield_map` is slow, have you instead tried replacing the cache map with actual variables in the MarqoIndex class? We're still paying penalty once per doc here

src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
```diff
@@ -43,119 +54,54 @@ class SemiStructuredVespaDocument(MarqoBaseModel):
...
-                # Here we will just take the list defined under marqo__string_array and put under string_arrays_list, which will later be used  to construct the SemiStructuredVespaDocumentFields object.
-                # The SemiStructuredVespaDocumentFields object will later be used to construct SemiStructuredVespaDocument object.
-                # Later when converting SemiStructuredVespaDocument object to Marqo document (which is a dictionary of fields to be returned to the user), we will process this list & convert it
-                # to a dictionary by splitting the list element by '::', making the prefix before '::' as the key and suffix after '::' as the value.
-                if field_name == common.STRING_ARRAY:
-                    string_arrays_list = fields[field_name]
-            fixed_fields = SemiStructuredVespaDocumentFields.construct(
-                marqo__id=cls.extract_field(fields, common.VESPA_FIELD_ID, None),
-                short_string_fields=cls.extract_field(fields, common.SHORT_STRINGS_FIELDS, dict()),
```
farshidz, line 136: Where is all this logic now?

src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
```diff
@@ -43,119 +54,54 @@ class SemiStructuredVespaDocument(MarqoBaseModel):
...
-                string_arrays=string_arrays_list,
```
farshidz, line 137: Have you verified there is backwards compatibility coverage for these changes?

tests/integ_tests/tensor_search/test_api.py
```diff
@@ -609,4 +613,37 @@ class PydanticV1Model(pydantic.v1.BaseModel):
 
         self.assertIn('field required', str(context.exception.errors()))
 
+    def test_handle_pydantic_v1_validation_errors(self):
```
farshidz, line 616: Can't this be a unit test?

tests/unit_tests/marqo/core/semi_structured_vespa_index/test_semi_structured_vespa_index.py
```diff
@@ -0,0 +1,316 @@
...
+        ))
+
+        self.vespa_index_prior_to_216 = SemiStructuredVespaIndex(self.semi_structured_marqo_index(
+            name='index1',
+            lexical_field_names=['title', 'desc'],
+            tensor_field_names=['title'],
+            marqo_version='2.15.0'  # no individual string arrays
+        ))
+
+    def test_get_filter_string_escaped_characters(self):
```
farshidz, line 24: This test seems outside the scope of this PR? Not an issue, but making sure I'm not misunderstanding

tests/unit_tests/marqo/core/semi_structured_vespa_index/test_semi_structured_vespa_index.py
```diff
@@ -0,0 +1,316 @@
...
+            "markets": ["fr", "au"],
+            "title": "some product",
+            "desc": "some awesome product",
+            "marqo__tensors": {"title": {"chunks": ["some product"], "embeddings": [[1.0, 2.0]]}},
+            'multimodal_params': {'multi_modal': {'type': 'multimodal_combination',
+                                                  'weights': {'desc': 0.5, 'title': 1.0}}},
+        }, marqo_doc)
+
+    def test_vespa_to_marqo_conversion_should_handle_combined_string_array_fields(self):
+        vespa_doc = {
```
farshidz, line 186: Great idea to cover this here

src/marqo/base_model.py
papa99do: No, most of the model classes are still using v1 version.

src/marqo/core/models/marqo_index.py
papa99do: It'll be faster if we cache it. yes, it's a bit slow to do for each field.

src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
papa99do: Yes, so we just need to do the field population from `fields` dict once.

src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
papa99do: I can try that. It requires more code change, will probably save us little in latency. It cuts the latency by 1ms when we changed from 1 call per field to 1 call per doc (about 30 fields are returned in the perf test). So from 1800(30*60) calls to 60 call, we save 1ms. I don't think it will save too much reducing it to 1 call.

src/marqo/core/models/marqo_index.py
papa99do: https://github.com/marqo-ai/marqo/pull/1196/files#diff-982c19ac8c788cd7737ff4acf0437abfcc395596b89ac545f029704e36a954f6R71

tests/integ_tests/tensor_search/test_api.py
papa99do: I don't think so. It's testing the exception handler we configured for FastAPI app

tests/unit_tests/marqo/core/semi_structured_vespa_index/test_semi_structured_vespa_index.py
papa99do: I didn't change this test case, just renamed the test file name. It showed up as I added this test case.

src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
papa99do: Yes, in the unit test I added. I've covered the logic for index prior to 2.17

src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
papa99do: Still in this class. I've removed the duplicate logic, just to treat the `string_arrays` fields differently for index prior to 2.17 and index created post 2.17.

src/marqo/core/models/marqo_index.py
papa99do: When we upgrade this model to v2, this custom cache can be replaced with the built-in @cache_property

papa99do: will add a TODO for future reference

src/marqo/core/semi_structured_vespa_index/semi_structured_document.py
papa99do: will verify that we have the test case in compatibility test: 
* create an index with 2.15, and try to search /get_doc in 2.19

papa99do: Verified that in the search test case, we have a tags field which is string array field. (for both structured and unstructured). https://github.com/marqo-ai/marqo/blob/3f18aa0c335b302a70133ea74a8e8eedd8af3129/tests/compatibility_tests/search/test_search.py#L63. 

So in the prepare phase, these docs will be populated with the old version. and in test phase, they will be retrieved using the new Marqo version. We just to make sure 2.19 runs compatibility test with 2.15 version

src/marqo/core/models/marqo_index.py
papa99do: Added the TODO

farshidz (approved):

PR #1203, opened 2025-04-29
Infer modality based on field type for structured index

papa99do:
## Change Summary
* Infer modality based on the field type for structured index
* Add more telemetry data point for add_document process

## Checklist

### For new field types:

### DIFF
--- a/src/marqo/core/structured_vespa_index/structured_add_document_handler.py
+++ b/src/marqo/core/structured_vespa_index/structured_add_document_handler.py
@@ -27,6 +27,13 @@ MODALITY_FIELD_TYPE_MAP = {
     Modality.AUDIO: FieldType.AudioPointer,
 }
 
+FIELD_TYPE_MODALITY_MAP = {
+    FieldType.Text: Modality.TEXT,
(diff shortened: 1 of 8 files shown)


src/marqo/core/structured_vespa_index/structured_add_document_handler.py
```diff
@@ -74,20 +81,22 @@ def _handle_field(self, marqo_doc, field_name, field_content):
         marqo_doc[field_name] = content
 
     def _infer_modality(self, tensor_field: TensorField) -> Modality:
+        """
```
wanliAlex, line 84: Will this also work for sub-fields of a multimodal_combination field? Sub-fields will be collected as tensor_fields even if it is not explicitly set as tensor_fields in the index_settings, right?

papa99do: Yes, correct. They will be collected as tensor_fields as well. And their type is also defined in the index settings.

wanliAlex (approved):

wanliAlex (approved):
LGTM

PR #1209, opened 2025-04-29
Add a modality check in streaming media preprocessor

wanliAlex:
## Change Summary

Add a modality check in the streaming media preprocessor, ensuring that the detected media type matches the expected type. This could happen in a structured index as the given modality is inferred from the field type instead of the file itself. Without this detection, ffmpeg will silently process the given video file even if the file is in an audio format.

## Checklist

### For new field types:

### DIFF
--- a/src/marqo/core/inference/api/exceptions.py
+++ b/src/marqo/core/inference/api/exceptions.py
@@ -29,3 +29,8 @@ class UnsupportedModalityError(InferenceError):
 class MediaExceedsMaxSizeError(InferenceError):
     """Raised when the media exceeds the maximum size limit"""
     pass
+
+
+class MediaMisMatchError(InferenceError):
+    """Raised when the media does not match the expected type"""
+    pass
--- a/src/marqo/core/inference/api/exceptions.py
+++ b/src/marqo/core/inference/api/exceptions.py
@@ -31,6 +31,6 @@ class MediaExceedsMaxSizeError(InferenceError):
     pass
 
 
-class MediaMisMatchError(InferenceError):
+class MediaMismatchError(InferenceError):
     """Raised when the media does not match the expected type"""
     pass
--- a/src/marqo/inference/media_download_and_preprocess/streaming_media_processor.py
+++ b/src/marqo/inference/media_download_and_preprocess/streaming_media_processor.py
@@ -47,7 +47,7 @@ class StreamingMediaProcessor:
         self.total_size, self.duration, self.probed_modality = self._fetch_file_metadata()
 
         if self.modality != self.probed_modality:
-            raise MediaMisMatchError(
+            raise MediaMismatchError(
                 f"Error processing media file {self.url}. The provided modality {self.modality} does not match the "
                 f"detected modality {self.probed_modality}. Please check your media file and try again. If you are using"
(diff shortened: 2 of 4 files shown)


src/marqo/inference/media_download_and_preprocess/streaming_media_processor.py
```diff
@@ -44,7 +44,14 @@ def __init__(
         self.modality = preprocessing_config.modality
         
         self.media_download_header = self._convert_headers_to_cli_format(preprocessing_config.download_header)
-        self.total_size, self.duration = self._fetch_file_metadata()
+        self.total_size, self.duration, self.probed_modality = self._fetch_file_metadata()
+
+        if self.modality != self.probed_modality:
```
papa99do, line 49: Is it possible that the probed modality differs from the mime type of the downloaded sample? Could this check impact unstructured index?

src/marqo/core/inference/api/exceptions.py
```diff
@@ -29,3 +29,8 @@ class UnsupportedModalityError(InferenceError):
 class MediaExceedsMaxSizeError(InferenceError):
     """Raised when the media exceeds the maximum size limit"""
     pass
+
+
+class MediaMisMatchError(InferenceError):
```
papa99do: nit. mismatch is one word, so it should be `MediaMismatchError`.

src/marqo/inference/media_download_and_preprocess/streaming_media_processor.py
wanliAlex: Theoretically, it is possible. But based on the integration test results, we don't see a problem. We can send a parameter (check_modality: bool) in the request.

src/marqo/core/inference/api/exceptions.py
wanliAlex: fixed

src/marqo/inference/media_download_and_preprocess/streaming_media_processor.py
```diff
@@ -81,11 +88,34 @@ def _convert_headers_to_cli_format(self, raw_media_download_headers: Optional[Di
             raise InternalError("media_download_headers should be a dictionary")
         return "\r\n".join([f"{key}: {value}" for key, value in raw_media_download_headers.items()])
 
-    def _fetch_file_metadata(self) -> Tuple[float, float]:
+    def _infer_modality_from_probe(self, modality_list: list[str], format_name: Optional[str]) -> Optional[Modality]:
+        """
+        Infer the modality from the probed media file. This is used to determine whether the media is audio or video.
+        """
+        if Modality.VIDEO in modality_list:
+            if "image" in format_name:
```
papa99do: I tried `ffprobe` command with a png file, and got this output:
```
ffprobe \
  -v error \
  -show_entries stream=codec_type,format=size,duration,format_name \
  -of json \
  -probesize 256K \
  "https://res.cloudinary.com/zenbusiness/image/upload/v1670445040/logaster/logaster-2013-06-jpg.png"
{
    "programs": [

    ],
    "stream_groups": [

    ],
    "streams": [
        {
            "codec_type": "video"
        }
    ]
}
```

papa99do (approved):

PR #1228, opened 2025-05-25
Context documents in search (Personalization)

vicilliar:
## Change Summary
Supports new key `documents` in `context` parameter of search. This allows for a dict of doc_id:weight pairs to be provided by user for vectors to be extracted and interpolated for tensor search or hybrid search with tensor retrieval or ranking.

Updated example context object:
```
    {
        "tensor": [{"vector": [1 for i in range(768)], "weight": 1}], 
        "documents": {
            "ids": {"doc1": 1},
            "parameters": {
                "excludeInputDocuments": False,
                "tensorFields": ["text_field_1"],
                "concurrency": 10
            }
        }
    }
```
Interpolation is now done on all vectors (query, context.tensor, context.document) instead of previous behavior of weighted average.
Absolute value is now used when summing vector weights in interpolation (This makes ZeroSumWeightsError obsolete, we now use AllZeroWeightsError instead)
Interpolation logic implemented in numpy.
Fixed some error messages for hybrid / pydantic model validation.

New Env Vars:
`MAX_SEARCH_CONTEXT_DOCS` - Max number of ids that can be used in context.documents.ids
`VESPA_ASYNC_POOL_SIZE` - Size of persistent async vespa pool
`MARQO_CONCURRENCY_LIMIT_PER_GET_REQUEST` - concurrency limit of semaphore per vespa_client.get_batch request 

Removed Env Vars:
`VESPA_GET_POOL SIZE` - replaced by both `VESPA_ASYNC_POOL_SIZE` and `MARQO_CONCURRENCY_LIMIT_PER_GET_REQUEST`

The following optimizations have been implemented:
Getting doc vectors fetches only essential embeddings fields from Vespa
A persistent `AsyncTransport` is created upon vespa client initialization instead of being created each request
Index is fetched from cache instead of Vespa
Fetched doc vectors are not formatted into document form with Pydantic. Vectors are just used directly

## Related Jira Ticket
https://s2search.atlassian.net/jira/polaris/projects/MOSD/ideas/view/4863474?selectedIssue=MOSD-23&issueViewLayout=sidebar&issueViewSection=overview&atlOrigin=eyJpIjoiNzcyNDBkNzVjM2JhNDE0Y2I4ODUzNDY3YTc2MmE0ZmUiLCJwIjoiaiJ9

## Checklist
py-marqo changes: https://github.com/marqo-ai/py-marqo/pull/287

### For new field types:


src/marqo/core/search/hybrid_search.py
```diff
@@ -67,6 +69,8 @@ def search(
                 hybrid_parameters: HybridParameters object to specify all parameters for hybrid search. If not provided,
                     default values will be used.
                 facets: FacetsParameters object to specify facets for the search. If not provided, no facets will be returned.
+                track_total_hits: if True, total hits will be returned. If not provided, no total hits will be returned.
```
wanliAlex: What is total hits? The total number of hits of lexical and tensor search before the fusion?

src/marqo/core/search/hybrid_search.py
```diff
@@ -182,19 +186,25 @@ def search(
             query_text_search = lexical_query
 
             if context is None:
+                # If no context, create it with a tensor component
                 context = SearchContext(
                     tensor=[SearchContextTensor(vector=tensor_query, weight=1)]
                 )
+            elif context.tensor is None:
```
wanliAlex, line 201: Is it possible to reach here with the pydantic validation?

src/marqo/core/search/recommender.py
```diff
@@ -135,6 +103,76 @@ def recommend(self,
...
+        """
+        # TODO - The dependence on Config in tensor_search is bad design. Refactor to require specific dependencies
+        from marqo import config
+        from marqo.tensor_search import tensor_search, index_meta_cache
+
+        t0 = timer()
+
+        marqo_index = index_meta_cache.get_index(index_management=self.index_management, index_name=index_name)
+
+        if interpolation_method is None:
```
wanliAlex, line 188: Is it a good idea to read the default here? Maybe put it in the API or something?

src/marqo/core/utils/vector_interpolation.py
```diff
@@ -69,23 +70,30 @@ def interpolate(self, vectors: List[List[float]], weights: List[float], prenorma
...
             raise ValueError('Vectors and weights must have the same length')
 
-        weight_sum = sum(weights)
+        # Convert inputs to NumPy arrays for faster processing
+        np_vectors = np.array(vectors)
+        np_weights = np.array(weights)
+        
+        # Check if all vectors have the same length
+        if len(set(len(v) for v in vectors)) != 1:
+            raise ValueError('Vectors must have the same length')
```
wanliAlex: Will this be interpreted as a 500 error? Should it be 400?

src/marqo/core/utils/vector_interpolation.py
```diff
@@ -69,23 +70,30 @@ def interpolate(self, vectors: List[List[float]], weights: List[float], prenorma
...
+        
+        # Get sum of absolute values of weights
```
wanliAlex, line 83: This will go to 0 if and only if all weights are 0. Should we actually raise 400 for a 0 weight as it is meaningless?

src/marqo/core/utils/vector_interpolation.py
```diff
@@ -69,23 +70,30 @@ def interpolate(self, vectors: List[List[float]], weights: List[float], prenorma
         if len(vectors) != len(weights):
             raise ValueError('Vectors and weights must have the same length')
 
-        weight_sum = sum(weights)
+        # Convert inputs to NumPy arrays for faster processing
```
wanliAlex: Let's ensure all the changes in this file are unit tested.

src/marqo/tensor_search/models/api_models.py
```diff
@@ -284,6 +286,10 @@ def get_context_tensor(self) -> Optional[List[SearchContextTensor]]:
         """Extract the tensor from the context, if provided"""
         return self.context.tensor if self.context is not None else None
 
+    def get_context_documents(self) -> Optional[SearchContextDocuments]:
+        """Extract the documents from the context, if provided"""
+        return self.context.documents if self.context is not None else None
```
wanliAlex, line 315: just `return self.context.documents`?

src/marqo/tensor_search/models/search.py
```diff
@@ -63,9 +63,33 @@ class SearchContextTensor(BaseModel):
     vector: List[float]
     weight: float
 
+    def __init__(self, **data):
+        try:
+            super().__init__(**data)
+        except ValidationError as e:
```
wanliAlex: Why do we add this change here? Did we change the 422 error to 400 error? This could be a breaking change.

src/marqo/tensor_search/models/search.py
```diff
@@ -63,9 +63,33 @@ class SearchContextTensor(BaseModel):
...
+            raise InvalidArgError(message=e.json())
+
```
wanliAlex, line 68: nit: 2 lanes space

src/marqo/tensor_search/models/search.py
```diff
@@ -63,9 +63,33 @@ class SearchContextTensor(BaseModel):
...
+class SearchContextDocumentsParameters(BaseModel):
+    tensorFields: Optional[List[str]] = None
+    excludeInputDocuments: Optional[bool] = True
```
wanliAlex: Should this be Optional[bool] = True or bool = True? Do you want to accept None here?

src/marqo/tensor_search/models/search.py
```diff
@@ -63,9 +63,33 @@ class SearchContextTensor(BaseModel):
...
+        # Manually raising error because pydantic gives confusing error message
+        if not v:
+            raise InvalidArgError('context["documents"]["ids"] must be present and a non-empty dict of '
+                                  'document id to weight pairs.')
+        return v
 
 class SearchContext(BaseModel):
-    tensor: List[SearchContextTensor]
+    tensor: Optional[List[SearchContextTensor]]
+    documents: Optional[SearchContextDocuments]
```
wanliAlex, line 99: Do we also need a restriction on the number of documents?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -853,29 +869,69 @@ def get_query_vectors_from_jobs(
...
+                    collected_vectors.extend(vector_list)
+                    collected_weights.extend([weight] * len(vector_list))
+
+                # Save original doc ids for exclusion filtering
+                all_document_ids = list(context_documents.ids.keys())
+
+                # Determine default interpolation method using normalize embeddings if documents provided
+                if interpolation_method is None:
+                    interpolation_method = config.recommender.get_default_interpolation_method(q.index)
+            else:
```
wanliAlex: Is this duplicated with the line `interpolation_method = config.recommender.get_default_interpolation_method(q.index)`? Can we delete else and the line `interpolation_method = config.recommender.get_default_interpolation_method(q.index)` and always run the block to determine the interpolation here?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -838,7 +851,10 @@ def get_query_vectors_from_jobs(
 
         if isinstance(q.q, dict) or q.q is None:
             ordered_queries = list(q.q.items()) if isinstance(q.q, dict) else None
-            weighted_vectors = []
+            # Store weights and vectors separately for use in interpolation
+            collected_weights: List[List[float]] = []
```
wanliAlex, line 892: nit: you can still use np to finish the computation inside this method and convert it to list[float] at the end.

wanliAlex (changes requested):

src/marqo/core/search/hybrid_search.py
vicilliar: Yes it can, because context.tensor can be None if context.documents is defined

src/marqo/core/search/recommender.py
vicilliar: We can't put it in the search/recommend API because we need to fetch the index type when deciding the default. So it has to be after fetching the index

src/marqo/core/utils/vector_interpolation.py
vicilliar: I feel like it should still be 500, because this could only happen from some internal error. Either:
1. Some doc vectors in an index are of different size
2. The query vectors generated are the wrong size
3. context vector is the wrong size, which should be caught in validation

src/marqo/core/utils/vector_interpolation.py
vicilliar: Updated error message, added AllZeroWeightsError. This should still raise a 400, for all 0 weights

src/marqo/tensor_search/models/api_models.py
vicilliar: But this will break if `self.context = None`

src/marqo/tensor_search/models/search.py
vicilliar: Added this to give a better error message, but fixed it by adding `skip_on_failure=True` to the `validate_at_least_one_context_exists` root validator. You're right, it might be a breaking change, I will remove this now.

src/marqo/tensor_search/models/search.py
vicilliar: fixed

src/marqo/tensor_search/models/search.py
vicilliar: good point, removed the `Optional`

src/marqo/tensor_search/models/search.py
vicilliar: Added env var MARQO_MAX_SEARCH_CONTEXT_DOCS (defaults to 10, as per Owen)

src/marqo/tensor_search/tensor_search.py
vicilliar: Updated the get_default_interpolation_method method and removed redundant block. Just added the logic inside the method:
```
        if marqo_index.normalize_embeddings:
            if context_documents is not None:
                return InterpolationMethod.SLERP
            else:
                # NLERP is used to preserve existing search behavior with no context docs.
                return InterpolationMethod.NLERP
        else:
            return InterpolationMethod.LERP
```

src/marqo/tensor_search/tensor_search.py
vicilliar: Isn't the only computation being done in `vector_interpolation.interpolate`, and that is already converting the weights and vectors into np arrays. The only thing being done in this method is collecting the list of vectors and weights

src/marqo/core/search/recommender.py
```diff
@@ -86,11 +63,7 @@ def recommend(self,
 
         marqo_index = index_meta_cache.get_index(index_management=self.index_management, index_name=index_name)
 
-        if interpolation_method is None:
-            interpolation_method = self._get_default_interpolation_method(marqo_index)
-
-        vector_interpolation = from_interpolation_method(interpolation_method)
-
+        # See if recommend breaks for unstructured
```
wanliAlex: Should this work for both structured and unstructured?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1185,3 +1252,128 @@ def delete_documents(config: Config, index_name: str, doc_ids: List[str]):
             document_ids=doc_ids,
         )
     )
+
+
+def get_embedding_field_names(marqo_index: MarqoIndex, tensor_field_names: Optional[List[str]] = None) -> List[str]:
```
wanliAlex: nit: return type

src/marqo/core/search/hybrid_search.py
vicilliar: It's total number of hits before fusion (excluding duplicates). So if tensor has 3 hits, lexical has 5, but 2 results show up in both lists, total hits will be 6.

src/marqo/core/search/hybrid_search.py
```diff
@@ -72,8 +74,10 @@ def search(
                 hybrid_parameters: HybridParameters object to specify all parameters for hybrid search. If not provided,
                     default values will be used.
                 facets: FacetsParameters object to specify facets for the search. If not provided, no facets will be returned.
+                track_total_hits: if True, total hits before reranking will be returned.
```
wanliAlex: This is just a missed parameter in the doc string right?

src/marqo/core/search/recommender.py
```diff
@@ -21,52 +22,29 @@ def __init__(self, vespa_client: VespaClient, index_management: IndexManagement,
...
-            verbose: Verbosity level
-            reranker: Reranker to use
-            filter: Filter string
-            attributes_to_retrieve: List of attributes to retrieve
-            score_modifiers: Score modifiers to apply
-            rerank_depth: Rerank depth
+            concurrency: Max number of concurrent requests to use when fetching documents by batch
+
+        Returns:
+            A dictionary mapping document IDs to lists of vector embeddings
```
wanliAlex: nit: It would be better to be clear that this method returns the vectors of the corresponding tensor fields. Also, is the order of the vectors guaranteed here?

wanliAlex (changes requested):

src/marqo/core/search/recommender.py
```diff
@@ -100,41 +74,124 @@ def recommend(self,
...
-                    vectors.append(tensor_facet['_embedding'])
-
-            doc_vectors[document['_id']] = vectors
+            
+            # Flatten all embeddings from all fields for this document
+            for field_name, embedding_list in field_embeddings.items():
+                # For legacy unstructured indices, field_name will be "marqo__embeddings"
+                # and we should include all embeddings regardless of tensor_fields filter
+                # since all embeddings are stored together in marqo__embeddings
+                if (tensor_fields is None or
```
wanliAlex, line 120: What would happen if `tensor_fields = []`?

src/marqo/core/search/recommender.py
```diff
@@ -100,41 +74,124 @@ def recommend(self,
...
+            index_name=index_name,
+            documents=documents,
+            tensor_fields=tensor_fields
+        )
+
+        # Save original document IDs for filtering
+        if isinstance(documents, dict):
+            all_document_ids = list(documents.keys())
+        else:
+            all_document_ids = documents
```
wanliAlex, line 204: Do we have duplication detection for document ids if this is a list?

src/marqo/core/search/recommender.py
```diff
@@ -150,19 +207,16 @@ def recommend(self,
...
-            else:  # lerp or nlerp
+                    'SLERP cannot interpolate vectors with all zero weights. '
+                    'Please ensure at least one weight is non-zero.'
+                )
+            else:
                 raise InvalidArgumentError(
-                    'Sum of weights is zero. LERP/NLERP requires non-zero sum of weights',
-                    cause=e
-                ) from e
+                    'All weights are zero. LERP/NLERP requires at least one non-zero weight'
```
wanliAlex: It seems you can combine the error catch here?

src/marqo/core/search/recommender.py
```diff
@@ -201,13 +255,25 @@ def recommend(self,
...
+        """
+        Returns the default interpolation method based on the index configuration and whether context documents
+        exist. For recommend endpoint, context documents always exist.
+
+        For indexes that normalize embeddings, SLERP is used if context documents are provided, NLERP is used
+        otherwise (None).
+        """
         if marqo_index.normalize_embeddings:
-            return InterpolationMethod.SLERP
+            if context_documents is not None:
```
wanliAlex, line 274: What if context_documents is []

src/marqo/tensor_search/models/search.py
```diff
@@ -59,13 +60,42 @@ def add_content(self, content: List[Union[str, List[str]]]) -> VectorisedJobPoin
...
         )
 
+
 class SearchContextTensor(BaseModel):
     vector: List[float]
     weight: float
 
 
+class SearchContextDocumentsParameters(BaseModel):
+    tensorFields: Optional[List[str]] = None
```
wanliAlex: One comment I got from @Yihan is that we use `tensor_fields: Optional[List[str]] = Field(None, aliase="tensorFields")` in the model so we can keep the snake case in python but also accept camelCase in API.

src/marqo/tensor_search/models/search.py
```diff
@@ -59,13 +60,42 @@ def add_content(self, content: List[Union[str, List[str]]]) -> VectorisedJobPoin
...
+    excludeInputDocuments: bool = True
+    concurrency: Optional[int] = None
+
+    @validator('tensorFields', pre=True, always=True)
+    def check_tensor_fields_not_empty(cls, v):
```
wanliAlex, line 75: Ah I see the check now. Makes sense.

src/marqo/tensor_search/models/search.py
```diff
@@ -59,13 +60,42 @@ def add_content(self, content: List[Union[str, List[str]]]) -> VectorisedJobPoin
...
+        if v == []:
+            raise InvalidArgError('context document tensorFields parameter must be non-empty list.'
```
wanliAlex: Raise `ValueError` in the pydantic validator

src/marqo/tensor_search/models/search.py
```diff
@@ -59,13 +60,42 @@ def add_content(self, content: List[Union[str, List[str]]]) -> VectorisedJobPoin
...
+class SearchContextDocuments(BaseModel):
+    ids: Optional[Dict[str, float]]
+    # If not provided, default parameters are created
+    parameters: Optional[SearchContextDocumentsParameters] = SearchContextDocumentsParameters()
+
+    @validator('ids', pre=True, always=True)
+    def check_ids_not_empty(cls, v):
+        # Manually raising error because pydantic gives confusing error message
+        if not v:
+            raise InvalidArgError('context["documents"]["ids"] must be present and a non-empty dict of '
```
wanliAlex: Raise ValueError in the pydantic validator. I think you are trying to return a 400 error instead of a 422 error to the users. But I would consider this as a breaking change.

src/marqo/tensor_search/models/search.py
```diff
@@ -59,13 +60,42 @@ def add_content(self, content: List[Union[str, List[str]]]) -> VectorisedJobPoin
...
+
+
+class SearchContextDocuments(BaseModel):
+    ids: Optional[Dict[str, float]]
+    # If not provided, default parameters are created
+    parameters: Optional[SearchContextDocumentsParameters] = SearchContextDocumentsParameters()
+
+    @validator('ids', pre=True, always=True)
+    def check_ids_not_empty(cls, v):
+        # Manually raising error because pydantic gives confusing error message
```
wanliAlex: I am not sure if this a good idea tbh. A proper change would be to update the error handling logic in the API.

src/marqo/tensor_search/models/search.py
```diff
@@ -75,10 +105,22 @@ def __init__(self, **data):
 
     @validator('tensor', pre=True, always=True)
     def check_vector_length(cls, v):
-        if not (1 <= len(v) <= 64):
-            raise InvalidArgError('The number of tensors must be between 1 and 64')
+        if v is not None:
```
wanliAlex: What if this is not a list? Will len(v) raise a proper error here?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1093,8 +1156,8 @@ def _vector_text_search(
         index=marqo_index, modelAuth=model_auth, text_query_prefix=text_query_prefix, rerankDepth=rerank_depth
     )]
 
-    with RequestMetricsStore.for_request().time(f"search.vector_inference_full_pipeline"):
-        qidx_to_vectors: Dict[Qidx, List[float]] = run_vectorise_pipeline(config, queries, device)
+    with RequestMetricsStore.for_request().time(f"search.vector.inference.full_pipeline"):
```
wanliAlex: why do we make this change? Should full_pipeline have a higher layer than sub-jobs?

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1205,3 +1268,139 @@ def delete_documents(config: Config, index_name: str, doc_ids: List[str]):
...
+            
+            return ([tf.name for tf in requested_tensor_fields],
+                    [tf.embeddings_field_name for tf in requested_tensor_fields])
+        else:
+            # Index has no tensor fields at all
+            raise core_exceptions.InvalidArgumentError(
+                f"Index '{marqo_index.name}' has no tensor fields, cannot retrieve embeddings"
+                f" for {tensor_field_names}"
+            )
+    else:
```
wanliAlex, line 1327: As discussed, we want to disable the support for this and raise and error.

src/marqo/tensor_search/tensor_search.py
```diff
@@ -1205,3 +1268,139 @@ def delete_documents(config: Config, index_name: str, doc_ids: List[str]):
...
+        if response.status == 200:
+            # Extract vectors directly (for structured and semi-structured)
+            # Skip turning into marqo document
+            raw_response_dict = response.document.fields
+            doc_id = raw_response_dict["marqo__id"]
+
+            # Initialize the result for this document ID
+            result[doc_id] = {}
+            # Legacy unstructured will have exactly 1 vespa & 1 marqo embedding field name. Treat it like 1 tensor field.
+            for i in range(len(viable_tensor_fields)):
```
wanliAlex: I am not sure if you can use both `i` in these loops. Maybe change the first one to `for response in batch_get.responses:`

wanliAlex (changes requested):

src/marqo/core/search/recommender.py
vicilliar: It used to be untested for unstructured (unstructured means legacy, because we now use semi-structured). But existing behavior of recommend should work for structured and semi-structured.

src/marqo/tensor_search/tensor_search.py
vicilliar: fixed

src/marqo/core/search/hybrid_search.py
vicilliar: yes, i didnt add a new param, it's just to complete the docstring

src/marqo/core/search/recommender.py
vicilliar: Added to docstring: This is flattened to 1 list per document ID (not separated by tensor field). Order of embeddings is not guaranteed.

src/marqo/core/search/recommender.py
vicilliar: In search with context documents: We would error our before getting here because tensor_fields cannot be empty list.

In recommend: doc_vectors would just end up empty because none of the fields in field_embeddings would be present in tensor_fields.

src/marqo/core/search/recommender.py
vicilliar: added duplicate id detection

src/marqo/core/search/recommender.py
vicilliar: combined

src/marqo/core/search/recommender.py
vicilliar: It would still return NLERP. But in validation for both recommend and search with context docs, it will raise an InvalidArgumentError anyway.

We have:
```
        if documents is None or len(documents) == 0:
            raise InvalidArgumentError('No document IDs provided')
```
And in context docs search, we have a pydantic validator to make sure it's not empty.

src/marqo/tensor_search/models/search.py
vicilliar: added alias:
```
tensor_fields: Optional[List[str]] = Field(None, alias='tensorFields')
exclude_input_documents: bool = Field(True, alias='excludeInputDocuments')
```

src/marqo/tensor_search/models/search.py
vicilliar: Ah I think I was confused by the error handling in `SearchContext` here:
```
    def __init__(self, **data):
        try:
            super().__init__(**data)
        except ValidationError as e:
            raise InvalidArgError(message=e.json())
```

but I don't think it was a good example to follow

src/marqo/tensor_search/models/search.py
vicilliar: done

src/marqo/tensor_search/models/search.py
vicilliar: Yep, I think the comment is inaccurate now. But this check for `v == []` seems like the right approach now

src/marqo/tensor_search/models/search.py
vicilliar: added a manual list type check

src/marqo/tensor_search/tensor_search.py
vicilliar: Our telemetry keys start with `search.vector`, `search.lexical`, or `search.hybrid` depending on what search method they used. I changed this key to be under the `search.vector` category, but I think that's wrong now, since hybrid uses it too. The keys could instead be something like:
```
search.vector.inference_pipeline.full
search.hybrid.inference_pipeline.full

# I can change the other new telemetry keys for uniformity:
search.inference_pipeline.vectorise_jobs
search.inference_pipeline.interpolate_vectors
search.inference_pipeline.get_doc_vectors_from_ids
```

vicilliar: But if we don't want to change these existing keys for full inference pipeline, I can keep them as is

src/marqo/tensor_search/tensor_search.py
vicilliar: you're right, added the error and tests

src/marqo/tensor_search/tensor_search.py
vicilliar: Good catch, this could've been an issue. I need to use indexes, but ill give a more descriptive name

src/marqo/tensor_search/models/search.py
wanliAlex: nit: use min_items or max_items in pydantic.Field: https://docs.pydantic.dev/1.10/usage/types/#arguments-to-conlist

tests/api_tests/v1/tests/api_tests/test_search_common.py
```diff
@@ -388,8 +386,101 @@ def test_hybrid_search_validations(self):
                             }
                         )
                     assert e.exception.status_code == 400
-                    assert "'hybridParameters.queryTensor' cannot be provided when 'retrievalMethod' and 'rankingMethod' are both 'lexical'." in str(
-                        e.exception)
+                    assert "'hybridParameters.queryTensor' cannot be provided when 'retrievalMethod' and 'rankingMethod' are both 'lexical'." in str(e.exception)
+
+    def test_search_with_context_documents(self):
```
wanliAlex, line 391: Try to add an error case test?  E.g., invalid documents or not existing documents.

tests/integ_tests/core/search/test_recommender.py
```diff
@@ -721,5 +695,77 @@ def test_recommend_rerank_depth_with_limit_and_offset(self):
                     # but for one searchable attribute and one shard is capped at ef_search
                     self.assertTrue(len(res["hits"]) <= 3)
 
+    def test_get_doc_vectors_from_ids_noDocumentsProvided_fails(self):
```
wanliAlex, line 698: Do we have tests for invalid document IDs? Do we have a validation on this?

wanliAlex (changes requested):

tests/api_tests/v1/tests/api_tests/test_search_common.py
vicilliar: no way to actually reach existing validation (pydantic model will typecast ids into str if possible)

tests/integ_tests/core/search/test_recommender.py
vicilliar: added validation for enforcing string type in ids, and tests

wanliAlex (approved):
LGTM

PR #1234, opened 2025-05-30
Implement hybrid RRF pagination deduplication with Vespa-based state

danyilq:
## Change Summary

This PR introduces a robust fix and enhancement for hybrid RRF pagination in Marqo’s Vespa backend. Previously, hybrid pagination applied the offset independently to both tensor and lexical results before fusion, resulting in inconsistent or incomplete paginated results.

### Key improvements include:
- **Correct Pagination Behavior**: Offsets are now applied after fusion instead of prematurely at the retrieval stage. The system sets `offset=0` and `limit=offset+limit`, retrieves and fuses full results, then slices the final output according to the original pagination parameters.
- **Pagination State Tracking**: Introduced a new schema to track and store previously seen document IDs for each (`query - offset`) combination using a hash. This avoids duplicates and ensures continuity between paginated requests.
- **Custom Searcher Enhancements**: The custom searcher is extended to support document access (get, update, visit, etc.) through documentAccess, allowing dynamic reads/writes from the pagination schema during search execution.
- **Automatic Garbage Collection**: Leveraged Vespa’s selection mechanism to purge expired pagination documents. Documents with an updated_at timestamp older than a configured TTL (default: 1800s) are automatically deleted.

#### Pagination Schema
- Fields:
  - `offsets`: Map<int, List<doc_id>>, tracks which document IDs were returned for each offset
  - `updated_at`: Timestamp, used for TTL-based cleanup by Vespa

#### Request Flow
1. **Hashing**: A hash is generated for each hybrid RRF or disjunction query using ORJSON, excluding offset to ensure stability across paginated calls.
2. **Document Fetch**: If offset > 0 and pagination schema/hash is available, the custom searcher attempts to fetch the corresponding pagination document.
3. **ID Exclusion**:
   - If found, all document IDs associated with offsets ≤ offset - limit are collected into an idsToExclude set.
   - Tensor and lexical results are filtered against this exclusion set prior to fusion.
4. **Result Fusion & Trimming**:
   - Hybrid RRF fusion occurs on filtered results.
   - The final hits list is sliced using the original (offset, limit) values accounting for number of excluded hit ids.
5. **Pagination State Update**:
   - If `offset` == 0 or `offset` - `limit` exists in the state, an async update is triggered.
   - The `offsets[offset]` entry is set to the list of returned document IDs.
   - `updated_at` is refreshed to the current time.
6. **Cache Control**:
   - If the previous state (offset - limit) is not found, a metadata `noCache` is returned, prompting Marqo to set a Cloudflare no-store header.
## Checklist

### For new field types:


src/marqo/core/index_management/vespa_application_package.py
```diff
@@ -125,14 +125,23 @@ def _add_component(self, parent: ET.Element, tag: str, name: str):
...
                 normalized.append(child)
             return normalized
 
         elements_self = sorted([ET.tostring(normalize(elem), encoding='unicode') for elem in self._root.findall(xml_path)])
         elements_other = sorted([ET.tostring(normalize(elem), encoding='unicode') for elem in other._root.findall(xml_path)])
+        if xml_path == "content/documents":
+            # Ignore 'documents' which is always the first element, only compare 'document' objects
+            elements_self[0] = elements_self[0][elements_self[0].find('>'):]
+            elements_other[0] = elements_other[0][elements_other[0].find('>'):]
+        print(elements_self, elements_other)
```
wanliAlex: remove print

src/marqo/core/index_management/vespa_application_package.py
```diff
@@ -125,14 +125,23 @@ def _add_component(self, parent: ET.Element, tag: str, name: str):
     def compare_element(self, other: 'ServicesXml', xml_path: str) -> bool:
         def normalize(elem: ET.Element):
             # Sort attributes and child elements to normalize
+            exclude_types = ["marqo__pagination"]
```
wanliAlex, line 128: why do we exclude marqo__pagination?

src/marqo/core/index_management/vespa_application_package.py
```diff
@@ -720,6 +730,37 @@ def _configure_query_profiles(self, backup: VespaAppBackup) -> None:
         logger.debug(f'Configuring query profiles {content}')
         self._store.save_file(content, 'search', 'query-profiles', 'default.xml', backup=backup)
 
+    def _configure_pagination_schema(self, backup: VespaAppBackup) -> None:
+        schema_content = textwrap.dedent('''\
+            schema marqo__pagination {
+                document {
+                    field offsets type map<string, array<string>> {
+                        indexing: summary
```
wanliAlex, line 737: Why not put it in the memory for fast retrieval?

src/marqo/core/models/marqo_query.py
```diff
@@ -11,11 +11,37 @@
 from marqo.core.models.hybrid_parameters import RankingMethod, HybridParameters
 from marqo.tensor_search.models.sort_by_model import SortByModel
 from marqo.tensor_search.models.relevance_cutoff_model import RelevanceCutoffModel
+import orjson
+
+def orjson_dumps(v, *, default, **kwargs):
```
wanliAlex, line 16: why do we add this?

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -615,6 +617,13 @@ def _to_vespa_hybrid_query(self, marqo_query: MarqoHybridQuery) -> Dict[str, Any
                         facet_queries.append(facets_query_skeleton % (query_yql, new_facets_term))
             facet_queries = QUERY_DELIMITER.join(facet_queries)
 
+        query_hash_without_offset = None
+        if marqo_query.offset % marqo_query.limit == 0:
+            if marqo_query.hybrid_parameters.retrievalMethod.lower() == RetrievalMethod.Disjunction:
+                query_hash_without_offset = marqo_query.get_query_hash_without_offset()
+        else:
+            logger.warning("Offset is not a multiple of limit. This is a wrong usage of pagination.")
```
wanliAlex, line 625: I guess no one will see this warning. Should we just raise an error and block this usage?

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -644,6 +653,9 @@ def _to_vespa_hybrid_query(self, marqo_query: MarqoHybridQuery) -> Dict[str, Any
             'marqo__ranking.lexical.tensor': common.RANK_PROFILE_HYBRID_BM25_THEN_EMBEDDING_SIMILARITY,
             'marqo__ranking.tensor.lexical': common.RANK_PROFILE_HYBRID_EMBEDDING_SIMILARITY_THEN_BM25,
 
+            "marqo__hybrid.pagination_schema": constants.MARQO_PAGINATION_SCHEMA_NAME,
+            "marqo__hybrid.pagination_hash": query_hash_without_offset,
+            "marqo__hybrid.pagination_limit_cutoff": os.environ.get("MARQO_PAGINATION_LIMIT_CUTOFF", 600),
```
wanliAlex, line 658: Not sure if it's a good idea to read the environment variable here. I understand the pain of injecting the dependencies from the start.  What do you think @farshidz ?

src/marqo/vespa/models/query_result.py
```diff
@@ -10,6 +10,7 @@ class MarqoFields(BaseModel):
     sort_candidates: Optional[int] = Field(None, alias='sortCandidates')
     relevant_candidates: Optional[int] = Field(None, alias='relevantCandidates')
     probe_candidates: Optional[int] = Field(None, alias='probeCandidates')
+    no_cache: Optional[bool] = Field(None, alias='noCache')
```
wanliAlex, line 13: nit: would has_cache be clearer to avoid the inverse logic?

src/marqo/core/index_management/vespa_application_package.py
danyilq: for 2 reasons:
1. We are not doing search, so retrieval is not gonna be improved anyhow.
2. vespa doesn't support "attribute" for maps that have non-trivial data types. In this case arrays of strings

src/marqo/core/index_management/vespa_application_package.py
danyilq: Because otherwise rollback fails as previous version didn't have the schema for that. I have brought it up with Yihan and he said that just ignoring it is okay. it doesn't matter for our rollback

src/marqo/core/models/marqo_query.py
danyilq: orjson dump is significantly more efficient than regular pydantic's dump as it is implemented in rust. It was Yihan's recommendation and it saves almost 1ms in comparison.

src/marqo/core/structured_vespa_index/structured_vespa_index.py
danyilq: Well, not really. It is okay to support it if there are some edge usecases, blocking it is a little extra. Warning is just for us to see if we are trying to understand if customer is doing something wrong.

src/marqo/vespa/models/query_result.py
danyilq: we could, I named it no-cache as it matches the header that cloudflare uses. By default we pretty much always have cache, it is rather an edgecase like this that we don't.

src/marqo/core/index_management/vespa_application_package.py
wanliAlex: I was talking about the summary performance here: https://docs.vespa.ai/en/document-summaries.html#performance

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -163,6 +180,36 @@ public Result search(Query query, Execution execution) {
...
+        // Get the pagination document if paginationHash and paginationSchema are set and
+        // retrievalMethod is disjunction
+        if (shouldUsePagination(paginationHash, paginationSchema, retrievalMethod)) {
+            try {
+                docId = getPaginationDocumentId(paginationSchema, paginationHash);
+                AsyncParameters asyncParameters = new AsyncParameters();
+                docAccess = documentAccess.createAsyncSession(asyncParameters);
+                // Do not get document is offset is 0, as we do not need to exclude any IDs in this
+                // case.
+                if (offset != 0) {
```
wanliAlex, line 205: Do we need to instantiate the above objects if offset == 0?

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -163,6 +180,36 @@ public Result search(Query query, Execution execution) {
...
+                    paginationDocResult = docAccess.get(docId);
+                    logIfVerbose("Search for pagination document: " + docId, verbose);
+                }
+            } catch (Exception e) {
```
wanliAlex, line 209: Avoid a general exception catch

tests/integ_tests/tensor_search/integ_tests/test_search_relevance_cutoff_feature.py
```diff
@@ -841,7 +841,7 @@ def test_sort_and_relevance_cutoff_pagination(self):
         )
 
         page_2_sort_candidates = page_2_results["_sortCandidates"]
-        self.assertEqual(11, page_2_sort_candidates)
+        self.assertEqual(7, page_2_sort_candidates)
```
wanliAlex: Let's confirm with Owen if this change is expected or not. I feel like we may need to return the true `sortCandidates`, instead of `sortCandidates - offset`.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
danyilq: we are gonna be using docAccess and docId for update docs even if offset is 0

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
danyilq: uhh I agree, but at the same time I'm unaware of what exception could be raised here. The purpose of it is more so that pagination is guaranteed to not interfere with search if there is any issue with it.

src/marqo/core/index_management/vespa_application_package.py
wanliAlex: As discussed, @danyilq did some performance tests and didn't observe any improvement.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: That feels weird. If we don't expect an error here maybe we should just let it raise and return a 500?

danyilq: not sure, but we could. My opinion is that since search can successfully complete with it failing - it shouldn't raise 500

farshidz (commented):
Backwards compatibility tests are missing

src/marqo/core/models/marqo_query.py
```diff
@@ -91,3 +117,13 @@ def validate_searchable_attributes_and_score_modifiers(cls, values):
...
         return values
+
+    def get_query_hash_without_offset(self):
+        normalized_json = self.json(
+            exclude={'offset'},
+            sort_keys=True,
+            exclude_unset=False,
+            exclude_none=True,
+        )
+        hash_value = hashlib.sha256(normalized_json.encode('utf-8')).hexdigest()
```
farshidz, line 128: Did you consider getting a prefix of this to use less memory?

src/marqo/core/search/hybrid_search.py
```diff
@@ -339,6 +338,10 @@ def search(
                         gathered_results.get("facets", {}).update({facet_field_name: {}})
             if track_total_hits is not None and "totalHits" not in gathered_results:
                 gathered_results["totalHits"] = 0
+        if responses.root.fields.marqo_fields.no_cache is not None:
+            if gathered_results.get('marqo__headers') is None:
+                gathered_results['marqo__headers'] = {}
+            gathered_results['marqo__headers']["cache-control"] = "no-cache"
```
farshidz, line 344: I wonder if this should be a constant since it's repeated many times. Also I can't figure out where we actually add this header to the response. Is it actually returned to the user/caller?

src/marqo/core/index_management/vespa_application_package.py
```diff
@@ -720,6 +729,37 @@ def _configure_query_profiles(self, backup: VespaAppBackup) -> None:
...
```
farshidz, line 737: There would still be performance gains from making this an attribute to avoid disk access. But you would have to store the array as a string and deserialize on retrieval.

farshidz: In fact why didn't we store the full state as a string attribute? Does storing a map in Vespa have an advantage?

src/marqo/core/index_management/vespa_application_package.py
```diff
@@ -720,6 +729,37 @@ def _configure_query_profiles(self, backup: VespaAppBackup) -> None:
...
+                }
+            }
+        ''')
+        logger.debug(f'Configuring pagination schema')
+        self._store.save_file(schema_content, 'schemas', 'marqo__pagination.sd', backup=backup)
+        self._service_xml.add_schema('marqo__pagination')
+
+        # Configure garbage collection for pagination documents
+        documents_elem = self._service_xml._ensure_only_one('content/documents')
+        documents_elem.set('garbage-collection', 'true')
```
farshidz, line 754: Are you sure this doesn't affect index schemas?

src/marqo/core/index_management/vespa_application_package.py
```diff
@@ -720,6 +729,37 @@ def _configure_query_profiles(self, backup: VespaAppBackup) -> None:
...
+        documents_elem.set(
+            'garbage-collection-interval',
+            os.environ.get('PAGINATION_GARBAGE_COLLECTION_INTERVAL', '1800')
```
farshidz, line 757: This isn't the right way to work with configuration env vars. There's an existing pattern for reading these and also storing default values (default value shouldn't be here). Look up `MARQO_ENABLE_DEBUG_API` for example.

src/marqo/core/structured_vespa_index/structured_vespa_index.py
farshidz: I noted elsewhere env vars aren't being used correctly and need to follow the existing pattern at least.

src/marqo/tensor_search/api.py
```diff
@@ -441,7 +441,10 @@ def search(index_name: str, search_query_dict: dict, device: str = Depends(api_v
                 sort_by = search_query.sort_by,
                 interpolation_method=search_query.interpolationMethod
             )
-            return ORJSONResponse(result)
+            result_exclusions = {'marqo__headers'}
```
farshidz, line 444: Shouldn't you add this response header? Otherwise it's lost

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -163,6 +180,36 @@ public Result search(Query query, Execution execution) {
...
+
+        Set<String> idsToExclude = new HashSet<>();
+        AsyncSession docAccess = null;
+        com.yahoo.documentapi.Result paginationDocResult = null;
+        DocumentId docId = null;
+        Set<Integer> paginationStateOffsets = null;
+        Boolean noCache = null;
+        // Get the pagination document if paginationHash and paginationSchema are set and
+        // retrievalMethod is disjunction
+        if (shouldUsePagination(paginationHash, paginationSchema, retrievalMethod)) {
```
farshidz, line 198: Is there an opportunity here to feature flag the new pagination? If disabled, we simply send null schema here. Is that something that would disable it here?

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
farshidz: Agreed catching Exception seems like a bad idea. You could catch a generic exception class to be safe but not exception which can be anything including bug in your code.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
farshidz: Also which line are you expecting to fail? You seem to be wrapping a lot of logic including getting pagination document ID

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -292,6 +344,36 @@ public Result search(Query query, Execution execution) {
                             + " || TENSOR RESULTS: "
                             + resultTensor.toString(),
                     verbose);
+            if (shouldUsePagination(paginationHash, paginationSchema, retrievalMethod)) {
```
farshidz, line 347: Can you call this just once and store?

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -338,8 +426,40 @@ public Result search(Query query, Execution execution) {
                             rerankDepthGlobal,
                             limit,
                             offset,
+                            idsToExclude,
                             verbose);
         }
+        // Save pagination state if the performed request is not a jump (offset - limit is present
+        // in pagination document)
+        if (shouldUsePagination(paginationHash, paginationSchema, retrievalMethod)) {
```
farshidz, line 434: As before no need to call this three times.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -292,6 +344,36 @@ public Result search(Query query, Execution execution) {
...
+            if (shouldUsePagination(paginationHash, paginationSchema, retrievalMethod)) {
+                if (paginationDocResult != null && paginationDocResult.isSuccess()) {
+                    // Get the Document from the Result
+                    // Extract IDs list from the document
+                    com.yahoo.documentapi.DocumentResponse paginationDocumentResponse =
+                            (com.yahoo.documentapi.DocumentResponse) docAccess.getNext();
+                    MapFieldValue<StringFieldValue, Array<StringFieldValue>> paginationStateMap =
+                            getPaginationStateMap(paginationDocumentResponse);
+                    idsToExclude = getIdsToExclude(paginationStateMap, offset);
+                    paginationStateOffsets = getExistingPaginationStateOffsets(paginationStateMap);
```
farshidz, line 356: This is defined here to be used in a different scope. I'm not even sure why it compiles.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -379,12 +499,149 @@ public Result search(Query query, Execution execution) {
...
+    // gets Ids to exclude based on the pagination state map and current offset.
+    private Set<String> getIdsToExclude(
+            MapFieldValue<StringFieldValue, Array<StringFieldValue>> paginationStateMap,
+            Integer offset) {
+        Set<String> idsToExclude = new HashSet<>();
+        if (paginationStateMap != null) {
+            // Collect IDs to exclude from all previous offset pages
+            for (Map.Entry<StringFieldValue, Array<StringFieldValue>> entry :
+                    paginationStateMap.entrySet()) {
+                int pageOffset = Integer.parseInt(entry.getKey().getString());
```
farshidz, line 547: It would be nice if parsed these keys once rather than repeatedly. You could store in a red black tree for log time access instead of linear.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -379,12 +499,149 @@ public Result search(Query query, Execution execution) {
...
+    Set<Integer> getExistingPaginationStateOffsets(
+            MapFieldValue<StringFieldValue, Array<StringFieldValue>> paginationStateMap) {
+        Set<Integer> existingOffsets = new HashSet<>();
+        if (paginationStateMap != null) {
+            // Collect existing offsets from the pagination state map
+            for (StringFieldValue key : paginationStateMap.keySet()) {
+                try {
+                    existingOffsets.add(Integer.parseInt(key.getString()));
+                } catch (NumberFormatException e) {
+                    logger.warn("Invalid offset key in pagination state: " + key.getString());
```
farshidz, line 570: Wouldn't this indicate we have a bug? In which case it should just raise.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -591,7 +852,19 @@ HitGroup postProcessBySort(
...
-        result.trim(offset, limit);
+        if (idsToExclude != null && !idsToExclude.isEmpty()) {
+            logIfVerbose(
+                    String.format(
+                            "Trimming result list. " + "limit: %d, offset: %d", limit, offset),
+                    verbose);
+            int totalExclusions = idsToExclude.size();
+            result.trim(
+                    Math.max(0, offset - totalExclusions),
+                    Math.max(0, (limit + offset) - totalExclusions));
```
farshidz, line 863: Is this tested? Is this correct?

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -338,8 +426,40 @@ public Result search(Query query, Execution execution) {
...
+                            idsToExclude,
                             verbose);
         }
+        // Save pagination state if the performed request is not a jump (offset - limit is present
+        // in pagination document)
+        if (shouldUsePagination(paginationHash, paginationSchema, retrievalMethod)) {
+            if ((offset == 0
+                            || (paginationStateOffsets != null
+                                    && paginationStateOffsets.contains(offset - limit)))
+                    && offset + limit <= paginationLimitCutoff) {
```
farshidz, line 438: Is paginationLimitCutoff tested integ or API?

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -379,12 +499,149 @@ public Result search(Query query, Execution execution) {
...
+                            docType.getField("updated_at"),
+                            new com.yahoo.document.datatypes.LongFieldValue(
+                                    System.currentTimeMillis())));
+            docUpd.setCreateIfNonExistent(true);
+            com.yahoo.documentapi.Result res = docAccess.update(docUpd);
+            docAccess.getNext(0);
+            // raise an error if the update failed
+            if (!res.isSuccess())
+                if (res.type() == com.yahoo.documentapi.Result.ResultType.TRANSIENT_ERROR) {
+                    System.out.println("send queue full, waiting for some response");
```
papa99do, line 621: nit. is this for debugging? we should either delete it or use a logger

src/marqo/core/models/marqo_query.py
danyilq: test with different, smaller 16-byte hash `md5`

PR #1235, opened 2025-06-02
Load clip model with weight_only=False if weights only load fails

papa99do:
## Change Summary
After the upgrade to torch 1.13.1 and open_clip 2.32.0, some models trained/fine-tuned using torch 1.12.1 cannot be loaded since it contains more than weight change. If the model cannot be loaded due to this reason, we'll try to load it with weights_only set to False again.

## Checklist

- [ ] Documentation has been updated (N/A)
- [ ] Breaking changes are clearly identified (N/A)
- [ ] Python client changes linked or N/A (N/A)

### For new field types:

### DIFF
--- a/src/marqo/inference/native_inference/embedding_models/open_clip_model.py
+++ b/src/marqo/inference/native_inference/embedding_models/open_clip_model.py
@@ -148,7 +148,8 @@ class OpenCLIPModel(AbstractCLIPModel):
 
         return aggregated_image_preprocess_config
 
-    def _load_model_and_image_preprocessor_from_checkpoint(self) -> Tuple[torch.nn.Module, Compose]:
+    def _load_model_and_image_preprocessor_from_checkpoint(self, weights_only: bool = True) \
+            -> Tuple[torch.nn.Module, Compose]:
         """Load the model and image preprocessor from a checkpoint file.
 
         The checkpoint file can be provided through a URL or a model_location object.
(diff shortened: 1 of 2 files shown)


tests/unit_tests/marqo/inference/native_inference/test_handling_corrupt_files_open_clip.py
```diff
@@ -106,6 +106,29 @@ def test_file_removal_failure_handling(self, mock_os_remove, mock_create_model_a
                 # Reset the mock
                 mock_os_remove.reset_mock()
 
+    @patch('open_clip.create_model', autospec=True)
+    @patch('marqo.inference.native_inference.embedding_models.open_clip_model.download_model', autospec=True)
+    def test_handling_unpickling_error_with_weights_only_true(self, mock_download, mock_create_model):
+        # Setup
+        mock_download.return_value = self.dummpy_corrupted_file
```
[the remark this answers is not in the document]
wanliAlex: @papa99do Let's solve this nit.

wanliAlex (changes requested):

src/marqo/inference/native_inference/embedding_models/open_clip_model.py
```diff
@@ -225,6 +218,34 @@ def _load_model_and_image_preprocessor_from_checkpoint(self) -> Tuple[torch.nn.M
...
+        try:
+            return open_clip.create_model(
+                model_name=self.model_properties.name,
+                jit=self.model_properties.jit,
+                pretrained=self.model_path,
+                precision=self.model_properties.precision,
+                device=self.device,
+                cache_dir=ModelCache.clip_cache_path,
+            )
+        except Exception as e:
```
wanliAlex: why don't we `except UnpicklingError` directly here?

wanliAlex (commented):
LGTM

src/marqo/inference/native_inference/embedding_models/open_clip_model.py
papa99do: good point. I'll change this

wanliAlex (approved):
LGTM

PR #1244, opened 2025-06-18
Sort and relevance cut-off Part 2

wanliAlex:
## Change Summary

This is the second part of the sort feature that includes the implementation of the sort functionality. 
This PR includes the following changes:
This pull request introduces enhancements to the hybrid search functionality, focusing on implementing support for sorting by fields (`sortBy`) and refining post-processing logic. Below are the key changes grouped by theme.

### Hybrid Search Enhancements:
* Introduced a new `SortField` class to define field-specific sorting configurations (`field_name`, `order`, `missing`) in `HybridSearcher` (`vespa/src/main/java/ai/marqo/search/HybridSearcher.java`).
* Added support for `sortBy` parameters (`fields`, `sortDepth`, `sortCandidates`) in the query processing pipeline. These parameters are used to configure sorting behavior during hybrid search (`vespa/src/main/java/ai/marqo/search/HybridSearcher.java`) [[1]](diffhunk://#diff-da6475dc2a9b7d803bba658add780db7404078c715c3e13d2cd93ad95b2f4077R81-R89) [[2]](diffhunk://#diff-da6475dc2a9b7d803bba658add780db7404078c715c3e13d2cd93ad95b2f4077R132-R137).
* Implemented a `postProcessBySort` method to sort search hits based on specified `sortBy` fields, with support for null handling and relevance as a tie-breaker. This method also resets relevance scores to reflect the new order (`vespa/src/main/java/ai/marqo/search/HybridSearcher.java`).
* Updated the main search logic to conditionally apply sorting or global score modifiers based on query parameters, ensuring compatibility with existing features (`vespa/src/main/java/ai/marqo/search/HybridSearcher.java`).

## Checklist

- [*] Tests have been added for changes
- [n/a] Documentation has been updated
- [n/a] Breaking changes are clearly identified
- [n/a] Python client changes linked or N/A

### For new field types:


vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -184,10 +204,33 @@ public Result search(Query query, Execution execution) {
...
+        Tensor queryAddWeightsGlobal =
+                extractTensorRankFeature(query, addQueryWrapper(QUERY_INPUT_ADD_WEIGHTS_GLOBAL));
+        if (sortByFields != null) {
+            // If sortBy is set, we will sort the hits after post-processing
+            processedHits =
+                    postProcessBySort(
+                            hitsForPostProcessing, sortByFields, sortBySortDepth, limit, offset);
+            processedHits.setField("marqo__sortCandidates", hitsForPostProcessing.size());
+        } else if ((queryMultWeightsGlobal != null && !queryMultWeightsGlobal.isEmpty())
+                || (queryAddWeightsGlobal != null && !queryAddWeightsGlobal.isEmpty())) {
```
papa99do: This logic does not match the previous one. Why do we need to add this branch here?

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -230,6 +273,81 @@ public Result search(Query query, Execution execution) {
...
     }
 
+    HitGroup postProcessBySort(
+            HitGroup hitsForPostProcessing,
+            String sortByFields,
+            Integer sortBySortDepth,
+            Integer limit,
+            Integer offset) {
+
+        ObjectMapper mapper = new ObjectMapper();
```
papa99do: There's better way to use jackson's thread-safe ObjectReader, see https://stackoverflow.com/questions/35335646/how-correctly-and-thread-safe-reuse-jackson-objectreader for details. We can declare the ObjectReader as a instance property.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -230,6 +273,81 @@ public Result search(Query query, Execution execution) {
...
+        List<SortField> parsedSortByFields;
+        try {
+            parsedSortByFields =
+                    mapper.readValue(sortByFields, new TypeReference<List<SortField>>() {});
+        } catch (Exception e) {
```
papa99do: Be more specific about what exception we catch. Here, we probably just want to catch IOException.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -230,6 +273,81 @@ public Result search(Query query, Execution execution) {
...
+            throw new RuntimeException(
```
papa99do, line 320: nit. we'll need to change the way we handle exception in CustomSearcher. Ideally, we should define some specific exceptions instead of just a general runtime exception. This is similar to throw an InternalError for everything in python app, which is not a good practice. We can leave this to a follow-up refactoring. 
RuntimeException will cause a 500 error response of Vespa, is this OK? We have validation logic in Marqo, right?

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -115,6 +129,12 @@ public Result search(Query query, Execution execution) {
         }
         // --- End facets subquery handling ---
 
+        // --- Update the query limit if sort is used
+        if (sortByFields != null && !sortByFields.isEmpty()) {
```
papa99do: nit. there's a helper method you can use to replace this: `!Strings.isNullOrEmpty(sortByFields)`

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -230,6 +273,81 @@ public Result search(Query query, Execution execution) {
...
+                    (sortComparator == null)
+                            ? fieldComparator
+                            : sortComparator.thenComparing(fieldComparator);
+        }
+
+        // Append relevance as final tie-breaker (always descending)
+        Comparator<Hit> relevanceTie =
+                Comparator.comparingDouble((Hit h) -> h.getRelevance().getScore()).reversed();
+        // always combine with relevance tie-break
+        sortComparator = sortComparator.thenComparing(relevanceTie);
```
papa99do: This piece of logic can be extracted to an  HitOrderer implementation. This might help with the unit testing.  see `HitSortOrderer` as an example.

papa99do: Ignore this. I realised that we are only sorting until the SortDepth. We could implement an orderer that does sort until depth, but leave it to the refactoring.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -230,6 +273,81 @@ public Result search(Query query, Execution execution) {
...
+        Comparator<Hit> sortComparator = null;
+        for (int i = 0; i < parsedSortByFields.size(); i++) {
+            final int idx = i;
+            SortField sf = parsedSortByFields.get(i);
+            Function<Hit, Double> keyExtractor =
+                    hit -> {
+                        FeatureData mf = (FeatureData) hit.getField("matchfeatures");
+                        if (mf == null) return null;
+                        double v = mf.getDouble("sort_field_value_" + idx);
+                        return (v == -1e50) ? null : v;
```
papa99do: `-1e50` is a magic number, extract it as a constant

vespa/src/test/java/ai/marqo/search/HybridSearcherTest.java
```diff
@@ -620,4 +623,401 @@ void shouldCreateProperFacetQueries() {
                             "lexical yql");
         }
     }
+
+    /**
+     * Test for sortBy feature in HybridSearcher.
+     */
+    @Nested
```
papa99do: I think it's OK to have this test in a separate file given it's size.  This is the test cases I think we need to cover for postProcessBySort method (can leverage parameterised tests):

- [x]  sortJson parsing logic - happy path - with 1, 2, 3 sort fields, various missing and order value)
- [x] sortJson parsing logic - corner case - parsing error (IOException)
- [x] sortJson parsing logic - corner case - 0, or 3+ sort fields
- [x] sortJson parsing logic - corner case - unsupported missing or order value

- [x] comparator - happy path - 1 sort field, test last/first, asc/desc combination
- [x] comparator - happy path - 2 sort fields, with first field fixed, test last/first, asc/desc combination of the second field
- [x] comparator - happy path - 3 sort fields, with first two fields fixed, test last/first, asc/desc combination of the third field
- [x] comparator - happy path - 3 sort fields, test the relevance tie breaker. 
- [ ] comparator - corner case - missing field in matchfeatures (is this possible?)

- [x] sortBySortDepth - null/smaller than hit size/larger than hit size/equal to hit size
- [x] sortBySortDepth - corner case, 0 / less than 0 (might not need since python has validated this, but we probably should assert) 

- [x] offset and limit - make sure the pagination works properly, only the trimmed result are returned in the hit group.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -47,6 +46,12 @@ public class HybridSearcher extends Searcher {
     private static String MARQO_SEARCH_METHOD_TENSOR = "tensor";
     private List<String> STANDARD_SEARCH_TYPES = new ArrayList<>();
 
+    private static class SortField {
+        public String field_name;
+        public String order;
+        public String missing;
+    }
+
```
papa99do: - Consider using java record. 
> Records are immutable data classes that require only the type and name of fields.
- Consider using Enum for order and missing fields to make sure only supported value are parsed
- Seems the field_name field isn't used anywhere in the code. Is this expected? 
- If we'd like to keep field_name, consider using camel case `fieldName` with an jackson annotation: `@JsonProperty("field_name")`

tests/integ_tests/tensor_search/integ_tests/test_search_sort_by_feature.py
```diff
@@ -1,10 +1,142 @@
...
 from marqo.core.models.marqo_index import *
 from marqo.tensor_search.api import search
 from marqo.tensor_search.enums import SearchMethod
+from tests.integ_tests.marqo_test import MarqoTestCase
+from marqo.core.exceptions import UnsupportedFeatureError
+from unittest.mock import patch, MagicMock
+from marqo.core.models.marqo_index import MarqoIndex
+
+
+class TestSearchSortByFeature(MarqoTestCase):
```
papa99do, line 14: nit. since this is an integration test, we can use real index here. just specify the marqo_version when you create it.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: done

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: Revert the changes. We use the old postProcess when sort is not provided.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: done

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: done

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: done

vespa/src/test/java/ai/marqo/search/HybridSearcherTest.java
wanliAlex: Added tests for sortJson Parsing.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: done

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: added todo.

vespa/src/test/java/ai/marqo/search/HybridSearcherTest.java
wanliAlex: SortBySortDepth can be null, and all the retrievals will be sorted. Add a test to ensure it must be greater than or equal to 1.

wanliAlex: > comparator - corner case - missing field in matchfeatures (is this possible?)

 If the field is missing, we return the magic number.

tests/integ_tests/tensor_search/integ_tests/test_search_sort_by_feature.py
wanliAlex: I lean towards using the mock instead of creating real indexes, as I feel this is more reliable and modifying the index version is a bit hacky, especially for legacy unstructured indexes. Let me know your thoughts.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -47,6 +51,47 @@ public class HybridSearcher extends Searcher {
     private static String MARQO_SEARCH_METHOD_TENSOR = "tensor";
     private List<String> STANDARD_SEARCH_TYPES = new ArrayList<>();
 
+    // Thread-safe ObjectReader for parsing SortField JSON
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+    private static final ObjectReader SORT_FIELD_READER =
+            OBJECT_MAPPER.readerFor(new TypeReference<List<SortField>>() {});
+
+    // Magic number for missing sort field values
+    private static final double MISSING_SORT_VALUE_SENTINEL = -1e50;
```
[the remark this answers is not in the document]
wanliAlex: Done

vespa/src/test/java/ai/marqo/search/HybridSearcherTest.java
```diff
@@ -3,17 +3,19 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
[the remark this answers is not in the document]
wanliAlex: Done

tests/compatibility_tests/conftest.py
```diff
@@ -27,7 +27,7 @@ def pytest_collection_modifyitems(config, items):
                 logger.debug(f"Testcase: {item.name} marked with skip_marqo_version: {skip_marqo_versions}. Skipping.")
                 item.add_marker(pytest.mark.skip(reason=f"Testcase: {item.name} marked with skip_marqo_version: {skip_marqo_versions}. Skipping."))
                 continue
-        elif test_case_version_marker:
+        if test_case_version_marker:
```
papa99do, line 30: is there any difference?

tests/compatibility_tests/search/unstructured/test_hybrid_search_unstructured_sort_by_feature.py
```diff
@@ -0,0 +1,213 @@
+import traceback
+from inspect import trace
+
+import pytest
+
+from tests.compatibility_tests.base_test_case.base_compatibility_test import BaseCompatibilityTestCase
+
+@pytest.mark.marqo_version('2.11.0')
```
papa99do, line 8: 2.21.0 or 2.16.0? 
Will this feature work for all indexes using the 2_16 schema template?

tests/integ_tests/tensor_search/integ_tests/test_search_sort_by_feature.py
```diff
@@ -743,4 +895,150 @@ def test_sort_by_when_fields_do_not_exist_three_fields(self):
...
+        # Other documents should be sorted by field3 values in descending order
+        self.assertEqual(ids[-1], '7', "Document with missing sort field should be last")
+        
+        # Verify that documents with sort field values come before missing ones
+        doc_7_pos = ids.index('7')
+        for i in range(doc_7_pos):
+            # All documents before position of doc 7 should have the sort field
+            doc_id = ids[i]
+            # Documents 0,1,2,3,4,5,6,8,9 should all have sort_field_3 and come before doc 7
+            self.assertIn(doc_id, ['0', '1', '2', '3', '4', '5', '6', '8', '9'])
```
papa99do: This does not actually test all ids are before doc_7, it actually does not test anything. I think the previous assertion is good enough to test 7 is the last. 
Also , can we add another doc with missing field3 to test tie breaking for missing fields?

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -46,10 +51,47 @@ public class HybridSearcher extends Searcher {
...
+
+    /** Sort order enum for better type safety */
+    private enum SortOrder {
+        ASC,
+        DESC;
+
+        @JsonCreator
+        public static SortOrder fromString(String value) {
+            if (value == null) return ASC; // default
+            return "desc".equalsIgnoreCase(value) ? DESC : ASC;
```
papa99do: A good practice of handling input is `validate input rigorously and fail fast on bad data.` 
So it would be better to let invalid input error out instead of falling back to default. 
`return SortOrder.valueOf(value.toUpperCase(Locale.ROOT));` . This will throw IllegalArgumentException if a value other than `desc` or `asc` is provided.

vespa/src/test/java/ai/marqo/search/SortByTest.java
```diff
@@ -0,0 +1,746 @@
...
+class SortByTest {
+
+    /**
+     * Test that verifies sorting of results based on a single sort field.
+     */
+    HitGroup helpGenerateHitGroupWithOnlySortFieldValue0() {
+        FeatureData f1 = mock(FeatureData.class);
+        when(f1.getDouble("sort_field_value_0")).thenReturn(-1e50);
+        Hit doc1 = new Hit("doc1", 0.55);
+        doc1.setField("matchfeatures", f1);
```
papa99do, line 35: nit. consider extract a method to reduce dup code

vespa/src/test/java/ai/marqo/search/SortByTest.java
```diff
@@ -0,0 +1,746 @@
...
+
+            for (String sortJson : testCases) {
+                // Should not throw exception for any of these cases
+                HitGroup result = searcher.postProcessBySort(hits, sortJson, null, 10, 0);
+                assertThat((Object) result).isNotNull();
+            }
+        }
+
+        @Test
+        void shouldThrowExceptionForNullOrderAndMissing() {
```
papa99do, line 448: is null is provided, we'll just use the default, right?

vespa/src/test/java/ai/marqo/search/SortByTest.java
```diff
@@ -0,0 +1,746 @@
...
+            // Test unsupported order values - should default to ASC
+            String[] unsupportedOrders = {"ascending", "descending", "up", "down", "invalid", ""};
+
+            for (String order : unsupportedOrders) {
+                String sortJson =
+                        "[{\"field_name\":\"f1\",\"order\":\""
+                                + order
+                                + "\",\"missing\":\"last\"}]";
+
+                // Should not throw exception, should default to ASC
```
papa99do: this is not a good practice. we should validate the input rigorously and fail fast

vespa/src/test/java/ai/marqo/search/SortByTest.java
```diff
@@ -0,0 +1,746 @@
...
+            // Test unsupported missing values - should default to FIRST
+            String[] unsupportedMissing = {"top", "bottom", "start", "end", "invalid", ""};
+
+            for (String missing : unsupportedMissing) {
+                String sortJson =
+                        "[{\"field_name\":\"f1\",\"order\":\"asc\",\"missing\":\""
+                                + missing
+                                + "\"}]";
+
+                // Should not throw exception, should default to FIRST
```
papa99do: same here

tests/compatibility_tests/conftest.py
wanliAlex: This fixes a bug that if `skip_marqo_version` exists, the` test_case_version_marker` check is skipped.

tests/compatibility_tests/search/unstructured/test_hybrid_search_unstructured_sort_by_feature.py
wanliAlex: This marker actually means the `minimum_marqo_version` to test on. It means this test will run on any version greater than or equal to `2.11.0`

vespa/src/test/java/ai/marqo/search/SortByTest.java
```diff
@@ -0,0 +1,746 @@
...
+                    "[{\"field_name\":\"f1\",\"order\":\"asc\",\"missing\":\"last\"},"
+                            + "{\"field_name\":\"f2\",\"order\":\"desc\",\"missing\":\"first\"},"
+                            + "{\"field_name\":\"f3\",\"order\":\"asc\",\"missing\":\"last\"}]";
+            HitGroup result3 = searcher.postProcessBySort(hits, threeFields, null, 10, 0);
+            assertThat((Object) result3).isNotNull();
+            assertThat(result3.asList()).hasSize(1);
+        }
+
+        @Test
+        void shouldValidateFieldsInCorrectOrder() {
```
papa99do: not sure if this test case tests what it claims to test :D . It looks like it's still testing missing fields.

tests/integ_tests/tensor_search/integ_tests/test_search_sort_by_feature.py
wanliAlex: 1. Made doc6 also missing field 3, and added a test on this relevance tie between doc 7 and doc 6.
2. Added a test for a sort on a non-existent field and ensured all results are sorted by relevance.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: Thanks! Made the change. We don't have to catch `IllegalArgumentException` as it will caught later and converted to RuntimeError

vespa/src/test/java/ai/marqo/search/SortByTest.java
wanliAlex: Revised this test

vespa/src/test/java/ai/marqo/search/SortByTest.java
wanliAlex: Revised this test

vespa/src/test/java/ai/marqo/search/SortByTest.java
wanliAlex: updated the test name to `shouldThrowWhenSortFieldJsonIsMissingOrderOrFieldNameAtCorrectIndex`

papa99do (dismissed):
LGTM

papa99do (approved):

PR #1246, opened 2025-06-23
Sort And Relevance Cuffoff Part 3 - Add Relevance Cutoff Feature

wanliAlex:
## Change Summary

This PR implements the relevance-cutoff functionality in Marqo Python and Java. Relevance Cutoff can be used to stub out irrelevant documents in the retrieval stage. Detailed changes are shown as below:

### Sorting Enhancements:
* Renamed `sortCandidates` to `minSortCandidates` across models, validation logic, and query construction for better clarity. (`src/marqo/tensor_search/models/sort_by_model.py`,
* Updated all related test cases to use `minSortCandidates` instead of `sortCandidates`. 

### Relevance Cutoff Implementation:
* Implement 3 different relevance cutoff methods in Java, namely `"relative_max_score"`, `"gap_detection"`, and `"mean_std_dev"`,   
    1. Relative Max Score (relative_max_score)
      - Algorithm: Uses a dynamic threshold based on the top search result's score
      - Implementation:
        - Takes the highest scoring result's score
        - Multiplies it by a relativeScoreFactor parameter (0-1 range)
        - Keeps all results with scores ≥ this threshold
     2. Gap Detection (gap_detection)
      - Algorithm: Finds the largest score gap between consecutive results (elbow detection)
      - Implementation:
        - Iterates through sorted results comparing adjacent scores
        - Finds the position with the maximum score difference (delta)
        - Cuts off at the position after the largest gap
      - Parameter: None required (no cutoff parameter)
     3. Mean Standard Deviation (mean_std_dev)
      - Algorithm: Statistical approach using mean and standard deviation
      - Implementation:
        - Calculates mean score across all probe results
        - Calculates population standard deviation
        - Sets threshold as: mean + (stdDev × stdDevFactor)
        - Keeps results with scores ≥ threshold

* New Response Metadata
  All relevance cutoff searches return additional metadata:
  - `_relevantCandidates`: Number of results passing relevance filter
  - `_probeCandidates`: Total candidates analyzed during probe phase

* New API Parameter: `relevanceCutoff`

  API Structure
 ```python
  {
    "relevanceCutoff": {
      "method": "relative_max_score|gap_detection|mean_std_dev",
      "parameters": {
        "relativeScoreFactor": 0.6,  # for relative_max_score
        "stdDevFactor": 0.3         # for mean_std_dev
      }
    }
  }
 ```

* Example Usage
```python
  # Just relevance cut-off
  response = client.index("my-index").search(
      q="machine learning AI",
      search_method="HYBRID",
      relevance_cutoff={
          "method": "relative_max_score",
          "parameters": {"relativeScoreFactor": 0.7}
      },
      limit=10
  )
```

```python
  # Search with both relevance cut-off and sorting
  response = client.index("my-index").search(
      q="machine learning artificial intelligence",
      search_method="HYBRID",
      relevance_cutoff={
          "method": "relative_max_score",
          "parameters": {"relativeScoreFactor": 0.7}
      },
      sort_by={
          "fields": [
              {"fieldName": "sort_value", "order": "desc", "missing": "last"}
          ],
          "minSortCandidates": 10  # Optional: override relevance cutoff if needed
      },
  )
```

## Related Jira Ticket

[ticket link](https://s2search.atlassian.net/jira/polaris/projects/MOSD/ideas/view/4863474?selectedIssue=MOSD-324&issueViewLayout=sidebar&issueViewSection=overview&atlOrigin=eyJpIjoiMDI0ZmJiZDhhNjE3NGZjMmFiNGIwYTEzN2JjYWM5YjkiLCJwIjoiaiJ9)

## Checklist

- [*] Tests have been added for changes
- [n/a] Documentation has been updated
- [n/a] Breaking changes are clearly identified
- [ n/a Python client changes linked or N/A

### For new field types:


src/marqo/core/constants.py
```diff
@@ -23,7 +23,7 @@
 MARQO_CUSTOM_VECTOR_NORMALIZATION_MINIMUM_VERSION = semver.VersionInfo.parse('2.13.0')
 MARQO_GLOBAL_SCORE_MODIFIERS_MINIMUM_VERSION = semver.VersionInfo.parse('2.15.0')
 MARQO_RERANK_DEPTH_MINIMUM_VERSION = semver.VersionInfo.parse('2.15.0')
-MARQO_SORT_BY_MINIMUM_VERSION = semver.VersionInfo.parse('2.21.0')
+MARQO_SORT_BY_RELEVANCE_CUTOFF_MINIMUM_VERSION = semver.VersionInfo.parse('2.21.0')
```
papa99do: 2.22.0? since 2.21.0 does not have cutoff impl in app package yet.

src/marqo/core/models/marqo_index.py
```diff
@@ -638,13 +638,13 @@ def index_supports_language(self) -> bool:
             lambda: self.parsed_marqo_version() >= constants.MARQO_LANGUAGE_MINIMUM_VERSION)
 
     @property
-    def index_supports_sorty_by(self) -> bool:
+    def index_supports_sorty_by_relevance_cutoff(self) -> bool:
```
papa99do: Do we need to separate relevance_cutoff feature from sort_by feature? Is it possible that user only want to use sort_by but not cutoff? 

nit. typo in the method name, sorty -> sort

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -676,7 +676,7 @@ def _to_vespa_hybrid_query(self, marqo_query: MarqoHybridQuery) -> Dict[str, Any
         if marqo_query.sort_by:
             query["marqo__hybrid.sortBy.fields"] = [field.dict() for field in marqo_query.sort_by.fields]
             query["marqo__hybrid.sortBy.sortDepth"] = marqo_query.sort_by.sort_depth
-            query["marqo__hybrid.sortBy.sortCandidates"] = marqo_query.sort_by.sort_candidates
+            query["marqo__hybrid.sortBy.minSortCandidates"] = marqo_query.sort_by.min_sort_candidates
```
papa99do, line 679: Have you written any compatibility tests using this parameter already? If so, the test might fail.

src/marqo/tensor_search/models/relevance_cutoff_model.py
```diff
@@ -12,11 +12,11 @@ class RelevanceCutoffMethod(str, Enum):
 
 
 class RelativeMaxScoreParameters(StrictBaseModel):
-    relative_score_factor: float = Field(..., gt=0, le=1, alias="relativeScoreFactor")
+    relative_score_factor: float = Field(..., ge=0, le=1, alias="relativeScoreFactor")
 
 
 class MeanStdParameters(StrictBaseModel):
-    std_dev_factor: float = Field(..., gt=0, alias="stdDevFactor")
+    std_dev_factor: float = Field(..., alias="stdDevFactor")
```
papa99do, line 19: why do we loosen the validation to allow negative or zero std_dev_factor?

src/marqo/tensor_search/models/relevance_cutoff_model.py
```diff
@@ -12,11 +12,11 @@ class RelevanceCutoffMethod(str, Enum):
 
 
 class RelativeMaxScoreParameters(StrictBaseModel):
-    relative_score_factor: float = Field(..., gt=0, le=1, alias="relativeScoreFactor")
+    relative_score_factor: float = Field(..., ge=0, le=1, alias="relativeScoreFactor")
```
papa99do, line 15: why do we loosen the validation to allow relative_score_factor to be 0?

tests/api_tests/v1/tests/api_tests/test_sort_by_feature.py
```diff
@@ -60,7 +58,7 @@ def test_sort_by_is_blocked_by_structured_index(self):
             )
 
         self.assertIn(
-            "feature is only supported for unstructured indexes created with Marqo version",
+            "are only supported for unstructured indexes created with Marqo version",
```
papa99do: nit. add the version number in the assertion

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -301,10 +346,91 @@ public Result search(Query query, Execution execution) {
...
+        Integer newTensorTargetHits = null;
+
+        if (isRelevanceCutoffEnabled && isSortByEnabled) {
+            // Both sort and relevance cut-off enabled: use the higher one as new limits.
+            // Note that sortByMinSortCandidates is guaranteed to be larger than limit+offset so no
+            // check is required
+            newHits = Math.max(relevanceCandidates, sortByMinSortCandidates);
+            if (currentTensorTargetHits != null) {
+                newTensorTargetHits = Math.max(newHits, currentTensorTargetHits);
+            }
```
papa99do, line 446: This check is common for all 3 scenarios, can be extracted outside and after this logic.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -680,6 +879,78 @@ void addFieldToRankFeatures(Cell cell, Query query, boolean verbose) {
...
+        }
+    }
+
+    private String overwriteTargetHits(String yql, int newTargetHits) {
+        // Validate input
+        if (newTargetHits < 0) {
+            throw new RuntimeException("targetHits value must be positive, got: " + newTargetHits);
+        }
+
+        if (newTargetHits == 0) {
```
papa99do, line 1005: why would the newTargetHits be 0? it has to be >= the original value, right?

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -680,6 +879,78 @@ void addFieldToRankFeatures(Cell cell, Query query, boolean verbose) {
...
+        // Count hnsw.exploreAdditionalHits occurrences
+        long hnswCount = HNSW_EXPLORE_ADDITIONAL_HITS_PATTERN.matcher(yql).results().count();
+
+        if (hnswCount == 0) {
+            throw new RuntimeException(
+                    "YQL does not contain hnsw.exploreAdditionalHits clause, but targetHits is"
+                            + " present. Both parameters must be present together.");
+        }
+
+        if (targetHitsCount != hnswCount) {
```
papa99do, line 1027: why would targetHits equal to hnswCount? `targetHits = 2000 - hnswCount`, right?

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -680,6 +879,78 @@ void addFieldToRankFeatures(Cell cell, Query query, boolean verbose) {
...
+                            + hnswCount
+                            + " hnsw.exploreAdditionalHits occurrences. Both must have the same"
+                            + " count.");
+        }
+
+        // Replace all targetHits occurrences
+        String updatedYql = TARGET_HITS_PATTERN.matcher(yql).replaceAll("$1" + newTargetHits);
+
+        // Also update hnsw.exploreAdditionalHits to 2000-newTargetHits for all occurrences
+        int newExploreAdditionalHits = 2000 - newTargetHits;
```
papa99do: ef_search is not a fixed value, we'll need to pass it in.
```
marqo_query.ef_search = marqo_query.ef_search if marqo_query.ef_search is not None else max(
            utils.read_env_vars_and_defaults_ints(EnvVars.MARQO_DEFAULT_EF_SEARCH),
            marqo_query.limit + marqo_query.offset
        )

        target_hits = min(marqo_query.limit + marqo_query.offset, marqo_query.ef_search)
        additional_hits = max(marqo_query.ef_search - (marqo_query.limit + marqo_query.offset), 0)
 ```

File: vespa/src/main/java/ai/marqo/search/HybridSearcher.java [hunk 11]
```diff
@@ -698,8 +969,9 @@ public Query createSubQuery(
      * @param retrievalMethod
      * @param rankingMethod
      * @param verbose
+     * @param exactQuery
      */
-    Query createSubQuery(
+    public Query createSubQuery(
```
papa99do: why do we need to make it public?

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -301,10 +346,91 @@ public Result search(Query query, Execution execution) {
...
+            return query;
+        }
+
+        if (relevanceCandidates == null && sortByMinSortCandidates == null) {
+            throw new RuntimeException(
+                    "Either relevanceCandidates or sortByMinSortCandidates must be provided");
+        }
+
+        sortByMinSortCandidates = (sortByMinSortCandidates == null) ? -1 : sortByMinSortCandidates;
+        relevanceCandidates = (relevanceCandidates == null) ? -1 : relevanceCandidates;
```
papa99do: Why do we need to set the default value to -1?

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -660,6 +834,31 @@ void raiseErrorIfPresent(Result resultLexical, Result resultTensor) {
...
+
+        // Overwrite the lexical score modifiers in the probe query
+        probeLexicalQuery
+                .getRanking()
+                .getFeatures()
+                .put("query(marqo__mult_weights_lexical)", Tensor.from("tensor(p{}):{}"));
+        probeLexicalQuery
+                .getRanking()
+                .getFeatures()
+                .put("query(marqo__add_weights_lexical)", Tensor.from("tensor(p{}):{}"));
```
papa99do, line 912: nit. but could we skip populating score modifier rank feature in the first place?

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -892,4 +1165,99 @@ HitGroup applyGlobalScoreModifiers(HitGroup hits, boolean verbose) {
...
+
+    /**
+     * Detects the cutoff count for relevance filtering based on the specified method
+     * @param probeCandidates The lexical search results to analyze
+     * @param cutoffMethodString The method to use for cutoff detection
+     * @param relevanceCutoffParameter The parameter for the cutoff method. It is unified across all methods.
+     * @param verbose Whether to log verbose information
+     * @return The number of relevant results to keep
+     */
+    private Integer detectCutoffCount(
```
papa99do: I'll review later with the tests. Are these method detailed in the design doc?

src/marqo/core/constants.py
wanliAlex: Will revert this change because the relevance cut-off is backwards compatible.

src/marqo/core/structured_vespa_index/structured_vespa_index.py
wanliAlex: Will revert this change because the relevance cut-off is backwards compatible.

src/marqo/core/constants.py
wanliAlex: Add tests in the backwards compatibility tests.

src/marqo/core/models/marqo_index.py
wanliAlex: Will revert this change because the relevance cut-off is backwards compatible.

src/marqo/core/structured_vespa_index/structured_vespa_index.py
wanliAlex: Add backwards compatibility tests that the index created on 2.21.0 can sort after migrating to 2.22.0

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: currentTensorTargetHits can be null if there is no tensorYQL.

There is a different case where `newTensorTargetHits = Math.min(newHits, currentTensorTargetHits);`

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -301,10 +346,91 @@ public Result search(Query query, Execution execution) {
...
+                newTensorTargetHits = Math.max(newHits, currentTensorTargetHits);
+            }
+        }
+
+        // Update query hits and offset
+        query.setHits(newHits);
+        query.setOffset(0);
+
+        // Update tensor YQL targetHits if it exists
+        if (currentTensorTargetHits != null && newTensorTargetHits != null) {
```
papa99do: only override if `currentTensorTargetHits != newTensorTargetHits`

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -892,4 +1165,99 @@ HitGroup applyGlobalScoreModifiers(HitGroup hits, boolean verbose) {
...
+                }
+                double meanScore = sum / lexicalHits.size();
+
+                // Calculate population standard deviation
+                double sumSquaredDiffs = 0.0;
+                for (Hit hit : lexicalHits) {
+                    double diff = hit.getRelevance().getScore() - meanScore;
+                    sumSquaredDiffs += diff * diff;
+                }
+                double stdScore = Math.sqrt(sumSquaredDiffs / lexicalHits.size());
```
papa99do: nit. consider using lib like [apache commons-statistics](https://commons.apache.org/proper/commons-statistics/userguide/index.html) to avoid re-invent the wheel. See if it's easy to convert scores of the hit to a list or stream to pass in to the stats api.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -892,4 +1165,99 @@ HitGroup applyGlobalScoreModifiers(HitGroup hits, boolean verbose) {
...
+                double threshold = meanScore + (stdScore * relevanceCutoffParameter);
+
+                // Count relevant results
+                int relevantCount = 0;
+                for (Hit hit : lexicalHits) {
+                    if (hit.getRelevance().getScore() >= threshold) {
+                        relevantCount++;
+                    }
+                }
+                return relevantCount;
```
papa99do: nit. Let's assume that the scores are already extracted into a list. since we know the scores are sorted in descending order, so a quicker way to find the count is to use binary search to find the index of first number that's smaller than the threshold. Something like this `int relevantCount = -Collections.binarySearch(listDesc, target, Comparator.reverseOrder()) - 1;` (need to verify the correctness).

vespa/src/test/java/ai/marqo/search/HybridSearcherTest.java
```diff
@@ -3,8 +3,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
papa99do: Revert this change? let's do not fold the imports. You can config this in Intellij. We'll use some tools later to check this as well.

vespa/src/test/java/ai/marqo/search/RelevanceCutoffTest.java
```diff
@@ -0,0 +1,558 @@
...
+
+            RuntimeException exception =
+                    assertThrows(
+                            RuntimeException.class,
+                            () -> callReadRelevanceCutoffParameter(query, "unknown_method"));
+            assertThat(exception.getMessage())
+                    .contains("Unknown relevance cutoff method: unknown_method");
+        }
+
+        private Double callReadRelevanceCutoffParameter(Query query, String method) {
```
papa99do, line 100: This is not a recommended way to do unit test. Ideally, we should only test public method, which represents the API/interface of a `unit`. In our case, the code structure is not unit-test friendly. Since we decided to refactor it later, one workaround is to make the method visible at package-level, and mark it with `@VisibleForTesting` from guava lib. Later, we can do refactoring to extract the logic to a class that's unit-testable.

src/marqo/core/constants.py
farshidz: Sort by relevance cutoff reads oddly. Maybe SORT_AND_RELEVANCE_CUTOFF or just relevance cutoff or just sort?

farshidz: 2.22.0

src/marqo/core/models/marqo_index.py
farshidz: Same issue with the name. Sounds like you're sorting by relevance cutoff?

papa99do (changes requested):

src/marqo/core/structured_vespa_index/structured_vespa_index.py
farshidz: I don't think you need to worry about backwards compatibility with sort (2.21) because we never officially releaesed sort

src/marqo/tensor_search/models/api_models.py
```diff
@@ -355,32 +355,32 @@ def _validate_sort_by_cannot_be_used_with_global_score_modifiers(cls, values):
         return values
 
     @root_validator(pre=False)
-    def _set_sort_by_sortCandidates_parameters(cls, values):
-        """Set the value for sortCandidates in sortBy if it is not provided.
+    def _set_sort_by_min_sort_candidates_parameters(cls, values):
```
farshidz: This is just validation, right? Function name doesn't match others. Other func are validate... not set...

src/marqo/tensor_search/tensor_search.py
```diff
@@ -423,20 +423,21 @@ def search(config: Config, index_name: str, text: Optional[Union[str, dict, Cust
             f"This index was created with Marqo {marqo_index_version}."
         )
 
-    if sort_by:
+    if sort_by or relevance_cutoff:
         if not isinstance(marqo_index, SemiStructuredMarqoIndex):
```
farshidz: Note we've generally managed to keep index types out of search. Though the downside of moving these into index type classes is you'll error out after vectorization.

src/marqo/vespa/models/query_result.py
```diff
@@ -7,6 +7,8 @@
 class RootFields(BaseModel):
     total_count: Optional[int] = Field(None, alias='totalCount')
     sort_candidates: Optional[int] = Field(None, alias='marqo__sortCandidates')
+    relevant_candidates: Optional[int] = Field(None, alias='marqo__relevantCandidates')
```
farshidz: These classes modelled the Vespa API. I wonder if it's safe for us to add fields to this part of the response?

src/marqo/tensor_search/models/api_models.py
wanliAlex: It's a mixture of set and validation. It has this logic:
```python
if sort_by.min_sort_candidates is None:
    sort_by.min_sort_candidates = max(
        cls._DEFAULT_SORT_CANDIDATES_MULTIPLIER * values.get('limit'),
        values.get('offset') + values.get('limit')
    )
```

wanliAlex: rename it to `_validate_and_set_sort_by_min_sort_candidates_parameters`

src/marqo/tensor_search/tensor_search.py
wanliAlex: Moved the check into the `HybridSearch` class

src/marqo/vespa/models/query_result.py
wanliAlex: What are your thoughts here to transmit metadata from Vespa to Marqo? What safety concerns do you have?

tests/api_tests/v1/tests/api_tests/test_sort_by_feature.py
wanliAlex: Updated API tests. Add tests for structured index.

src/marqo/core/structured_vespa_index/structured_vespa_index.py
wanliAlex: I have set the minimum required version to 2.22.0 so we don't have to worry about any backwards compatibility issues.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: Added ef_search logic in Vespa

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: reverted change

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: The built-in method will fail if it contains multiple elements exactly equal to the threshold. So I implemented our own binary search.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: done

vespa/src/test/java/ai/marqo/search/HybridSearcherTest.java
wanliAlex: reverted

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: done

src/marqo/core/constants.py
wanliAlex: Done

wanliAlex: Done

src/marqo/core/models/marqo_index.py
wanliAlex: Change back to `index_supports_sort_by`

vespa/src/test/java/ai/marqo/search/RelevanceCutoffTest.java
wanliAlex: Done. I also made the changes to other private method calls.

src/marqo/core/search/hybrid_search.py
```diff
@@ -152,6 +152,23 @@ def search(
...
+                marqo_index_version < constants.MARQO_SORT_BY_MINIMUM_VERSION or
+                not isinstance(marqo_index, SemiStructuredMarqoIndex)
+        ):
+            raise core_exceptions.UnsupportedFeatureError(
+                f"The 'sortBy' features is only supported for unstructured indexes created "
+                f"with Marqo version {constants.MARQO_SORT_BY_MINIMUM_VERSION} or later "
+            )
+
+        if (relevance_cutoff and isinstance(marqo_index, UnstructuredMarqoIndex) and
+                not isinstance(marqo_index, SemiStructuredMarqoIndex)):
```
papa99do: I think you can do this: `marqo_index.type == IndexType.Unstructured`

src/marqo/tensor_search/tensor_search.py
```diff
@@ -422,21 +422,6 @@ def search(config: Config, index_name: str, text: Optional[Union[str, dict, Cust
...
-                f"{constants.MARQO_SORT_BY_MINIMUM_VERSION} or later. "
-                f"Your index is either a structured index or an old unstructured index"
-            )
-        if not marqo_index.index_supports_sorty_by:
-            raise core_exceptions.UnsupportedFeatureError(
-                f"The 'sortBy' feature is only supported for unstructured indexes created with Marqo version "
-                f"{constants.MARQO_SORT_BY_MINIMUM_VERSION} or later. "
-                f"This unstructured index was created with Marqo {marqo_index_version} "
-            )
-
```
papa99do, line 469: why is the validation logic removed?

vespa/pom.xml
```diff
@@ -26,6 +26,18 @@
             <artifactId>jackson-core</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-statistics-descriptive</artifactId>
+            <version>1.1</version>
+            <scope>provided</scope>
```
papa99do: how do you find out the lib with the specific version is provided by the Vespa runtime? BTW, if the lib is provided, we do not need to specify the version. But if it's not provided, we'll need to remove the `<scope>provided</scope>`

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -301,10 +353,94 @@ public Result search(Query query, Execution execution) {
...
+                newTensorTargetHits = Math.max(newHits, currentTensorTargetHits);
+            }
+        }
+
+        // Update query hits and offset
+        query.setHits(newHits);
+        query.setOffset(0);
+
+        // Update tensor YQL targetHits if it exists
+        if (!Objects.equals(currentTensorTargetHits, newTensorTargetHits)) {
```
papa99do: I think you still need to test `currentTensorTargetHits` is not null.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -680,6 +892,96 @@ void addFieldToRankFeatures(Cell cell, Query query, boolean verbose) {
...
+                            + " targetHits occurrences but "
+                            + hnswCount
+                            + " hnsw.exploreAdditionalHits occurrences. Both must have the same"
+                            + " count.");
+        }
+
+        // Replace all targetHits occurrences
+        String updatedYql = TARGET_HITS_PATTERN.matcher(yql).replaceAll("$1" + newTargetHits);
+
+        // Also update hnsw.exploreAdditionalHits to 2000-newTargetHits for all occurrences
```
papa99do: nit. comment still says 2000. BTW, what if user specifies a efserach less than the newTargetHits? Should we use `max(efSearch - newTargetHits, 0)`?

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -92,8 +98,23 @@ public static MissingOrder fromString(String value) {
...
 
+    private enum RelevanceCutoffMethod {
+        RELATIVE_MAX_SCORE,
+        MEAN_STD_DEV,
+        GAP_DETECTION;
+
+        @JsonCreator
+        public static RelevanceCutoffMethod fromString(String value) {
+            return RelevanceCutoffMethod.valueOf(value.toUpperCase(Locale.ROOT));
+        }
```
papa99do, line 132: where is this method used?

src/marqo/core/search/hybrid_search.py
```diff
@@ -152,6 +152,23 @@ def search(
...
+            # Legacy unstructured indexes do not support relevance cutoff
+            raise core_exceptions.UnsupportedFeatureError(
```
farshidz, line 167: This isn't erroring out for structured indexes due the second condition

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -892,4 +1196,92 @@ HitGroup applyGlobalScoreModifiers(HitGroup hits, boolean verbose) {
...
+                        maxDelta = delta;
+                        bestIndex = i + 1;
+                    }
+                }
+                return bestIndex;
+            }
+            case MEAN_STD_DEV -> {
+                logIfVerbose("Using normalFit method for relevance cutoff", verbose);
+                double mean = Mean.of(probeLexicalScores).getAsDouble();
+                double stdDev = StandardDeviation.of(probeLexicalScores).getAsDouble();
```
papa99do: nit. you can save one pass of the array by using this:
```
DoubleStatistics stats = DoubleStatistics.of(
    EnumSet.of(Statistic.MEAN, Statistic. STANDARD_DEVIATION), data);
stats.getAsDouble(Statistic.MEAN);   
stats.getAsDouble(Statistic.STANDARD_DEVIATION);  
```

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -892,4 +1196,92 @@ HitGroup applyGlobalScoreModifiers(HitGroup hits, boolean verbose) {
...
+        switch (cutoffMethod) {
+            case GAP_DETECTION -> {
+                logIfVerbose("Using gapDetection method for relevance cutoff", verbose);
+                // Find the elbow point in the lexical hits
+                double maxDelta = -1.0;
+                int bestIndex = lexicalHits.size(); // default: keep all
+                for (int i = 0; i < lexicalHits.size() - 1; i++) {
+                    double score1 = lexicalHits.get(i).getRelevance().getScore();
+                    double score2 = lexicalHits.get(i + 1).getRelevance().getScore();
+                    double delta = score1 - score2;
```
papa99do, line 1305: nit. values are already in the probeLexicalScores array, so can be just `probeLexicalScores[i] - probeLexicalScores[i+1]`

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -892,4 +1196,92 @@ HitGroup applyGlobalScoreModifiers(HitGroup hits, boolean verbose) {
...
+     * Returns the number of elements in a descending-sorted array
+     * that are greater than or equal to the given threshold.
+     */
+    @VisibleForTesting
+    static int countGreaterOrEqual(double[] descSorted, double threshold) {
+        int low = 0, high = descSorted.length;
+        while (low < high) {
+            int mid = (low + high) >>> 1;
+            if (descSorted[mid] >= threshold) {
+                // still ≥ threshold → move low up
```
papa99do: nit. use only ascii chars in comments.

tests/api_tests/v1/tests/api_tests/test_relevance_cutoff_feature.py
```diff
@@ -0,0 +1,616 @@
...
+        test_indexes = [self.structured_index_name, self.unstructured_index_name]
+        for index_name in test_indexes:
+            if index_name == self.structured_index_name:
+                tensor_fields = None
+            else:
+                tensor_fields = ["content"]
+            with self.subTest(index_name):
+                docs = [
+                    {"_id": "h1", "content": "Machine learning algorithms in artificial intelligence", "score": 1.0},
+                    {"_id": "h2", "content": "Artificial intelligence relies on machine learning algorithms", "score": 0.9},
```
papa99do: add a doc with score of 0.6 as an edge case.

tests/api_tests/v1/tests/api_tests/test_relevance_cutoff_feature.py
```diff
@@ -0,0 +1,616 @@
...
+                    relevance_cutoff={
+                        "method": "gap_detection"
+                    },
+                    limit=10
+                )
+
+                # Should detect gap and filter appropriately
+                self.assertIn("_relevantCandidates", response)
+                self.assertLess(response["_relevantCandidates"], 6)
+                self.assertGreater(response["_relevantCandidates"], 0)
```
papa99do: assert hits?

tests/api_tests/v1/tests/api_tests/test_relevance_cutoff_feature.py
```diff
@@ -0,0 +1,616 @@
...
+
+                # Should filter out low relevance documents
+                self.assertIn("_relevantCandidates", response)
+                self.assertLess(response["_relevantCandidates"], len(docs))
+                self.assertGreater(response["_relevantCandidates"], 0)
+
+                # High relevance docs should be present
+                result_ids = [hit["_id"] for hit in response["hits"]]
+                self.assertIn("h1", result_ids)
+                self.assertIn("h2", result_ids)
```
papa99do: what about other records? do we expect them in the result? why don't we just assert the whole result_ids list?

tests/api_tests/v1/tests/api_tests/test_relevance_cutoff_feature.py
```diff
@@ -0,0 +1,616 @@
...
+        
+        # Should filter out low relevance docs
+        ids_with_cutoff = [hit["_id"] for hit in response_with_cutoff["hits"]]
+        self.assertNotIn("l1", ids_with_cutoff)  # Should be filtered out
+        self.assertNotIn("l2", ids_with_cutoff)  # Should be filtered out
+        
+        # Should still include high relevance docs
+        self.assertIn("h1", ids_with_cutoff)
+        self.assertIn("h2", ids_with_cutoff)
+        self.assertIn("h3", ids_with_cutoff)
```
papa99do: nit. simple way is to assert the ids_with_cutoff equals to the expected

tests/api_tests/v1/tests/api_tests/test_relevance_cutoff_feature.py
```diff
@@ -0,0 +1,616 @@
...
+            {"_id": "h1", "content": "Machine learning algorithms in artificial intelligence", "sort_value": 8.1},
+            {"_id": "h2", "content": "Artificial intelligence relies on machine learning algorithms",
+             "sort_value": 9.2},
+            {"_id": "h3", "content": "Researchers develop artificial intelligence machine learning", "sort_value": 7.4},
+            {"_id": "l1", "content": "Engineers use machine tools for cutting", "sort_value": 10.0},
+            # High sort, low relevance
+            {"_id": "l2", "content": "Bright morning sunlight streams through", "sort_value": 9.5},
+            # High sort, low relevance
+        ]
+        self.client.index(self.structured_index_name).add_documents(
```
papa99do: nit. no need to populate docs

tests/api_tests/v1/tests/api_tests/test_relevance_cutoff_feature.py
```diff
@@ -0,0 +1,616 @@
...
+        # Should have metadata
+        self.assertIn("_relevantCandidates", response)
+        self.assertIn("_sortCandidates", response)
+        
+        # minSortCandidates should override relevance filtering
+        self.assertEqual(response["_sortCandidates"], 5)
+        
+        # Should include low relevance docs due to override
+        ids = [hit["_id"] for hit in response["hits"]]
+        self.assertIn("l1", ids)  # Should be included due to minSortCandidates override
```
papa99do: should we verify the order of the ids list instead?

tests/api_tests/v1/tests/api_tests/test_relevance_cutoff_feature.py
```diff
@@ -0,0 +1,616 @@
...
+            q="machine learning artificial intelligence",
+            search_method="HYBRID",
+            sort_by={
+                "fields": [{"fieldName": "sort_value", "order": "desc", "missing": "last"}]
+            },
+            limit=5
+        )
+        ids_no_cutoff = [hit["_id"] for hit in response_no_cutoff["hits"]]
+        # Should include low relevance docs with high sort values
+        self.assertIn("l1", ids_no_cutoff[:3])  # Should be in top 3 due to high sort value
```
papa99do: why isn't it the top 1?

tests/api_tests/v1/tests/api_tests/test_relevance_cutoff_feature.py
papa99do, line 1: Why do we we only verify the metadata, but not the number and order of hits in some of the tests? Is the search result deterministic?

src/marqo/vespa/models/query_result.py
farshidz: Does the Vespa documentation have an opinion on this?

tests/integ_tests/tensor_search/integ_tests/test_search_relevance_cutoff_feature.py
```diff
@@ -1,102 +1,1771 @@
...
 from marqo.core.models.add_docs_params import AddDocsParams
 from marqo.core.models.marqo_index import *
+from marqo.core.models.marqo_index_request import FieldRequest
 from marqo.tensor_search.api import search
 from marqo.tensor_search.enums import SearchMethod
+from tests.integ_tests.marqo_test import MarqoTestCase
 
 
-class TestSearchRelevanceCutoffFeatureSort1Field(MarqoTestCase):
+class TestSearchRelevanceCutoffFeature(MarqoTestCase):
```
papa99do, line 14: Is the search result (order and score of each doc) deterministic? what about with multi-shards and multi-replica?
Also, why do we use greaterThan and lessThan to verify expected candidates? Shouldn't we verify with a specific number if the result is deterministic?

tests/integ_tests/tensor_search/integ_tests/test_search_relevance_cutoff_feature.py
```diff
@@ -1,102 +1,1771 @@
...
+        for index_name in [self.unstructured_index_name, self.structured_index_name]:
+            if 30 !=self.monitoring.get_index_stats_by_name(index_name).number_of_documents:
+                raise RuntimeError(
+                    f"Expected 30 documents in index {index_name} for sorting tests"
+                )
+
+    def tearDown(self):
+        """Ensure documents are not changed after each test."""
+        for index_name in [self.unstructured_index_name, self.structured_index_name]:
+            if 30 !=self.monitoring.get_index_stats_by_name(index_name).number_of_documents:
```
papa99do: I don't get this, why do we need to verify the doc number in setup and tearDown?

tests/integ_tests/tensor_search/integ_tests/test_search_sort_by_feature.py
```diff
@@ -268,6 +275,110 @@ def test_simple_sort_with_default_settings(self):
...
+            ]
+        }
+
+        hybrid_search_parameters = {
+            "retrievalMethod": "lexical",
+            "rankingMethod": "tensor",
+        }
+
+        for _ in range(10):
+            # We run it several times to ensure that the results are consistent
```
papa99do: why do we need to run multiple times? What could cause the result to be different?

tests/unit_tests/marqo/core/semi_structured_vespa_index/test_semi_structured_vespa_index_to_vespa_query.py
```diff
@@ -695,21 +695,78 @@ def test_gap_detection_default_and_custom_probeDepth(self):
...
+        # Test minimum valid relativeScoreFactor
+        params = RelativeMaxScoreParameters(relativeScoreFactor=0.001)
+        self.hybrid_query.relevance_cutoff = RelevanceCutoffModel(
+            method=RelevanceCutoffMethod.RelativeMaxScore,
+            parameters=params,
+            probe_depth=1  # minimum probe depth
+        )
+        
+        r = self.index.to_vespa_query(self.hybrid_query)
+        self.assertAlmostEqual(0.001,
```
papa99do: 0.001 is not too small, will we encounter precision issue when use `assertEqual`? I think we should use `assertEqual` if it can pass.

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -892,4 +1197,92 @@ HitGroup applyGlobalScoreModifiers(HitGroup hits, boolean verbose) {
...
+    /**
+     * Detects the cutoff count for relevance filtering based on the specified method
+     * @param probeCandidates The lexical search results to analyze
+     * @param cutoffMethodString The method to use for cutoff detection
+     * @param relevanceCutoffParameter The parameter for the cutoff method. It is unified across all methods.
+     * @param verbose Whether to log verbose information
+     * @return The number of relevant results to keep
+     */
+    @VisibleForTesting
+    Integer detectCutoffCount(
```
papa99do: nit. return a `int` instead, so the caller will not need to test if the result is null

vespa/src/test/java/ai/marqo/search/RelevanceCutoffTest.java
```diff
@@ -0,0 +1,734 @@
...
+            void shouldCountZeroWhenNoneAboveThreshold() {
+                HitGroup hits = createHitGroupWithScores(0.5, 0.4, 0.3, 0.2, 0.1);
+
+                Integer result = callDetectCutoffCount(hits, "mean_std_dev", 3.0, false);
+                assertThat(result).isEqualTo(0);
+            }
+        }
+
+        @Nested
+        class RelativeMaxScoreTest {
```
papa99do, line 193: Add a test case that the max score is less than 1.0.

vespa/src/test/java/ai/marqo/search/RelevanceCutoffTest.java
```diff
@@ -0,0 +1,734 @@
...
+        void shouldOverwriteTargetHitsInYql() {
+            String originalYql =
+                    "select * from sources * where {targetHits: 100, hnsw.exploreAdditionalHits:"
+                            + " 1900}";
+
+            String result = callOverwriteTargetHits(originalYql, 200);
+            assertThat(result).contains("targetHits: 200");
+            assertThat(result).doesNotContain("targetHits: 100");
+            assertThat(result).contains("hnsw.exploreAdditionalHits: 1800");
+            assertThat(result).doesNotContain("hnsw.exploreAdditionalHits: 1900");
```
papa99do: I think a better way is to just `assertThat(result).isEqualTo(expectedResult)` to verify it's exactly what we want it to be. Same for the following test cases

vespa/src/test/java/ai/marqo/search/RelevanceCutoffTest.java
```diff
@@ -0,0 +1,734 @@
...
+        }
+
+        private Integer callDetectCutoffCount(
+                HitGroup hits, String method, Double parameter, boolean verbose) {
+            return hybridSearcher.detectCutoffCount(hits, method, parameter, verbose);
+        }
+    }
+
+    @Nested
+    class TargetHitsRegexTest {
```
papa99do, line 258: There are quite a lot of test cases here. It's tricky to test a regex that handles free text since there could so many edge cases. Maybe out of scope of this project, but I was wondering if we could leverage something like [parameter substitution](https://docs.vespa.ai/en/reference/query-language-reference.html#parameter-substitution) or [Query property](https://docs.vespa.ai/en/reference/query-language-reference.html#query-properties) to parameterise a YQL query in the future.

vespa/src/test/java/ai/marqo/search/RelevanceCutoffTest.java
```diff
@@ -0,0 +1,734 @@
...
+            String yql2 = "{targetHits:500, hnsw.exploreAdditionalHits:4500}"; // efSearch = 5000
+            String result2 = callOverwriteTargetHits(yql2, 1000);
+            assertThat(result2).contains("targetHits:1000");
+            assertThat(result2).contains("hnsw.exploreAdditionalHits:4000"); // 5000 - 1000 = 4000
+
+            // Scenario 3: Edge case where newTargetHits equals efSearch
+            String yql3 = "{targetHits:100, hnsw.exploreAdditionalHits:900}"; // efSearch = 1000
+            String result3 = callOverwriteTargetHits(yql3, 1000);
+            assertThat(result3).contains("targetHits:1000");
+            assertThat(result3).contains("hnsw.exploreAdditionalHits:0"); // 1000 - 1000 = 0
```
papa99do: Add a test case that override is larger than efsearch?

src/marqo/core/search/hybrid_search.py
wanliAlex: Changed the condition. Relevance cut-off is only available to unstructured indexes now

src/marqo/core/search/hybrid_search.py
wanliAlex: Done

src/marqo/tensor_search/tensor_search.py
wanliAlex: We move them into the hybrid search class

tests/api_tests/v1/tests/api_tests/test_relevance_cutoff_feature.py
wanliAlex: As discussed, we remove the relevance score field as it is confusing

tests/api_tests/v1/tests/api_tests/test_relevance_cutoff_feature.py
wanliAlex: As discussed, we assert on the exact order of the response documents ids.

tests/api_tests/v1/tests/api_tests/test_relevance_cutoff_feature.py
wanliAlex: updated

tests/api_tests/v1/tests/api_tests/test_relevance_cutoff_feature.py
wanliAlex: done

tests/api_tests/v1/tests/api_tests/test_relevance_cutoff_feature.py
wanliAlex: done

tests/api_tests/v1/tests/api_tests/test_relevance_cutoff_feature.py
wanliAlex: done

tests/api_tests/v1/tests/api_tests/test_relevance_cutoff_feature.py
wanliAlex: Done

wanliAlex: We still can't be very strict with the order of returned documents, so we assert on the returned documents ids set.

tests/integ_tests/tensor_search/integ_tests/test_search_relevance_cutoff_feature.py
wanliAlex: updated

wanliAlex: Added a pass for setUp

tests/integ_tests/tensor_search/integ_tests/test_search_relevance_cutoff_feature.py
wanliAlex: Made the tests more restrictive

tests/integ_tests/tensor_search/integ_tests/test_search_sort_by_feature.py
wanliAlex: removed

tests/unit_tests/marqo/core/semi_structured_vespa_index/test_semi_structured_vespa_index_to_vespa_query.py
wanliAlex: use self.assertEqual

vespa/pom.xml
wanliAlex: It's not provided. I removed <scope>provided</scope>

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: done

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: Good catch! Done

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: Updated my code. Use this method in the rest of the code as it is a better practice

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: done

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: done

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: done

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: done

vespa/src/test/java/ai/marqo/search/RelevanceCutoffTest.java
wanliAlex: added

vespa/src/test/java/ai/marqo/search/RelevanceCutoffTest.java
wanliAlex: done

vespa/src/test/java/ai/marqo/search/RelevanceCutoffTest.java
wanliAlex: added

vespa/src/test/java/ai/marqo/search/RelevanceCutoffTest.java
papa99do, line 1: nit. I feel like there are duplicate test cases in this tests, (might be auto generated by AI), maybe try ask AI to identify what test cases are duplicate and consolidate them a bit. Also consider using parameterised test to further reduce the size of the test class for easier reading.

vespa/src/test/java/ai/marqo/search/SortByTest.java
```diff
@@ -742,4 +741,469 @@ void shouldAcceptValidSortDepth() {
...
+            Query result =
+                    searcher.updateQueryHitsOffsetsAndTargetHits(query, null, 20, false, true);
+
+            // Should use Math.max(sortByMinSortCandidates, limit+offset) = Math.max(20, 15) = 20
+            assertThat(result.getHits()).isEqualTo(20);
+            assertThat(result.getOffset()).isEqualTo(0);
+        }
+
+        @Test
+        void shouldUpdateHitsWhenSortByCandidatesLowerThanLimitOffset() {
```
papa99do: should not?

vespa/src/test/java/ai/marqo/search/SortByTest.java
```diff
@@ -742,4 +741,469 @@ void shouldAcceptValidSortDepth() {
...
+                                    searcher.updateQueryHitsOffsetsAndTargetHits(
+                                            query, 20, null, true, false))
+                    .isInstanceOf(RuntimeException.class)
+                    .hasMessageContaining(
+                            "The targetHits in the tensor query should not be smaller than"
+                                    + " limit+offset");
+        }
+
+        @Test
+        void shouldHandleNullParametersCorrectly() {
```
papa99do: is this a duplicate test?

vespa/src/test/java/ai/marqo/search/SortByTest.java
```diff
@@ -742,4 +741,469 @@ void shouldAcceptValidSortDepth() {
...
+            assertThat(updatedYql).doesNotContain("targetHits: 100");
+            assertThat(updatedYql).doesNotContain("hnsw.exploreAdditionalHits: 1900");
+        }
+    }
+
+    @Nested
+    class IntegrationRelevanceCutoffAndSortByTest {
+
+        @Test
+        void shouldUseMaxCandidatesWhenBothFeaturesEnabledRelevanceHigher() {
```
papa99do: dup with shouldUpdateHitsWhenBothFeaturesEnabled?

vespa/src/test/java/ai/marqo/search/SortByTest.java
```diff
@@ -742,4 +741,469 @@ void shouldAcceptValidSortDepth() {
...
+
+            Query result = searcher.updateQueryHitsOffsetsAndTargetHits(query, 50, 30, true, true);
+
+            // Should use Math.max(50, 30) = 50
+            assertThat(result.getHits()).isEqualTo(50);
+            assertThat(result.getOffset()).isEqualTo(0);
+        }
+
+        @Test
+        void shouldUseMaxCandidatesWhenBothFeaturesEnabledSortByHigher() {
```
papa99do: dup with shouldUpdateHitsWhenBothFeaturesEnabledReversed?

vespa/src/test/java/ai/marqo/search/SortByTest.java
```diff
@@ -742,4 +741,469 @@ void shouldAcceptValidSortDepth() {
...
+
+            Query result = searcher.updateQueryHitsOffsetsAndTargetHits(query, 40, 40, true, true);
+
+            // Should use Math.max(40, 40) = 40
+            assertThat(result.getHits()).isEqualTo(40);
+            assertThat(result.getOffset()).isEqualTo(0);
+        }
+
+        @Test
+        void shouldValidateTensorTargetHitsBeforeProcessingWithBothFeatures() {
```
papa99do: dup with shouldThrowExceptionWhenTensorTargetHitsLowerThanLimitOffset?

vespa/src/test/java/ai/marqo/search/SortByTest.java
wanliAlex: Done

vespa/src/test/java/ai/marqo/search/SortByTest.java
wanliAlex: deleted

vespa/src/test/java/ai/marqo/search/SortByTest.java
wanliAlex: deleted

vespa/src/test/java/ai/marqo/search/SortByTest.java
wanliAlex: deleted

vespa/src/test/java/ai/marqo/search/SortByTest.java
wanliAlex: deleted

farshidz (dismissed):

vespa/src/test/java/ai/marqo/search/SortByTest.java
```diff
@@ -296,7 +296,7 @@ void whenSortByFields_set_postProcessBySortIsCalled() {
                                 return null;
                             }
                             // non-null so createSubQuery and co. won't blow up
-                            return Tensor.from("tensor<float>()");
```
papa99do, line 300: qq. why this change?

papa99do (dismissed):

vespa/src/test/java/ai/marqo/search/SortByTest.java
wanliAlex: Just to better match the tensor format in vespa

papa99do (dismissed):

papa99do (dismissed):

papa99do (approved):

PR #1247, opened 2025-06-23
Log out slow or failed Marqo query

papa99do:
## Change Summary
  ✅ Features Implemented:

  1. E2E Latency Measurement: Measures complete search latency from request parsing through to response. Please note that the SearchQuery model construction is excluded since it used to be done by FastAPI framework, and we do not want to log out a malformed query with potential secret fields or long query string fields in a field with typo. 
  2. Configurable Logging: Three environment variables control the behavior:
    - MARQO_VESPA_SLOW_QUERY_THRESHOLD_MS (default: 500ms)
    - MARQO_VESPA_LOG_QUERY_DETAILS (default: FALSE)
    - MARQO_LOG_QUERY_MAX_LENGTH: (default: 10_000)
  3. Full Marqo Query Logging: Logs the complete search query dict with following sanitisation
    - fields containing secrets are excluded (download headers, model auth)
    - vectors are replaces by empty list
    - long query strings (root level, in hybrid parameters and custom vector content) are truncated to max length
  4. Log query under following scenarios:
    - query latency in Marqo (excluding SearchQuery construction) is larger than a threshold
    - any exception raises during the search process and json serialisation of the response 
  5. query is logged out as a python dict, so query strings are logged out as is, without extra escaping

  ✅ Usage Examples:

  Slow Query Log (when MARQO_VESPA_LOG_QUERY_DETAILS=TRUE), in json format:
  `{"timestamp":"2025-07-03 11:13:13,991","process":79163,"level":"WARNING","name":"marqo_query","message":"Slow search query detected: 3784.8ms. Query: {'q': 'hello', 'searchMethod': 'TENSOR', 'limit': 10, 'offset': 0, 'showHighlights': True}"}`

  Failed Query Log (in plain format):
  `ERROR: Failed search query - Error: Connection timeout. Query: {"q": "search term", "limit": 10}`

## Related Jira Ticket
https://s2search.atlassian.net/browse/MOSD-342

## Checklist

### For new field types:
- [N/A] Test indexes updated to cover the new type for all APIs (add docs, search, partial update, etc.)


src/marqo/core/models/hybrid_parameters.py
```diff
@@ -21,6 +21,9 @@ class RankingMethod(str, Enum):
 
 
 class HybridParameters(StrictBaseModel):
+    class Config(StrictBaseModel.Config):
+        use_enum_values = True
```
farshidz, line 25: What does this do?

src/marqo/core/search/query_logger.py
```diff
@@ -0,0 +1,84 @@
...
+                query_dict["q"]["customVector"]["content"] = _truncate_if_long(q.customVector.content)
+            # remove custom vector
+            query_dict["q"]["customVector"]["vector"] = []
+        elif isinstance(q, (str, dict)):
+            query_dict['q'] = _sanitise_str_or_dict_query(q)
+        else: # q is None, handle separate tensor and lexical q in hybrid parameter
+            if self.search_query.hybridParameters.queryTensor:
+                query_dict["hybridParameters"]["queryTensor"] = _sanitise_str_or_dict_query(
+                    self.search_query.hybridParameters.queryTensor)
+            if self.search_query.hybridParameters.queryLexical:
```
farshidz, line 65: If none of these, should we do something? e.g. log an error

src/marqo/core/search/query_logger.py
```diff
@@ -0,0 +1,84 @@
...
+
+        return query_dict
+
+    def log_error_query(self, error_message: str):
+        if log_query_details:
+            marqo_query_logger.error(f'Failed search query: Error: {error_message}. Query: {self.sanitised_query}')
+            self.error_logged = True  # Mark that error was logged
+
+    def log_slow_query(self, elapsed_time_ms: float):
+        if log_query_details and not self.error_logged and elapsed_time_ms >= slow_query_threshold_ms:
```
farshidz: nit should it be > not >=?

tests/unit_tests/marqo/core/semi_structured_vespa_index/test_semi_structured_vespa_index_to_vespa_query.py
```diff
@@ -472,8 +472,8 @@ def test_sort_by_with_missing_first(self):
 
         self.assertEqual(1, len(sort_fields))
         self.assertEqual("description", sort_fields[0]["field_name"])
-        self.assertEqual("asc", sort_fields[0]["order"].value)
-        self.assertEqual("first", sort_fields[0]["missing"].value)
+        self.assertEqual("asc", sort_fields[0]["order"])
```
farshidz, line 475: Why did we need to change these?

src/marqo/core/models/hybrid_parameters.py
papa99do: This populates the model with the value of the enum (the input still need to be a valid enum), so when you call model.dict() to serialise the model later, you get the value automatically. For example, you get `disjunction` instead of `<RetrievalMethod.Disjunction: disjunction>`. 

see https://docs.pydantic.dev/1.10/usage/model_config/#options for details

Ideally, we should use StrEnum instead so it auto converts to its string value when accessed. StrEnum is only supported in python 3.11+.

src/marqo/core/search/query_logger.py
papa99do: The validation should be done when constructing this pydantic model. If we add a branch, there's no way to cover it in tests.

src/marqo/core/search/query_logger.py
papa99do: good point. I'll change

tests/unit_tests/marqo/core/semi_structured_vespa_index/test_semi_structured_vespa_index_to_vespa_query.py
papa99do: because of `use_enum_values = True`

src/marqo/core/models/hybrid_parameters.py
farshidz: Can you please search the code base for .dict, .val and other calls affected by this, to be sure?

farshidz (approved):

src/marqo/core/models/hybrid_parameters.py
papa99do: query dict is used in following places:
* when the query is serialised to json to pass to Vespa, in which case the enum values are used in json serialisation
* in pagination fix, its json representation is used to generate a hash of the query. Enum value will also be used when encoded to json
* debug log before sending to Vespa, which is just used in local env, so should be fine
* in the query logging, which will prefer to use the enum value. (here, we do not encode to json on purpose to avoid query string or filters being escaped by json enccoding)

PR #1251, opened 2025-06-25
recommend: hardcoded tensor field order

vicilliar:
## Change Summary
Test fix.
Personalization changes `recommend` such that the order of tensor fields is now deterministic based on the `tensor_fields` parameter instead of it being determined by the vespa return object. This updates the tests to hardcode the tensor field order.

## Checklist

### For new field types:


vicilliar:
Passing backwards compatibility test (Except for the global score modifiers known issue): https://github.com/marqo-ai/marqo/actions/runs/15869853517/job/44748192917

farshidz (approved):

PR #1254, opened 2025-06-26
Context Search With Documents (Personalization) - Revised

vicilliar:
Updated version of personalization PR intended for 2.21.0 which was reverted due to an issue with the Vespa client.
Differences from old personalization PR:
-Use persistent SSL context instead of persistent transport, because it caused hanging during concurrent requests. Performance gains are comparable.
-Integration test added for concurrent get documents requests to ensure no hanging

## Change Summary
Supports new key `documents` in `context` parameter of search. This allows for a dict of doc_id:weight pairs to be provided by user for vectors to be extracted and interpolated for tensor search or hybrid search with tensor retrieval or ranking.

Updated example context object:
```
    {
        "tensor": [{"vector": [1 for i in range(768)], "weight": 1}], 
        "documents": {
            "ids": {"doc1": 1},
            "parameters": {
                "excludeInputDocuments": False,
                "tensorFields": ["text_field_1"],
                "concurrency": 10
            }
        }
    }
```
-Interpolation is now done on all vectors (query, context.tensor, context.document) instead of previous behavior of weighted average.
-Absolute value is now used when summing vector weights in interpolation (This makes ZeroSumWeightsError obsolete, we now use AllZeroWeightsError instead)
-Interpolation logic implemented in numpy.
-Fixed some error messages for hybrid / pydantic model validation.

New Env Vars:
-`MARQO_MAX_SEARCH_CONTEXT_DOCS` - maximum number of documents that can be used as context documents in a search request.

The following optimizations have been implemented:
-Getting doc vectors fetches only essential embeddings fields from Vespa
-A persistent SSL Context is created upon vespa client initialization and is used for the `verify` param of every async client instantiation. This saves time spent creating new contexts.
-All endpoints that use `vespa_client.get_batch()` also benefit from speedup from the persistent SSL Context (~5ms). These include: (1) get_documents, (2) partial_updates with maps, (3) add docs with use_existing_tensors=True, (4) recommend.
-Index is fetched from cache instead of Vespa
-Fetched doc vectors are not formatted into document form with Pydantic. Vectors are just used directly

## Related Jira Ticket
https://s2search.atlassian.net/jira/polaris/projects/MOSD/ideas/view/4863474?selectedIssue=MOSD-23&issueViewLayout=sidebar&issueViewSection=overview&atlOrigin=eyJpIjoiNzcyNDBkNzVjM2JhNDE0Y2I4ODUzNDY3YTc2MmE0ZmUiLCJwIjoiaiJ9

## Checklist
py-marqo changes: https://github.com/marqo-ai/py-marqo/pull/294

### For new field types:


tests/api_tests/v1/tests/api_tests/test_no_model_feature.py
```diff
@@ -205,7 +205,7 @@ def test_no_model_work_with_custom_vectors_in_search(self):
                                                          context={"tensor": [{"vector": [1, ] * self.DIMENSION,
                                                                               "weight": -1},
                                                                              {"vector": [1, ] * self.DIMENSION,
-                                                                              "weight": 1}], })
```
wanliAlex, line 208: Maybe add a test to ensure a proper error is raised?

tests/api_tests/v1/tests/api_tests/test_search_common.py
```diff
@@ -388,8 +386,101 @@ def test_hybrid_search_validations(self):
...
+                    ],
+                    tensor_fields=["title", "content"] if index_name == self.unstructured_text_index_name else None
+                )
+
+                test_cases = [
+                    ("disjunction", "rrf", {"best pets": 1}, "animals"),
+                    ("tensor", "tensor", {"best pets": 1}, None),
+                    ("tensor", "lexical", {"best pets": 1}, "animals"),
+                    ("lexical", "tensor", {"best pets": 1}, "animals")
+                ]
```
wanliAlex, line 460: What will happen if users provide context with documents to lexical-lexical hybrid search?

tests/integ_tests/tensor_search/integ_tests/test_get_documents_by_ids.py
```diff
@@ -317,6 +321,101 @@ def run():
                         return True
                 assert run()
 
+    def test_get_documents_concurrent(self):
```
wanliAlex: Do we need to test the error behaviour here? e.g., one thread fail

tests/integ_tests/tensor_search/search/test_search_combined.py
```diff
@@ -207,6 +213,39 @@ def tearDown(self) -> None:
         super().tearDown()
         self.device_patcher.stop()
 
+    def _populate_index_orchids(self, index):
```
wanliAlex: Why are we adding this method but it is not used in the class?

tests/integ_tests/tensor_search/search/test_search_with_context.py
```diff
@@ -0,0 +1,1126 @@
...
+from marqo.core.models.interpolation_method import InterpolationMethod
+from marqo.core.utils.vector_interpolation import Slerp, Lerp, Nlerp, AllZeroWeightsError, ZeroMagnitudeVectorError
+from marqo.exceptions import InvalidArgumentError, InternalError
+from marqo.tensor_search.models.score_modifiers_object import ScoreModifierLists, ScoreModifierOperator
+from tests.integ_tests.marqo_test import MarqoTestCase
+from tests.integ_tests.utils.transition import *
+from marqo.core.models.hybrid_parameters import RetrievalMethod, RankingMethod, HybridParameters
+from marqo.tensor_search.models.api_models import CustomVectorQuery
+
+class TestSearchWithContext(MarqoTestCase):
```
wanliAlex, line 22: Add tests/error handling for different search methods:
1. Lexical;
2. Tensor
3. Hybrid(tensor-tensor, lexical-lexical, ect)

tests/integ_tests/tensor_search/search/test_search_with_context.py
```diff
@@ -0,0 +1,1126 @@
...
+            {"_id": "continent2", "text_field_1": "Asia is Earth's largest and most populous continent, located primarily in the Eastern and Northern Hemispheres.", "tags": ["continent"]},
+            {"_id": "continent3", "text_field_1": "Africa is the world's second-largest and second-most populous continent, after Asia in both cases.", "tags": ["continent"]},
+        ]
+        
+        self.add_documents(
+            config=self.config,
+            add_docs_params=AddDocsParams(
+                index_name=index.name,
+                docs=docs,
+                tensor_fields=["text_field_1"] if isinstance(index, UnstructuredMarqoIndex) else None
```
wanliAlex, line 115: I would recommend adding a new test class where documents have more than 1 tensor field in the documents and index.

tests/unit_tests/core/search/test_recommender_vectors.py
```diff
@@ -0,0 +1,680 @@
...
+from marqo.vespa.vespa_client import VespaClient
+from marqo.core.index_management.index_management import IndexManagement
+from marqo.core.inference.api import Inference
+from marqo.core.models.interpolation_method import InterpolationMethod
+from marqo.core.utils.vector_interpolation import AllZeroWeightsError
+from marqo.tensor_search.models.search import SearchContext, SearchContextTensor
+from marqo.api.exceptions import InvalidDocumentIdError
+
+
+class TestRecommenderGetDocVectorsFromIds:
```
wanliAlex: Refactor the test class so it uses our testing structure.
1. This test class should be a subclass of unittest.TestCase;
2. Use self.assert instead of assert for clarification

wanliAlex (changes requested):

tests/api_tests/v1/tests/api_tests/test_search_common.py
vicilliar: Actually it be allowed and just ignored, we should really error out if any context is provided for this, but if we add that now it will be a breaking change.

tests/integ_tests/tensor_search/search/test_search_combined.py
vicilliar: Good catch, I moved this to another file but forgot to delete this method! will delete.

tests/unit_tests/core/search/test_recommender_vectors.py
vicilliar: done

tests/api_tests/v1/tests/api_tests/test_no_model_feature.py
vicilliar: not sure what you mean. this test isn't supposed to raise an error, I just changed it because it now accidentally makes the context tensor a zero magnitude vector, which raises an error. I'll add the following tests to integration tests:
1. context tensors that cancel each other out (weights opposite of each other) to make a zero magnitude vector should raise an error.
2. search with context tensors with all zero weight should raise an error
3. search with context documents with all zero weight should raise an error

tests/integ_tests/tensor_search/integ_tests/test_get_documents_by_ids.py
vicilliar: added test for it

tests/integ_tests/tensor_search/search/test_search_with_context.py
vicilliar: added new test case

vicilliar: done

tests/integ_tests/tensor_search/search/test_search_with_context.py
vicilliar: For lexical, hybrid lexical/lexical erroring out, I added these to API tests and api model unit tests, since integration tests calling tensor_search won't catch it, since it's in the API model. 

As for tensor, all the tests in this suite use tensor search.

Hybrid search tests are in api_tests/test_hybrid_search. I can add hybrid search tests to integration tests as well

wanliAlex (approved):
LGTM

PR #1255, opened 2025-06-27
Fix media file chunking issue when overlap is larger than duration

papa99do:
## Change Summary
Fixed the bug that generates zero chunks when `split_overlap` is larger than the duration of the media file. 

## Related Jira Ticket
https://s2search.atlassian.net/browse/MOSD-402

## Checklist

### For new field types:
- [N/A] Test indexes updated to cover the new type for all APIs (add docs, search, partial update, etc.)

### DIFF
--- a/src/marqo/inference/media_download_and_preprocess/streaming_media_processor.py
+++ b/src/marqo/inference/media_download_and_preprocess/streaming_media_processor.py
@@ -15,6 +15,62 @@ from marqo.core.models.marqo_index import *
 from marqo.inference.native_inference.embedding_models.languagebind_model import LanguagebindPreprocessor
 
 
+class ChunkTimingGenerator:
+    """Generator class to emit (chunk_start, chunk_end) pairs for media processing."""
+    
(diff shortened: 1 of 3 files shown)


src/marqo/inference/media_download_and_preprocess/streaming_media_processor.py
```diff
@@ -15,6 +15,54 @@
 from marqo.inference.native_inference.embedding_models.languagebind_model import LanguagebindPreprocessor
 
 
+class ChunkTimingGenerator:
```
wanliAlex, line 18: Why do you use a class instead of creating a method inside the StreamingMediaProcessor here?

wanliAlex (approved):

src/marqo/inference/media_download_and_preprocess/streaming_media_processor.py
papa99do: It's easier to test a class then a method in StreamingMediaProcessor (so that we do not need to instantiate a StreamingMediaProcessor to test the simple chunking logic).

PR #1256, opened 2025-06-27
Omit base64 in search response

farshidz:
## Change Summary
Base64 queries can be large and returning these in the search response could create unnecessarily large responses. This change replaces base64 queries with a fixed string, including as part of a dict.

## Related Jira Ticket
N/A

## Checklist

### For new field types:
- [x] Tests cover score modifier usage of this new type
- [x] Test indexes updated to cover the new type for all APIs (add docs, search, partial update, etc.)

### DIFF
--- a/src/marqo/tensor_search/tensor_search.py
+++ b/src/marqo/tensor_search/tensor_search.py
@@ -87,7 +87,7 @@ from marqo.tensor_search.models.relevance_cutoff_model import RelevanceCutoffMod
 logger = get_logger(__name__)
 
 
-def _sanitize_query_for_response(query: str):
+def _sanitize_query_for_response(query: Optional[Union[str, dict]]):
     """
     Replace base64 image content in queries with 'data:image/[omitted]' for response.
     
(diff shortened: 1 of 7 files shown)


src/marqo/tensor_search/tensor_search.py
```diff
@@ -87,6 +87,37 @@
 logger = get_logger(__name__)
 
 
+def _sanitize_query_for_response(query):
```
wanliAlex: nit: TypeHint

src/marqo/tensor_search/tensor_search.py
```diff
@@ -87,6 +87,37 @@
...
+        
+    Returns:
+        The sanitized query object with base64 content replaced
+    """
+    if query is None:
+        return query
+    
+    if isinstance(query, str):
+        if is_base64_image(query):
+            return 'data:image/[omitted]'
```
wanliAlex, line 105: Would it be better as `'data:image/omittedBased64Image'`

wanliAlex (changes requested):

src/marqo/tensor_search/tensor_search.py
farshidz: We agreed on this on Slack so best not to open that can of worms now

src/marqo/tensor_search/tensor_search.py
farshidz: Added

wanliAlex (approved):
LGTM

wanliAlex (approved):

PR #1259, opened 2025-07-03
Using 3 instances to run integration tests

wanliAlex:
## Change Summary

Use 3 instances to run integration tests to avoid the rate limit from HuggingFace.
Fix a test issue that the `marqo__id` field is not in the test schema.

## Checklist

### For new field types:

### DIFF
--- a/scripts/vespa_local/vespa_local.py
+++ b/scripts/vespa_local/vespa_local.py
@@ -52,7 +52,10 @@ class VespaLocal:
         return textwrap.dedent("""
             schema test_vespa_client {
                 document test_vespa_client {
-
+                    field marqo__id type string {
+                        indexing: summary | attribute
+                    }
+                
                     field id type string {
                         indexing: summary | attribute
                     }
--- a/src/marqo/inference/native_inference/load_model.py
+++ b/src/marqo/inference/native_inference/load_model.py
@@ -372,7 +372,7 @@ def _load_model(
             logger.info(f"✅ Model `{model_name}` loaded successfully on `{device}`.")
             return model  # ✅ Success, return the model
 
-        except (ReadTimeoutError, Timeout, OSError, RuntimeError) as e:
+        except InvalidModelPropertiesError as e:
             logger.info(f"⚠️ Error loading model `{model_name}` on `{device}`: {e}")
             attempt += 1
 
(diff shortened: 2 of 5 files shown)


danyilq (dismissed):
lgtm

src/marqo/inference/native_inference/load_model.py
```diff
@@ -352,14 +358,31 @@ def _load_model(
...
+                device=device,
+                model_properties=model_properties,
+                model_auth=model_auth,
+            )
+
+            model.load()
+            logger.info(f"✅ Model `{model_name}` loaded successfully on `{device}`.")
+            return model  # ✅ Success, return the model
+
+        except InvalidModelPropertiesError as e:
```
farshidz: Is there a way to catch the rate limiting issue specifically and not retry on genuinely invalid model properties?

farshidz: Why are we changing Marqo code to get around a test issue? We should be changing our test code

scripts/vespa_local/vespa_local.py
```diff
@@ -52,7 +52,10 @@ def get_test_vespa_client_schema_content(self) -> str:
         return textwrap.dedent("""
             schema test_vespa_client {
                 document test_vespa_client {
-
+                    field marqo__id type string {
```
farshidz, line 55: Why this change?

src/marqo/inference/native_inference/load_model.py
wanliAlex: This is  the error message exposed to marqo: `OSError: We couldn't connect to 'https://huggingface.co/' to load this file, couldn't find it in the cached files and it looks like sentence-transformers/all-MiniLM-L6-v2 is not the path to a directory containing a file named config.json.` So I think it's generic and does not mention the specific rate limit.

wanliAlex: Having more instances to run tests?

scripts/vespa_local/vespa_local.py
wanliAlex: Tests are failing after we split the tests into 3 groups. Some tests were interfering with each other. Failed tests require this field in the schema.

papa99do (approved):

PR #1260, opened 2025-07-03
Emit statsD metrics directly from Marqo, replacing RP metrics

ommirzaei:
## Change Summary
* Added StatsD middleware to emit all RP metrics (requests.completed, marqo_processing_time, search_processing_time, index_processing_time, x-count-success, x-count-failure, x-count-error) from Marqo.

* Implemented global "common tags" support via STATSD_COMMON_TAGS env var.

* Ensured path redaction for document IDs to prevent high-cardinality tags.

* Updated and added tests (unit and integration) to cover all metric paths, headers, and tag logic.

## Related Jira Ticket
https://s2search.atlassian.net/jira/polaris/projects/MOSD/ideas/view/4863472?selectedIssue=MOSD-446

## Checklist

### For new field types:


src/marqo/core/monitoring/statsd_client.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        Send a message to the StatsD server.
+        Args:
+            msg: The message to send, formatted as per DogStatsD protocol.
+        """
+        try:
+            # UDP – fire and forget
+            self._sock.sendto(msg.encode("utf-8"), self.addr)
+        except Exception:
+            # Metrics must never break request handling
```
papa99do, line 115: should we log a warning message here? how do we know if the metrics collection works properly? If we are concerned about the performance impact or being too verbose, we can either:
* have an flag (probably env var) to turn the logging on/off
* log in debug level, just for testing purpose.

src/marqo/core/monitoring/statsd_client.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def __init__(
+        self,
+        host: Optional[str] = None,
+        port: Optional[int] = None,
+        prefix: str = "",
+    ) -> None:
+        self.addr = (
+            host or os.getenv("STATSD_HOST", "127.0.0.1"),
+            int(port or os.getenv("STATSD_PORT", 8125)),
```
papa99do: leverage utils.read_env_vars_and_defaults and utils.read_env_vars_and_defaults_ints. Have these env vars configured in the `EnvVars` class in enums.py, and default values in `config.py` . same for other env vars.

src/marqo/core/monitoring/statsd_client.py
```diff
@@ -0,0 +1,118 @@
...
+        try:
+            # UDP – fire and forget
+            self._sock.sendto(msg.encode("utf-8"), self.addr)
+        except Exception:
+            # Metrics must never break request handling
+            pass
+
+
+# Module-level singleton – keeps one UDP socket open per process         #
+_default_client = StatsDClient(prefix=os.getenv("STATSD_PREFIX", ""))
```
papa99do: This module level singleton is not test-friendly. Is this only used by middleware? why don't we instantiate it in middleware __init__ method? or better, pass it in, and create the client in api.py file. something like `app.add_middleware(StatsDMiddleware, statsd_client=StatsDClient(...))`

src/marqo/core/monitoring/statsd_middleware.py
```diff
@@ -0,0 +1,86 @@
...
+        if request.url.path == "/":
+            return await call_next(request)
+        t_start = time.perf_counter()
+        response: Response = await call_next(request)
+        duration_ms = int((time.perf_counter() - t_start) * 1000)
+
+        statsd = get_client()
+
+        status = response.status_code
+        status_tag = f"{status // 100}XX"      # 2XX / 3XX / 4XX / 5XX
```
papa99do, line 42: Not a scope of current project. but for a future improvement, it might be also useful to count 404, 429 and 504 individually.

src/marqo/core/monitoring/statsd_client.py
```diff
@@ -0,0 +1,118 @@
...
+        self.addr = (
+            host or os.getenv("STATSD_HOST", "127.0.0.1"),
+            int(port or os.getenv("STATSD_PORT", 8125)),
+        )
+        self.prefix = prefix
+        self._sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        self._sock.setblocking(False)
+
+        # Parse once; reused for every metric
+        self._common_tags = _parse_common_tags(os.getenv("STATSD_COMMON_TAGS"))
```
papa99do: Do we group by instance_id or pod_id? or just at account/index level?

tests/integ_tests/core/monitoring/test_metrics_udp.py
```diff
@@ -0,0 +1,148 @@
...
+        self.packets: List[bytes] = []
+        self._thread = threading.Thread(target=self._loop, daemon=True)
+        self._thread.start()
+
+    def _loop(self):
+        """Continuously listen for UDP packets and store them."""
+        while self._running:
+            try:
+                data, _ = self._sock.recvfrom(4096)
+                self.packets.append(data)
```
papa99do: nit. there might be a thread safety issue when we iterate the packets in a thread and append data in this daemon thread. I run this test 100 times and all passed, so I guess the chance of hitting race condition is low. worst case is a flaky test. So feel free to leave it as is.

tests/integ_tests/core/monitoring/test_metrics_udp.py
```diff
@@ -0,0 +1,148 @@
...
+        os.environ["STATSD_HOST"] = "127.0.0.1"
+        os.environ["STATSD_PORT"] = str(sink.port)
+
+        # Re‑exec statsd_client so the singleton picks the new env‑vars
+        importlib.reload(sc)
+
+        # Also reload the middleware module so its imported get_client() refers
+        # to the newly reloaded statsd_client module.
+        import marqo.core.monitoring.statsd_middleware as sm
+        importlib.reload(sm)
```
papa99do: if we inject the StatsDClient to the middleware, we won't need to reload modules.

tests/integ_tests/core/monitoring/test_metrics_udp.py
```diff
@@ -0,0 +1,148 @@
...
+    """Test that the StatsD middleware emits the expected metrics."""
+    client, sink = client_and_sink
+
+    # exercise all code‑paths the middleware cares about
+    client.get("/")
+    client.get("/indexes/foo/search")  # search timing
+    client.post("/indexes/foo/documents")  # index timing + headers
+    client.get("/indexes/foo/documents/abc123")  # redaction
+
+    sink.wait(n=8)
```
papa99do: why 8? we only asserted 6 packets

tests/unit_tests/marqo/core/monitoring/test_statsd_and_middleware.py
```diff
@@ -0,0 +1,273 @@
...
+
+@pytest.fixture(autouse=True)
+def stub_statsd(monkeypatch):
+    """
+    Replace both get_client (in statsd_client and statsd_middleware)
+    with a stub so every call in the middleware is captured.
+    """
+    stub = _StubStatsD()
+    monkeypatch.setattr(sc, "get_client", lambda: stub, raising=False)
+    monkeypatch.setattr(sm, "get_client", lambda: stub, raising=False)
```
papa99do: If we can inject the statsD client, we don't need to monkeypatch this.

tests/unit_tests/marqo/core/monitoring/test_statsd_and_middleware.py
```diff
@@ -0,0 +1,273 @@
...
+def test_send_format_increment_and_timing(monkeypatch):
+    # test the _send helpers directly
+    captured: List[bytes] = []
+
+    # patch the _send helper (easier than socket.sendto)
+    monkeypatch.setattr(
+        sc.StatsDClient,
+        "_send",
+        lambda self, msg: captured.append(msg.encode()),
+        raising=True,
```
papa99do: why don't we use the stub?

tests/unit_tests/marqo/core/monitoring/test_statsd_and_middleware.py
```diff
@@ -0,0 +1,273 @@
...
+    client.get("/")
+
+    # No increments or timings expected
+    assert _extract(stub_statsd, "increment", "requests.completed") == []
+    assert not any(m.startswith("marqo_processing_time") for _, m in stub_statsd.sent)
+
+
+def test_search_metrics(client, stub_statsd):
+    """Test that search requests emit the correct StatsD metrics."""
+    stub_statsd.sent.clear()
```
papa99do: nit. this is required for each test, can be put in a setUp or tearDown method.

tests/unit_tests/marqo/core/monitoring/test_statsd_and_middleware.py
```diff
@@ -0,0 +1,273 @@
...
+        stub_statsd.sent.clear()
+        client.patch("/indexes/foo/documents")
+
+    # index_processing_time should be present
+    assert any(k == "timing" and msg.startswith("index_processing_time") for k, msg in stub_statsd.sent)
+    # malformed headers should NOT raise, and no increment for NaN values should be present
+    assert not any(msg.startswith("x-count-success") for k, msg in stub_statsd.sent)
+
+
+def test_statsd_prefix(monkeypatch):
```
papa99do: following tests are for StatsDClient, I recommend to move it to its seperate class, and maybe add more tests to cover increment and timing.

tests/integ_tests/core/monitoring/test_metrics_udp.py
```diff
@@ -0,0 +1,148 @@
...
+        os.environ.pop("STATSD_HOST", None)
+        os.environ.pop("STATSD_PORT", None)
+
+
+def _has(pkt: List[str], pattern: str) -> bool:
+    """ Check if any packet in the list matches the given regex pattern. """
+    return any(re.search(pattern, p) for p in pkt)
+
+
+def test_metrics_roundtrip(client_and_sink):
```
papa99do: Conventionally, we use unittest and pytest runner. @farshidz , do you think we can accept this new style? or should this be converted to unittest style?

src/marqo/core/monitoring/statsd_client.py
```diff
@@ -0,0 +1,118 @@
+import os
+import socket
+from typing import Dict, Optional
+
+
+def _parse_common_tags(raw: Optional[str]) -> Dict[str, str]:
```
papa99do: why is this a top level method? should it be moved in the StatsDClient since it's only used there?

ommirzaei: Moved to the class level as static method

src/marqo/core/monitoring/statsd_client.py
ommirzaei: For the metrics, we group by just the index name and account_id. But for the logs, we also include instance_id and pod_id in the log context.

src/marqo/core/monitoring/statsd_middleware.py
ommirzaei: Yes, for this PR we want to keep the metric as per the reverse proxy to make sure the current dashboards keep working. I guess we haven’t added per-response code metrics to reduce cardinality and cost.

tests/integ_tests/core/monitoring/test_metrics_udp.py
ommirzaei: I have updated the tests to follow the existing style.

tests/integ_tests/core/monitoring/test_metrics_udp.py
ommirzaei: Update the test to explicitly validate that the StatsD middleware emits each expected metric

tests/integ_tests/core/monitoring/test_metrics_udp.py
```diff
@@ -111,38 +118,36 @@ def get_doc(name: str, doc_id: str):
...
+        self.client.get("/indexes/foo/search")  # search timing
+        self.client.post("/indexes/foo/documents")  # index timing + headers
+        self.client.get("/indexes/foo/documents/abc123")  # redaction
+
+        patterns = [
+            r"marqo_processing_time:\d+\|ms",
+            r"requests\.completed:1\|c\|#status_code:\dXX",
+            r"search_processing_time:\d+\|ms",
+            r"index_processing_time:\d+\|ms",
+            r"x-count-success:\d+\|c",
```
papa99do, line 151: qq. why are x-count-failure and x-count-error missing from the metrics? Do we only emit positive numbers for a counter?

tests/unit_tests/marqo/core/monitoring/test_statsd_client.py
```diff
@@ -0,0 +1,60 @@
...
+
+        # order of tags not guaranteed
+        assert client.sent in [
+            [b"foo:2|c|#env:dev,team:ops"],
+            [b"foo:2|c|#team:ops,env:dev"],
+        ]
+
+    def test_parse_common_tags_empty_and_malformed(self):
+        assert sc.StatsDClient._parse_common_tags("") == {}
+        assert sc.StatsDClient._parse_common_tags("foo:bar,,baz:") == {"foo": "bar"}
```
papa99do, line 62: qq. should this raise an error?

tests/unit_tests/marqo/core/monitoring/test_statsd_middleware.py
```diff
@@ -0,0 +1,161 @@
...
+        ))
+
+    def test_index_docs_metrics_and_headers(self):
+        self.client.post("/indexes/foo/documents")
+
+        self.assertTrue(any(m.startswith("index_processing_time") for _, m in self.stub.sent))
+        incs = _extract(self.stub, "increment")
+        self.assertTrue(any(m.startswith("x-count-success:5") for m in incs))
+        self.assertTrue(any(m.startswith("x-count-failure:1") for m in incs))
+        self.assertTrue(any(m.startswith("x-count-error:0") for m in incs))
```
papa99do, line 126: qq. it seems counter with 0 is also emitted.

tests/unit_tests/marqo/core/monitoring/test_statsd_middleware.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def test_requests_completed_5xx(self):
+        self.client_ctx.__exit__(None, None, None)
+        self.client_ctx = TestClient(_app_with_docs_and_fail(self.stub))
+        self.client = self.client_ctx.__enter__()
+        self.stub.sent.clear()
+
+        self.client.get("/fail")
+        msgs = _extract(self.stub, "increment", "requests.completed")
+        self.assertTrue(any("status_code:5XX" in m for m in msgs))
```
papa99do, line 149: nit. maybe add a test case for any 4xx error as well,  and verify 2xx status code in the happy path,  for completeness.

papa99do (dismissed):

tests/unit_tests/marqo/core/monitoring/test_statsd_middleware.py
ommirzaei: Yes, if `x-count-error: "0"` exists in the header (as in the test), we expect to emit it, even though it’s zero.

tests/unit_tests/marqo/core/monitoring/test_statsd_client.py
ommirzaei: I don't think so. A misconfiguration adding extra tags shouldn't stop Marqo from running. We can always update the common tags if they don't appear correctly in the CloudWatch dashboard.

tests/unit_tests/marqo/core/monitoring/test_statsd_middleware.py
ommirzaei: Done.

tests/integ_tests/core/monitoring/test_metrics_udp.py
ommirzaei: Added them to the patterns list.

papa99do (dismissed):

tests/unit_tests/marqo/core/monitoring/test_statsd_middleware.py
```diff
@@ -0,0 +1,170 @@
...
+        self.client_ctx.__exit__(None, None, None)
+        self.stub.sent.clear()
+
+    def test_root_request_metrics(self):
+        """Health-check path ‘/’ should NOT emit metrics."""
+        self.client.get("/")
+        self.assertEqual(_extract(self.stub, "increment", "requests.completed"), [])
+        self.assertFalse(any(m.startswith("marqo_processing_time") for _, m in self.stub.sent))
+
+    def test_search_metrics(self):
```
farshidz, line 106: Can we have a short pydocs description for all tests please?

src/marqo/core/monitoring/statsd_middleware.py
```diff
@@ -0,0 +1,88 @@
...
+from starlette.responses import Response
+
+from marqo.core.monitoring.statsd_client import StatsDClient
+
+_SEARCH_RE = re.compile(r"/indexes/[^/]+/search$")
+_DOCS_RE = re.compile(r"/indexes/[^/]+/documents$")
+_DOCUMENT_ID_RE = re.compile(r"(/documents/)[^/]+")
+
+
+class StatsDMiddleware(BaseHTTPMiddleware):
```
farshidz, line 16: nit I wonder if this and StatsD client belong under `api` because they're specific to HTTP APIs while everything under core makes no interface assumption.

ommirzaei: This folder was suggested by @papa99do , but happy to move it to a folder like `src/marqo/api/monitoring/` or `src/marqo/api/metrics/` if you'd prefer.

tests/unit_tests/marqo/core/monitoring/test_statsd_middleware.py
ommirzaei: Added the pydocs description to all tests.

src/marqo/core/monitoring/statsd_middleware.py
papa99do: @farshidz  do you mean files in marqo/core should not depend on the FastAPI framework? I actually asked @ommirzaei to move the files here since I reckon we should keep all monitoring related logic in one place. The dependency might be an issue when we introduce strict check to enforce the code structured. But it should be easy to move them around in the future.

farshidz: > do you mean files in marqo/core should not depend on the FastAPI framework? 

I'd say so, but it's a bit subjective I guess. Happy to leave these here, and yes they're easy to move later if needed.

farshidz (dismissed):

papa99do (approved):

PR #1269, opened 2025-07-18
Quality of life improvements for backwards compatibility tests

vicilliar:
## Change Summary
We have several issues with backwards compatibility tests that can be solved with some quality of life improvements. These changes will make running these tests easier and faster.

### Configurable path to test
The `path_to_test` parameter was added so we can choose specific tests to run. This is useful if we only want to test 1 or multiple specific tests. This can accept dirs, files, or specific test classes. Leaving this blank will just run all tests in `tests/compatibility_tests/`

Example: 
```
create_structured_index create_index/test_create_index.py add_or_replace_documents/structured/test_add_documents_v2_2.py::TestAddDocumentsv2_2
```
This will run: (1) All tests in the dir `create_structured_index/`, (2) the test case in `create_index/test_create_index.py`, (3) the test case `add_or_replace_documents/structured/test_add_documents_v2_2.py::TestAddDocumentsv2_2`.

### Configurable ECR Marqo Image 
- The `existing_marqo_image_github_sha` parameter was added so we can use an existing ECR image for a compatibility test run. This is useful because if we're fixing a test, we may be pushing commits/triggering runs that use the exact same marqo source code (test-only changes). This will prevent the unnecessary building of a new marqo image every time. Just paste a SHA from a previous run that has an image built.

Example: `86c38b67db61e00e9dc9028bcbe0c5d49d0136f1`
 
### Other Changes
- 60 minute timeout added to execution pipeline (to stop hanging runs)
- Improved readability by removing levelname from logs, setting logger level to INFO

## Checklist

### For new field types:


.github/workflows/backwards_compatibility_marqo_execution.yml
```diff
@@ -53,11 +57,16 @@ on:
           - rollback
         required: true
         type: choice
+      path_to_test:
+        description: 'Path sent to pytest for running test mode'
```
wanliAlex, line 61: Can you add an example in the description to show what the path should look like?

wanliAlex (changes requested):

PR #1272, opened 2025-07-22
Emit generic high cardinality metrics

ommirzaei:
## Change Summary
Replaces legacy RP-style metrics with a generic StatsD middleware in Marqo.

Emits:
* request.duration_ms (with path, method, status_code tags)
* batch.success, batch.failure, batch.error (for document endpoints)

Also includes:
Path sanitization for document IDs
No-op for health-check route /
Full test suite update for new metrics behaviour

## Related Jira Ticket
https://s2search.atlassian.net/jira/polaris/projects/MOSD/ideas/view/4863474?selectedIssue=MOSD-468&issueViewSection=comments

## Checklist

### For new field types:

### DIFF
--- a/src/marqo/core/monitoring/statsd_middleware.py
+++ b/src/marqo/core/monitoring/statsd_middleware.py
@@ -8,22 +8,18 @@ from starlette.responses import Response
 
 from marqo.core.monitoring.statsd_client import StatsDClient
 
-_SEARCH_RE = re.compile(r"/indexes/[^/]+/search$")
 _DOCS_RE = re.compile(r"/indexes/[^/]+/documents$")
 _DOCUMENT_ID_RE = re.compile(r"(/documents/)[^/]+")
 
 
 class StatsDMiddleware(BaseHTTPMiddleware):
     """
-    Emits the former Reverse-Proxy (RP) CloudWatch metrics via StatsD.
+    Emits high-cardinality generic metrics.
 
-    Metrics implemented (parity with RP):
-      • requests.completed                counter   • status_code
-                                              also  • path, method, status_code
-      • marqo_processing_time             timing    (no tags)
-      • search_processing_time            timing    (no tags)
-      • index_processing_time             timing    (no tags)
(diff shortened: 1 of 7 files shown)


src/marqo/core/monitoring/statsd_middleware.py
```diff
@@ -8,22 +8,18 @@
 
 from marqo.core.monitoring.statsd_client import StatsDClient
 
-_SEARCH_RE = re.compile(r"/indexes/[^/]+/search$")
```
edfoh, line 11: is something else emitting search metrics?

ommirzaei: Yes, in the new reverse-proxy-sidecar we will parse metrics emitted by Marqo and emit specific metrics for search and add_doc with low cardinality for CW. I will raise the changes for the reverse-proxy-sidecar in a follow-up PR.

papa99do (approved):

src/marqo/core/monitoring/statsd_middleware.py
```diff
@@ -38,44 +34,30 @@ async def dispatch(self, request: Request, call_next):
...
+            "path": path_tag,
+            "method": request.method,
+            "status_code": str(response.status_code),
+        }
+
+        # latency
+        self.statsd.timing("request.duration_ms", duration_ms, tags=tags)
+
+        # batch outcome counters
+        if _DOCS_RE.fullmatch(request.url.path) and request.method in {"POST", "PATCH", "GET"}:
```
papa99do, line 48: nit. please test the performance of regex matching for very long path (for example 4KB?). Just to make sure this won't overwhelm CPU to cause performance degradation.

edfoh (approved):

src/marqo/core/monitoring/statsd_middleware.py
ommirzaei: It uses [^/]+ which is a greedy match for index name anchored at end with $, so I think it's not a perf concern, but I'll do a quick perf test to make sure.

PR #1273, opened 2025-07-29
Implement configurable stemming

farshidz:
## Change Summary
Implement configurable text field stemming.

This implementation is similar to language.
* Stemming can be configured per text field (and only for text fields). It's configured in the Vespa schema.
* Stemming cannot be changed once the field has been created
* Unlike language, search API does not support changing stemming. It's picked by Vespa based on the fields being searched.

Note only full text search is stemmed. Filtering isn't affected.

## Related Jira Ticket
https://s2search.atlassian.net/browse/MOSD-451

## Checklist

### For new field types:


tests/unit_tests/marqo/core/unstructured_vespa_index/test_unstructured_validation.py
```diff
@@ -76,9 +76,9 @@ def test_validate_mappings_object_invalid_cases(self):
                 "exception_type": InvalidArgError
             },
             {
-                "description": "text field missing language",
+                "description": "text field missing language and stemming",
                 "mapping": {"field": {"type": "text_field"}},
-                "expected_error": "'language' is a required property",
+                "expected_error": "not valid under any of the given schemas",
```
farshidz, line 81: Noting the json schema validation is probably something to replace. Error messages are difficult to customize and aren't very good.

src/marqo/core/models/marqo_index.py
```diff
@@ -63,6 +63,13 @@ class FieldFeature(Enum):
     Filter = 'filter'
 
 
+class Stemming(str, Enum):
+    None_ = 'none'
```
wanliAlex, line 77: Why do we want to explicitly accept the string `"none"` here? We never had this for other API fields?

src/marqo/core/semi_structured_vespa_index/semi_structured_add_document_handler.py
```diff
@@ -104,27 +105,53 @@ def _get_field_language(self, field_name):
             return None
 
         if field_mapping.get('type') == 'text_field':
-            if not self.marqo_index.index_supports_language:
+            language = field_mapping.get('language')
```
wanliAlex, line 149: Do we validate the value of language later? Or is this a future work?

src/marqo/core/semi_structured_vespa_index/semi_structured_add_document_handler.py
```diff
@@ -104,27 +105,53 @@ def _get_field_language(self, field_name):
...
                     f'Language is only supported for indexes created with Marqo version '
                     f'{constants.MARQO_LANGUAGE_MINIMUM_VERSION} or later. This index was created with  '
                     f'Marqo {self.marqo_index.marqo_version}.'
                 )
-            return field_mapping.get('language')
+            return language
 
         return None
 
+    def _get_field_stemming(self, field_name):
```
wanliAlex: Return type, `Optional[Stemming]` ?

tests/api_tests/v1/tests/api_tests/test_stemming.py
```diff
@@ -0,0 +1,219 @@
...
+        # Try to add document with different stemming config for same field
+        docs2 = [{"_id": "change_test2", "title": "Second document"}]
+        mappings2 = {"title": {"type": "text_field", "language": "en", "stemming": "shortest"}}
+
+        response2 = self.client.index(self.semi_structured_index_name).add_documents(
+            docs2, mappings=mappings2, tensor_fields=[]
+        )
+
+        # Should have errors
+        self.assertTrue(response2['errors'])
```
wanliAlex, line 168: Should the whole request be rejected here? Should a 400 error be raised instead of an individual document error?

wanliAlex (changes requested):
Can you add a test for relevance cut-off to ensure the language, stemming works for the probe lexical search?

src/marqo/core/models/marqo_index.py
farshidz: None string means no stemming which differs from the default, which is best. Do you think the name is confusing?

src/marqo/core/semi_structured_vespa_index/semi_structured_add_document_handler.py
farshidz: Future work.

tests/api_tests/v1/tests/api_tests/test_stemming.py
farshidz: This is consistent with language behavior. Also some documents in the batch could succeed if they don't have a value for this field, or if they have a non-string value.

src/marqo/core/models/marqo_index.py
wanliAlex: I was not clear about the default behaviour. Is my understanding correct that:
```
None -> default behaviour
Not providing -> default behaviour
`"none"` -> none behaviour
```

farshidz: Yes. Is it confusing?

src/marqo/core/semi_structured_vespa_index/semi_structured_add_document_handler.py
farshidz: Added. It's a string

tests/api_tests/v1/tests/api_tests/test_language.py
```diff
@@ -175,6 +178,43 @@ def test_field_language_override(self):
...
 
+    def test_facets_and_relevance_cutoff(self):
+        self.populate_index()
+
+        cases = [
+            ("pt", True),
+            ("en", False)
+        ]
+
+        for language, matches in cases:
```
wanliAlex, line 189: Can you also assert on the `probeCandidates`? This is the number of documents returned from the probe lexical search before the relevance cut-off counting

src/marqo/core/models/marqo_index.py
wanliAlex: I thought `"none" `was the same as default before. Since it is not the case, this looks good.

tests/api_tests/v1/tests/api_tests/test_language.py
farshidz: Addressed

wanliAlex (approved):
LGTM

PR #1276, opened 2025-08-06
Support collapse fields (PR 1/2)

papa99do:
### Add CollapseField support to MarqoIndex core models

  This PR implements Phase 1 of the collapse fields (variant grouping) feature by adding core data model support and document validation for collapse fields in semi-structured indexes.

### Key Changes

  **Core Models & Validation**:
  - Added CollapseField model with name validation and minGroups parameter (default: 500)
  - Extended SemiStructuredMarqoIndex with collapse_fields property and validation (exactly one field required)
  - Added is_collapse_field() method to check if a field is a collapse field
  - Updated API models (IndexSettings, UnstructuredMarqoIndexRequest) with collapseFields support

  **Document Processing**:
  - Document Addition Validation: Added validation in SemiStructuredAddDocumentsHandler to ensure collapse fields are:
    - Present in all documents
    - String type only
    - Non-empty/non-whitespace values
  - Collapse Field Handling: Updated SemiStructuredVespaDocument to process collapse fields without requiring them in the lexical field map
  - Lexical Field Exclusion: Excluded collapse fields from automatic lexical field creation during document processing

  **Optimizations**:
  - Added check to skip Vespa operations when no valid documents remain after validation failures

## Checklist

### For new field types:


src/marqo/core/semi_structured_vespa_index/semi_structured_add_document_handler.py
```diff
@@ -43,6 +43,43 @@ def __init__(self, marqo_index: SemiStructuredMarqoIndex, add_docs_params: AddDo
...
+        if not self.marqo_index.collapse_fields:
+            return  # No collapse fields configured
+        
+        collapse_field = self.marqo_index.collapse_fields[0]  # Only one allowed per spec
+        collapse_field_name = collapse_field.name
+
+        # TODO confirm if all these validations are required
+        if collapse_field_name not in doc:
+            raise AddDocumentsError(
+                f"Document missing required field '{collapse_field_name}'. "
```
farshidz, line 65: Why not allow docs without this value? Not every search utilizes this field.

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_schema_template_2_16.sd.jinja2
```diff
@@ -295,6 +338,26 @@ schema {{ index.schema_name }} {
             expression: modify(lexical_score(), query(marqo__mult_weights_lexical), query(marqo__add_weights_lexical))
         }
     }
+
+    {% if collapse_field -%}
```
farshidz, line 342: Let's make sure we have backwards compatibility tests that add fields after prepare and in the test, so we know this doesn't break anything.

src/marqo/core/models/marqo_index_request.py
```diff
@@ -41,6 +41,17 @@ class UnstructuredMarqoIndexRequest(MarqoIndexRequest):
     treat_urls_and_pointers_as_images: bool
     treat_urls_and_pointers_as_media: bool
     filter_string_max_length: int
+    collapse_fields: Optional[List[marqo_index.CollapseField]] = None
+
+    @root_validator
+    def validate_collapse_fields(cls, values):
+        collapse_fields = values.get('collapse_fields')
```
farshidz, line 48: I think you need a version check because the only jinja template won't support this.

tests/integ_tests/tensor_search/integ_tests/test_collapse_fields.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+class TestCollapseFields(MarqoTestCase):
+    """Integration tests for collapse fields functionality."""
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        super().setUpClass()
+
+        default_text_index = cls.unstructured_marqo_index_request(
```
farshidz, line 17: Test with 2.16 and also pre 2.16 indexes.

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_schema_template_2_16.sd.jinja2
papa99do: good point. will add a compatibility test in the next PR

tests/integ_tests/tensor_search/integ_tests/test_collapse_fields.py
papa99do: What test case do you have in mind?

src/marqo/core/semi_structured_vespa_index/semi_structured_add_document_handler.py
papa99do: confirmed with team that these validation are still required. see [this thread](https://marqo-ai.slack.com/archives/C08SRDY38CD/p1754542987908689) for details

src/marqo/core/models/marqo_index_request.py
papa99do: we'll never try to create an old index with collapse_fields since it is only supported in 2.23

farshidz (approved):

PR #1278, opened 2025-08-07
[Refactoring] Move facets, sortBy and relevanceCutoff query parameters to SemiStructuredVespaIndex

papa99do:
## Change Summary
A refactoring to move the following code from structured_vespa_index to semi_structured_vespa_index:
- facets query
- sort-by and relevance cutoff 

These pieces of logic are only used in semi-structured vespa index. They were put in the structured one due to the original design of  semi-structured vespa index which inherits from structured index. This design leads to more confusion than convenience, and will be revised in future refactoring. 

For now, this refactoring is required to add collapse_fields support without polluting structured vespa index code. 

## Checklist

### For new field types:

### DIFF
--- a/src/marqo/core/structured_vespa_index/structured_vespa_index.py
+++ b/src/marqo/core/structured_vespa_index/structured_vespa_index.py
@@ -571,49 +571,6 @@ class StructuredVespaIndex(VespaIndex):
 
         tensor_yql = f'select {select_attributes} from {self._marqo_index.schema_name} where {tensor_term}{filter_term}'
         lexical_yql = f'select {select_attributes} from {self._marqo_index.schema_name} where ({lexical_term}){filter_term}'
-        facet_queries = None
-
-        if marqo_query.facets or marqo_query.track_total_hits:
-            facets_query_skeleton = '%s limit 0 | %s'
-            QUERY_DELIMITER = "\n---MARQO-YQL-QUERY-DELIMITER---\n"
-            unique_exclusions = []
-            facet_queries = []
-            facets_lexical_term = self._get_lexical_search_term(marqo_query, is_facets_term=True)
-            base_yql = f'select {select_attributes} from {self._marqo_index.schema_name} where {facets_lexical_term}'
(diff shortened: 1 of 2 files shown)


farshidz (commented):
How confident are you this isn't breaking any structured features? Structures is probably not as well tested as semi. Also could there be features we unintentionally has exposed in structured? e.g. did facets actually work?

papa99do:
Good point. I didn't change any existing logic in structured, just remove facets, sort by and relevance cutoff handling which is not used in structured index in the first place. So I'm quite confident this won't break structured index. The risk of breaking structured is higher if I leave the logic there and add more logic about collapse field in the same place. 

I'm pretty sure facets logic is not used in structured since it calls the method `_get_facets_term` which only exists in `semi_structured_vespa_index.py`. This (calling a non-existent method in the base class) magically works due to python's weak type check. 😞  As for the sort-by and relevance-cutoff feature, I confirmed with @wanliAlex that they are only used for semi-structured index as well.

src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -571,49 +571,6 @@ def _to_vespa_hybrid_query(self, marqo_query: MarqoHybridQuery) -> Dict[str, Any
 
         tensor_yql = f'select {select_attributes} from {self._marqo_index.schema_name} where {tensor_term}{filter_term}'
         lexical_yql = f'select {select_attributes} from {self._marqo_index.schema_name} where ({lexical_term}){filter_term}'
-        facet_queries = None
```
wanliAlex, line 574: Maybe add a sanity check for these unsupported parameters to avoid silent failure?

farshidz (approved):

src/marqo/core/structured_vespa_index/structured_vespa_index.py
papa99do: The validation is already done at an early stage, see https://github.com/marqo-ai/marqo/blob/acd8f4e1507c4407b5f5a6d00700de1f7d39ddd8/src/marqo/core/search/hybrid_search.py#L147

wanliAlex (approved):

PR #1280, opened 2025-08-11
Fix the flaky api test

papa99do:
### increase the cache_reading time to fix the flaky API tests
Sometimes, the latency exceeds 2ms even when cache is hit during tests. but it seldom exceeds 3ms, and we'll average 10 runs, so 3ms should be safe enough.  here's a sample failed build: https://github.com/marqo-ai/marqo/actions/runs/16867480531/job/47776381301

## Checklist

- [N/A] Tests have been added for changes

### For new field types:


papa99do:
This will be addressed in https://github.com/marqo-ai/marqo/pull/1285

PR #1282, opened 2025-08-14
Fix a bug that causes wrong facets count when we use lexical retrieval method with filter and multiple 'OR' phrases.

papa99do:
## Change Summary
The lexical term used in the facets query is missing wrapping parentheses, which will cause facets query to return wrong count when we search with filter and multiple words in the query (multiple 'OR' phrases). 
This is due to that the OR phrases without  parentheses breaks the query semantics and returns more results than expected. 
The fix is simple, just to add the parentheses around lexical term nevertheless. 

## Checklist

### For new field types:


farshidz (approved):
Is there an

PR #1289, opened 2025-08-21
Implement typeahead search

farshidz:
## Change Summary
Implement typeahead search. (Phase1)

The typeahead API can:
* Index popular queries to suggest from
* Generate suggestions from partial or full user input

Suggestion are generated in two stages:
1. Retrieval: Match all indexed queries where at least one input word matches a prefix of at least one query word, within a configured edit distance.
2. Rank by a score function consisting of query popularity and the match BM25 score.

We will implement this feature in two phases, and this PR only covers phase 1.
### Phase 1 (in this PR): Basic features
  * Add a new schema for typeahead, during index creation time
  * Indexing with query normalisation
  * Prefix query with fuzziness support, tunable weight on bm25 score and popularity
  * Support random number fields in a metadata map fields (for display purpose only, e.g. hit count, etc.)
  * Timestamp field to track the last updated epoch time
  * Add the schema to existing indexes at bootstrapping time. 

### Phase 2 (future work): Advanced features
  * Support score modifiers based on the metadata field, so user can define ranking based on random fields or multiple fields.
  * Support Vespa garbage collection mechanism to automatically delete queries based on a sliding window.. This is to facilitate any ETL pipeline. 
  * Support schema upgrade (just for typeahead schema) to support new typeahead features for existing indexes
  * (TBD) Configurable/Smart normalisation?

## Related Jira Ticket
https://s2search.atlassian.net/browse/MOSD-394

## Checklist

### For new field types:


src/marqo/core/index_management/index_management.py
```diff
@@ -171,13 +173,23 @@ def batch_create_indexes(self, marqo_index_requests: List[MarqoIndexRequest]) ->
                 request.model.text_chunk_prefix = request.model.get_default_text_chunk_prefix()
 
             schema, marqo_index = vespa_schema_factory(request).generate_schema()
-            index_to_create.append((schema, marqo_index))
-            logger.debug(f'Creating index {str(request.name)} with schema:\n{schema}')
+            logger.debug(f'Creating index {request.name} with schema:\n{schema}')
+
+            typeahead_schema, updated_marqo_index = TypeaheadVespaSchema(marqo_index).generate_schema()
```
wanliAlex, line 178: The typeahead schema is mandatory not optional right?

src/marqo/core/models/marqo_index.py
```diff
@@ -287,6 +287,7 @@ class MarqoIndex(ImmutableBaseModel, ABC):
     """
     name: str
     schema_name: str
+    typeahead_schema_name: Optional[str] = None
```
wanliAlex, line 290: Optional for backwards compatibility right?

src/marqo/core/models/typeahead.py
```diff
@@ -0,0 +1,129 @@
...
+    bm25_weight: Optional[float] = Field(
+        default=None,
+        alias="bm25Weight",
+        description="Weight for BM25 score in ranking"
+    )
+
+    @field_validator('q')
+    def validate_q(cls, v: str) -> str:
+        if not v or not v.strip():
+            raise ValueError("q is required")
```
wanliAlex: nit: q is required and must not be an emtpy string

src/marqo/core/models/typeahead.py
```diff
@@ -0,0 +1,129 @@
+"""Pydantic models for typeahead API requests and responses."""
+
+from typing import List, Optional, Dict
+
+from pydantic import Field, field_validator
```
wanliAlex, line 5: These are pydantic 2 right?

src/marqo/core/typeahead/typeahead.py
```diff
@@ -0,0 +1,315 @@
...
+            request: TypeaheadRequest containing all parameters
+            
+        Returns:
+            TypeaheadResponse with suggestions and processing time
+        """
+        start_time = timer()
+
+        # Check if index exists and get typeahead schema name
+        from marqo.tensor_search import index_meta_cache
+        marqo_index = index_meta_cache.get_index(index_management=self.index_management, index_name=index_name)
```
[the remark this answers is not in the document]
wanliAlex: Good catch!

src/marqo/core/typeahead/typeahead.py
```diff
@@ -0,0 +1,315 @@
...
+        if not normalized_input:
+            return TypeaheadResponse(suggestions=[])
+
+        # Tokenize by whitespace
+        tokens = normalized_input.split()
+        if not tokens:
+            return TypeaheadResponse(suggestions=[])
+
+        # Build YQL query conditions for each token
+        retrieval_terms = []
```
wanliAlex: Should we break these logics into small methods?

src/marqo/core/typeahead/typeahead.py
```diff
@@ -0,0 +1,315 @@
...
+            if len(token) < request.min_fuzzy_match_length:
+                # Use exact prefix matching for short tokens
+                retrieval_terms.append(
+                    f"query_words contains ({{prefix:true}}\"{token}\")"
+                )
+            else:
+                # Use fuzzy matching for longer tokens
+                retrieval_terms.append(
+                    f"query_words contains "
+                    f"({{maxEditDistance:{request.fuzzy_edit_distance}, prefix:true}}fuzzy(\"{token}\"))"
```
wanliAlex: Have we considered using `prefixLength: n` when building the query? Reference here: https://docs.vespa.ai/en/text-matching.html#fuzzy-prefix-match

src/marqo/core/typeahead/typeahead.py
```diff
@@ -0,0 +1,315 @@
...
+                )
+
+            ranking_terms.append(f"query_index contains \"{token}\"")
```
wanliAlex: As suggested here: https://docs.vespa.ai/en/text-matching.html#fuzzy-prefix-match

Should we add a check between len(token) and request.fuzzy_edit_distance to avoid matching every document?

src/marqo/core/typeahead/typeahead.py
```diff
@@ -0,0 +1,315 @@
...
+            "hits": request.limit,
+            "ranking": "suggestions-rank-profile"
+        }
+
+        # Add query features if weights are provided
+        query_features = {}
+        if request.popularity_weight is not None:
+            query_features["popularity_weight"] = request.popularity_weight
+        if request.bm25_weight is not None:
+            query_features["bm25_weight"] = request.bm25_weight
```
wanliAlex, line 100: Add a comment here to explain that bm25_weight is used to boost exact match results?

src/marqo/tensor_search/api.py
```diff
@@ -737,6 +739,94 @@ def schema_validation(index_name: str, settings_object: dict):
     )
 
 
+# TODO verify: not throttling. Would require a separate category, but we're not using Marqo throttling anymore?
```
wanliAlex: Need @Omid confirmation here.

src/marqo/tensor_search/api.py
```diff
@@ -737,6 +739,94 @@ def schema_validation(index_name: str, settings_object: dict):
...
+
+@app.get("/indexes/{index_name}/suggestions/stats")
+def get_typeahead_stats(index_name: str, marqo_config: config.Config = Depends(get_config)):
+    """
+    Get statistics about the typeahead queries for an index.
+    
+    Args:
+        index_name: Name of the index to get stats for
+    """
+    try:
```
wanliAlex: Why do we catch error here?

tests/api_tests/v1/tests/api_tests/test_typeahead.py
```diff
@@ -0,0 +1,466 @@
...
+                self.assertIn(response.status_code, [400, 422])
+
+    def test_index_queries_with_invalid_format(self):
+        """Test indexing queries with invalid format."""
+        invalid_requests = [
+            {},  # Missing queries
+            {"queries": "not a list"},  # queries is not a list
+            # {"queries": [{"query": "test"}]},  # Missing popularity is allowed, default is 1.0
+            {"queries": [{"popularity": 1.0}]},  # Missing query
+            {"queries": [{"query": "", "popularity": 1.0}]},  # Empty query
```
wanliAlex, line 54: Other invalid formats:
1. An unsupported field in the query;
2. Unsupported format for popularity, metadata

tests/api_tests/v1/tests/api_tests/test_typeahead.py
```diff
@@ -0,0 +1,466 @@
+import json
+import uuid
+
+import requests
+from tests.marqo_test import MarqoTestCase
+
+
+class TestTypeahead(MarqoTestCase):
```
wanliAlex, line 8: Do we need to test the delete index will also delete the type_ahead index?

wanliAlex (changes requested):

tests/api_tests/v1/tests/api_tests/test_typeahead.py
wanliAlex, line 1: Do we need backwards compatibility tests?

src/marqo/core/index_management/index_management.py
```diff
@@ -219,23 +231,25 @@ def update_index(self, marqo_index: SemiStructuredMarqoIndex) -> None:
...
-            raise InternalError(f'Index {marqo_index.name} created by Marqo version {marqo_index.marqo_version} '
-                                f'can not be updated.')
-
-        def is_subset(dict_a, dict_b):
-            # check if dict_a is a subset of dict_b
-            return all(k in dict_b and dict_b[k] == v for k, v in dict_a.items())
-
-        if (is_subset(marqo_index.tensor_field_map, existing_index.tensor_field_map) and
-                is_subset(marqo_index.field_map, existing_index.field_map) and
+        # !!! Please note that we need to acquire the lock before retrieving the index setting so that we know the
```
wanliAlex, line 234: Was this part changed in a recent PR and you revert the change here?

src/marqo/core/models/typeahead.py
```diff
@@ -0,0 +1,129 @@
...
+from marqo.core.exceptions import InvalidArgumentError
+from marqo.tensor_search.enums import EnvVars
+from marqo.tensor_search.utils import read_env_vars_and_defaults_ints
+
+
+class TypeaheadRequest(ImmutableStrictBaseModelV2):
+    """Request model for typeahead suggestions."""
+
+    q: str = Field(..., description="Partial user search input")
+    limit: int = Field(default=10, ge=0, description="Maximum number of suggestions to return")
```
wanliAlex: Can limit be 0?

src/marqo/core/models/typeahead.py
```diff
@@ -0,0 +1,129 @@
...
+            raise ValueError("q is required")
+        return v.strip()
+
+
+class TypeaheadSuggestion(ImmutableStrictBaseModelV2):
+    """Individual suggestion in typeahead response."""
+
+    suggestion: str = Field(..., description="The suggested query text")
+    score: float = Field(..., alias="_score", description="Relevance score for the suggestion")
+    metadata: Optional[dict] = Field(default=None, description="Additional metadata")
```
wanliAlex: nit: Dict[str, float]

wanliAlex: Do we have an upper bound for the metadata value here? Remember someone used a very large integer number in the score modifier fields before?

src/marqo/core/typeahead/typeahead_vespa_schema_template.sd.jinja2
```diff
@@ -0,0 +1,56 @@
...
+        field last_updated_at type long {
+            indexing: summary | attribute
+            attribute: fast-access
+        }
+    }
+
+    fieldset default {
+        fields: query_words
+    }
+
```
wanliAlex, line 45: Is a document-summary needed here? E.g., do we need to retrieve `field query_words` in the search results?

wanliAlex (changes requested):

src/marqo/core/models/marqo_index.py
farshidz: Yes

src/marqo/core/typeahead/typeahead.py
farshidz: We did. The problem with `prefixLength: n` is we get zero fuzziness for that length. With this approach, once the input is long enough, we handle fuzziness/typos anywhere.

src/marqo/core/typeahead/typeahead.py
farshidz: Are you suggesting we infer min fuzzy length from the edit distance? Right now we're leaving it to the user to make sure the two parameters are configured correctly.

src/marqo/core/typeahead/typeahead.py
```diff
@@ -0,0 +1,315 @@
...
+
+        # Create single YQL query that ORs all token conditions
+        yql_retrieval = " OR ".join(retrieval_terms)
+        yql_ranking = " OR ".join(ranking_terms)
+        yql = (f"SELECT * FROM {typeahead_schema_name} WHERE rank({yql_retrieval}, {yql_ranking})")
+
+        search_params = {
+            "yql": yql,
+            "hits": request.limit,
+            "ranking": "suggestions-rank-profile"
```
farshidz, line 91: Should this be a constant somewhere? I wonder what we do for tensor search (I'll check later)

src/marqo/core/index_management/index_management.py
papa99do: correct. all index created by marqo post 2.24.0 will have it.

src/marqo/core/models/marqo_index.py
papa99do: Please note this is forward compatible as well

src/marqo/core/models/typeahead.py
papa99do: correct. we use pydantic v2 since it does not depend on any existing pydantic v1 model

tests/api_tests/v1/tests/api_tests/test_typeahead.py
papa99do: Yes, that is covered in integration test `test_index_management.py` already. Won't test in API test since we probably should not rely on Vespa in API test cases

tests/api_tests/v1/tests/api_tests/test_typeahead.py
papa99do: We might need. I'll probably do this in another PR. We are missing quite a few compatibility tests now

src/marqo/core/index_management/index_management.py
papa99do: Yes. the local test kept failing without this change. I think it makes sense to do the check when holding the lock. @farshidz I think you did this change before, do you mind me changing it back?

src/marqo/core/models/typeahead.py
papa99do: I don't think so, I'll change to `gt=0`

src/marqo/core/typeahead/typeahead_vespa_schema_template.sd.jinja2
papa99do: good point. @farshidz  why do we need `indexing: summary` for query_words? We will not need to retrieve them back, right?

src/marqo/tensor_search/api.py
papa99do: good point. we don't need to, will remove

src/marqo/tensor_search/api.py
papa99do: confirmed with Omid that throttling is done in Haproxy in new indexes: https://marqo-ai.slack.com/archives/C060W79JNH3/p1756773713321269?thread_ts=1756773247.240749&cid=C060W79JNH3

papa99do: TODO: Will mention in the Cloud migration guide about throttling mechanism of this endpoint

src/marqo/core/index_management/vespa_application_package.py
```diff
@@ -744,6 +757,38 @@ def _persist_index_settings(self):
...
+        index_setting_changed = False
+
+        for _, index in self._index_setting_store._index_settings.items():
+            if (index.typeahead_schema_name is not None or
+                    index.parsed_marqo_version() < MARQO_TYPEAHEAD_SCHEMA_MINIMUM_VERSION):
+                # already has typeahead schema or the version is too old
+                continue
+
+            # Generate typeahead schema
+            typeahead_schema, updated_index = TypeaheadVespaSchema(index).generate_schema()
```
farshidz, line 777: Note we log the schema in debug when creating the index. Should do the same here

src/marqo/core/models/typeahead.py
papa99do: I don't think we need a upper bound here. the typeahead index ingestion is an internal process

src/marqo/core/typeahead/typeahead.py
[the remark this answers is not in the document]
papa99do: Given get_suggestion usually has higher load, we use the index_settings from the cache. Other places uses the index_setting directly from Vespa container, which is the source of truth.

src/marqo/core/typeahead/typeahead.py
papa99do: I'm OK with the size of this method. Will refactor if we need to add more logic to it.

src/marqo/core/typeahead/typeahead.py
papa99do: I'm OK to leave it as is. When more rank profiles are added, we can extract this to a typeahead_constants.py file.

tests/api_tests/v1/tests/api_tests/test_typeahead.py
papa99do: added two cases. metadata is not part of the suggestion request. More errors will be covered in unit test cases.

src/marqo/core/typeahead/typeahead.py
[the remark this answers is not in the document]
farshidz: This is intentional and consistent with the rest of the code base. We use the cache only for latency sensitive functionality.

src/marqo/core/typeahead/typeahead.py
```diff
@@ -0,0 +1,340 @@
...
+            vespa_docs.append(vespa_doc)
+
+        if vespa_docs:
+            response = self.vespa_client.feed_batch(vespa_docs, schema=typeahead_schema_name)
+            for resp in response.responses:
+                doc_id = resp.id.split('::')[-1] if resp.id else None
+                query = doc_id_query_map.get(doc_id, None)
+                status, message = self.vespa_client.translate_vespa_document_response(resp.status, message=resp.message)
+                if status != 200:
+                    errors.append(TypeaheadIndexError(query=query, message=message, code=status))
```
farshidz: I think Index in `TypeaheadIndexError` and `TypeaheadIndexResponse` is ambiguous (we mean the act of indexing, but we also have a concept named index, i.e. a Marqo index). Perhaps we could call these `TypeaheadIndexingError` etc?

src/marqo/core/typeahead/typeahead.py
```diff
@@ -0,0 +1,340 @@
...
+            
+        Returns:
+            TypeaheadResponse with suggestions and processing time
+        """
+        start_time = timer()
+
+        # Check if index exists and get typeahead schema name
+        from marqo.tensor_search import index_meta_cache
+        marqo_index = index_meta_cache.get_index(index_management=self.index_management, index_name=index_name)
+        self._check_typeahead_support(marqo_index)
```
farshidz: nit you could replace these calls with (in order of preference):
* a metaclass that calls this function before any other function
* an annotation
The current approach seems prone to regression when we add future functions and forget to call this check.

src/marqo/core/typeahead/typeahead_vespa_schema_template.sd.jinja2
farshidz: That's a good point. Seems safe to remove. Also it would be best to specify what to retrieve anyway (in which case I think an extra summary has no perf effect). e.g., we don't need `last_updated_at` when returning suggestions.

src/marqo/core/typeahead/typeahead.py
```diff
@@ -0,0 +1,340 @@
...
+                    f"query_words contains "
+                    f"({{maxEditDistance:{request.fuzzy_edit_distance}, prefix:true}}fuzzy(\"{token}\"))"
+                )
+
+            ranking_terms.append(f"query_index contains \"{token}\"")
+
+        # Create single YQL query that ORs all token conditions
+        yql_retrieval = " OR ".join(retrieval_terms)
+        yql_ranking = " OR ".join(ranking_terms)
+        yql = (f"SELECT * FROM {typeahead_schema_name} WHERE rank({yql_retrieval}, {yql_ranking})")
```
farshidz: Let's specify fields to retrieve for better readability and also to make sure we don't over retrieve.

src/marqo/core/typeahead/typeahead.py
wanliAlex: OK that's fine. We just need to be clear in the documentation like what Vespa did:
```For instance, a query such as {maxEditDistance:2,prefix:true}fuzzy("XY") will end up matching every document, since all possible strings can have their prefix transformed to "XY" with at most 2 edits.```

File: src/marqo/core/index_management/index_management.py (hunk 14 above)
Comment (farshidz) [reply src/marqo/core/index_management/index_management.py:234]: Why does this make sense though?

File: src/marqo/core/typeahead/typeahead_vespa_schema_template.sd.jinja2 (hunk 17 above)
Comment (papa99do) [reply src/marqo/core/typeahead/typeahead_vespa_schema_template.sd.jinja2:45]: removed summary for both `query_words` and `last_updated_at`

File: src/marqo/core/typeahead/typeahead.py (hunk 20 above)
Comment (papa99do) [reply src/marqo/core/typeahead/typeahead.py]: changed

File: src/marqo/core/typeahead/typeahead.py (hunk 21 above)
Comment (papa99do) [reply src/marqo/core/typeahead/typeahead.py]: changed to use annotation

File: src/marqo/core/typeahead/typeahead.py (hunk 22 above)
Comment (papa99do) [reply src/marqo/core/typeahead/typeahead.py]: change to only select `query, metadata`

File: src/marqo/core/models/typeahead.py [hunk 23]
```diff
@@ -0,0 +1,125 @@
...
+from marqo.base_model import ImmutableStrictBaseModelV2, ImmutableBaseModelV2
+from marqo.core.exceptions import InvalidArgumentError
+from marqo.tensor_search.enums import EnvVars
+from marqo.tensor_search.utils import read_env_vars_and_defaults_ints
+
+
+class TypeaheadRequest(ImmutableStrictBaseModelV2):
+    """Request model for typeahead suggestions."""
+
+    q: str = Field(..., description="Partial user search input")
```
wanliAlex, line 16: Should this field be optional?

src/marqo/core/typeahead/typeahead.py
```diff
@@ -0,0 +1,321 @@
...
+        """
+        start_time = timer()
+
+        # Check if index exists and get typeahead schema name
+        from marqo.tensor_search import index_meta_cache
+        marqo_index = index_meta_cache.get_index(index_management=self.index_management, index_name=index_name)
+        typeahead_schema_name = marqo_index.typeahead_schema_name
+        query = request.q.strip()
+
+        if query == "":
```
wanliAlex, line 51: Should we also accept None?

wanliAlex (changes requested):

src/marqo/core/index_management/vespa_application_package.py
papa99do: logging added

src/marqo/core/models/typeahead.py
papa99do: No, it's still required. but it can be empty string.

wanliAlex (approved):

farshidz (commented):
lgtm

PR #1292, opened 2025-08-26
Fix total hit count when collapse field is provided [Patch 2.23]

papa99do:
## Change Summary
Change the vespa query to group the total hit count result by collapse field when it is provided

## Checklist

### For new field types:

### DIFF
--- a/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
+++ b/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
@@ -187,8 +187,11 @@ class SemiStructuredVespaIndex(StructuredVespaIndex, UnstructuredVespaIndex):
 
         if marqo_query.track_total_hits is not None:
             # 0 is byte representation of letter "t"
-            facet_queries.append(facets_query_skeleton % (
-            f'{base_yql}{filter_term}', f"all(group({self._TOTAL_HITS_GROUP_CONST}) each(output(count())))"))
+            if marqo_query.collapse_field_name:
+                total_hit_query = f"all(group({self._TOTAL_HITS_GROUP_CONST}) each(group({marqo_query.collapse_field_name}) output(count())))"
+            else:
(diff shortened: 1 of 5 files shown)


src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
```diff
@@ -187,8 +187,11 @@ def _generate_facet_queries(self, marqo_query):
 
         if marqo_query.track_total_hits is not None:
             # 0 is byte representation of letter "t"
-            facet_queries.append(facets_query_skeleton % (
-            f'{base_yql}{filter_term}', f"all(group({self._TOTAL_HITS_GROUP_CONST}) each(output(count())))"))
+            if marqo_query.collapse_field_name:
+                total_hit_query = f"all(group({self._TOTAL_HITS_GROUP_CONST}) each(group({marqo_query.collapse_field_name}) output(count())))"
+            else:
```
wanliAlex, line 192: Can this be an empty string?

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
wanliAlex, line 1: Bump the version

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
papa99do: I think so. but in that case, we'll not collapse, so logic is the same. ideally we should error out, but since this is only used internally (by ecommerce api), we didn't add that validation.

src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
papa99do: yes, bumped in another commit.

wanliAlex (approved):

PR #1303, opened 2025-09-15
Partially fix pagination for disjunction and relevance cutoff 

papa99do:
## Change Summary
Pagination in hybrid search - RRF has the wrong behavior. It currently implements pagination (using limit & offset) in the individual tensor and lexical searches, then performs the RRF algorithm on those 2 result lists. 

This PR fixes this behaviour by retrieving `0 to offset + limit` results in individual tensor and lexical searches, fusing them, applying global score modifiers, then trimming. In this way, we can reduce the number of duplicates and missing docs. 

The final solution to eliminate duplicates and missing docs requires storing pagination state (docs in previous pages) and excluding them from the search result. This will be addressed in [a separate PR](https://github.com/marqo-ai/marqo/pull/1234/files#top).

## Changes made in this PR
- For disjunction search without relevance cutoff or sorting, fetch offset + limit results from the beginning (offset=0) instead of applying offset before fusion.
- Increase rerankDepthGlobal by the `offset` to ensure score modifiers reach the current page's documents.
- Trim the result of disjunction search and relevance cutoff search to `(offset, offst + limit)` after applying the global modifiers

## Related Jira Ticket
https://s2search.atlassian.net/browse/MOSD-311

## Checklist

### For new field types:

### DIFF
--- a/vespa/src/main/java/ai/marqo/search/HybridSearcher.java
+++ b/vespa/src/main/java/ai/marqo/search/HybridSearcher.java
@@ -431,6 +431,17 @@ public class HybridSearcher extends Searcher {
 
         // Validate input parameters
         if (!isRelevanceCutoffEnabled && !isSortByEnabled) {
+            // For disjunction hybrid search, we need to set offset to 0 to handle the pagination
+            // correctly
+            // TODO the logic to extract retrievalMethod is duplicated here to reduce the effort of
+            //   changing method signature. We will do a refactoring in the near future
+            String retrievalMethod =
+                    query.properties().getString("marqo__hybrid.retrievalMethod", "");
+            if ("disjunction".equalsIgnoreCase(retrievalMethod)) {
+                query.setHits(query.getOffset() + query.getHits());
+                query.setOffset(0);
+            }
+
             return query;
         }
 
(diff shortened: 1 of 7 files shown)


danyilq (approved):

vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -415,22 +421,30 @@ List<Future<Result>> getFacetsFutureList(
...
+     * @param isDisjunctionSearch      whether the retrieval method is disjunction
      * @return The updated query with new hits, offsets, and targetHits.
      */
     public Query updateQueryHitsOffsetsAndTargetHits(
             Query query,
             Integer relevantCandidates,
             Integer sortByMinSortCandidates,
             boolean isRelevanceCutoffEnabled,
-            boolean isSortByEnabled) {
+            boolean isSortByEnabled,
```
farshidz, line 438: Have you double checked any other code paths that depend on this? e.g. do facets depend on this at all?

papa99do: Yes. I've tested on KC's data that the lexical-lexical (Avg rank mismatch: 0.000828) and tensor-tensor (Avg rank mismatch: 0.003404) result are almost the same before and after this fix (based on top 100 quries). I also checked the code path that the pagination of lexical and tensor retrieval method will not be impacted by this change.

PR #1360, opened 2026-02-06
Add Vespa convergence checks to add_documents flow

papa99do:
### Context
Adding documents to semi-structured index with new string and string array fields will trigger a Vespa app package deployment process (with updated schema and index settings). It takes some time for Vespa cluster to become converged. Adding document to a un-converged Vespa cluster might result in partial document/missing fields. To prevent this from happening, we need to do a convergence check before processing an add-document batch. 

### Changes in this PR
- Add a convergence check before retrieving the index setting to make sure we won't get stale setting. We use an env var `MARQO_ENABLE_ADD_DOCUMENTS_CONVERGENCE_CHECK` (default `TRUE`) to guard this check so we disable this check if needed. 
  - Please note I didn't add it before feeding the docs to Vespa since it is redundant.  If we add new fields, we already wait for the convergence. If we don't need to add new fields, that means the schema and index setting from a converged Vespa cluster already has all the fields we need.
- Throw a `VespaNotConvergedError` if convergence check fails. It also log out detailed information about the convergence check. The exception is mapped to `OperationConflictError` (409) error in API layer.
- Improve the convergence waiting logic
  - Introduce an env var `MARQO_VESPA_CONVERGENCE_TIMEOUT` (default 120) to allow configure the total time of waiting for convergence after triggering an app deployment
  - Use a smarter backoff strategy when retrying
  - Log warnings for slow convergence (takes >10s), including non-converged service details to capture useful information for investigating issues
- Fix a minor issue that the updated_at timestamp of index settings is never updated. This fix will capture the timeline info of schema change to help us investigate app deployment issues.

## Checklist


src/marqo/vespa/vespa_client.py
```diff
@@ -41,10 +41,22 @@ class VespaClient:
...
             self.converged = converged
+            self.non_converged_services = non_converged_services or []
+
+        def to_dict(self) -> Dict[str, Any]:
+            result = {
+                'currentGeneration': self.current_generation,
+                'wantedGeneration': self.wanted_generation,
+                'converged': self.converged,
+            }
+            if self.non_converged_services:
```
wanliAlex: nit: Do we need to always return this attribute in the response? We can make it an empty list if all services are converged.

wanliAlex (commented):
LGTM, overall

src/marqo/api/configs.py
```diff
@@ -30,6 +30,8 @@ def default_env_vars() -> dict:
         EnvVars.MARQO_MAX_STRING_ARRAY_FIELD_COUNT_UNSTRUCTURED: 100,
         EnvVars.MARQO_MAX_LEXICAL_FIELD_COUNT_UNSTRUCTURED: 100,
         EnvVars.MARQO_INDEX_DEPLOYMENT_LOCK_TIMEOUT: 5,  # index operations acquire this distributed lock with a timeout
+        EnvVars.MARQO_VESPA_CONVERGENCE_TIMEOUT: 120,
+        EnvVars.MARQO_ENABLE_ADD_DOCUMENTS_CONVERGENCE_CHECK: "TRUE",
```
farshidz, line 34: Do we really need this? We don't flag other convergence checks

src/marqo/core/document/document.py
```diff
@@ -39,6 +41,11 @@ def __init__(self, vespa_client: VespaClient, index_management: IndexManagement,
 
     def add_documents(self, add_docs_params: AddDocsParams,
                       field_count_config=SemiStructuredFieldCountConfig()) -> MarqoAddDocumentsResponse:
+        # Check Vespa convergence before reading index settings to avoid stale schema
+        if tensor_search_utils.read_env_vars_and_defaults(
+                EnvVars.MARQO_ENABLE_ADD_DOCUMENTS_CONVERGENCE_CHECK).lower() == 'true':
+            self.vespa_client.check_for_application_convergence()
```
farshidz, line 53: The problem with checking here, as opposed to after reading the settings, is fields could get added between the check and settings read. So fields seen during settings read could still belong to a non-converged change.

src/marqo/tensor_search/api.py
```diff
@@ -174,6 +173,13 @@ def marqo_base_exception_handler(request: Request, exc: base_exceptions.MarqoErr
             "Vector store request timed out. Try your request again later.",
             None
         ),
+        (
+            vespa_exceptions.VespaNotConvergedError,
```
farshidz, line 177: We've been returning 409 for this before, how did it work then without this change?

src/marqo/vespa/vespa_client.py
```diff
@@ -193,25 +209,68 @@ def get_application_has_converged(self) -> bool:
...
+                        )
+                    else:
+                        logger.debug('Waiting for Vespa application to converge')
             # TODO Find out what exceptions is raised here
             except (httpx.TimeoutException, httpcore.TimeoutException):
                 logger.error("Marqo timed out waiting for Vespa application to converge. Will retry.")
 
-        raise VespaError(f"Vespa application did not converge within {timeout} seconds. "
-                         f"The convergence status is {self._get_convergence_status()}")
+            if attempt < _BACKOFF_AFTER_ATTEMPTS:
```
farshidz, line 239: You mentioned improved retry logic. Could you please clarify what we're getting out of retry changes?

src/marqo/api/configs.py
papa99do: This is my rationale for this decision:

Other checks/waits are only needed when schema changes, if not convergent, we should not proceed. On the other hand, adding documents rarely triggers an app deployment. It might hurt the availability SLA when blocked by a long Vespa operation (scaling in/out, etc.)
Ideally, we should always leave this check `ON`, but in case we need to temporarily turn it `OFF` in some extreme circumstances, we can just set this env var and rolling updates Marqo, instead of releasing a new version of Marqo.

src/marqo/core/document/document.py
papa99do: Yes, good point. I'll move the check to after the reading the settings.

src/marqo/tensor_search/api.py
papa99do: 409 is returned only for distribution lock contention, not for convergence check. Convergence check raises `VespaNotConvergedError`, which does not have a mapping in `api_exception_mappings`, and returns a default api_exceptions.MarqoWebError (500)

src/marqo/vespa/vespa_client.py
papa99do: There are a few reasons for this specific backoff logic:
- I want to print out more information for investigation, so I will log out detailed convergence status for a slow convergence (>10s based on stats).
- I want this method to detect convergence as fast as possible if it's not a slow convergence, so I set it to check every 1s for the first 8s (8s is a arbitrary number, could be 10s as well)
- I don't want it to add much load on the config servers (which seems to struggle to distribute files for some reason recently). and I don't want to print out too many redundant information in the log, hence the exponential backoff with a max backoff value. Also, these params should be adjusted based on the real data for prod indexes once we deploy this change.

src/marqo/vespa/vespa_client.py
papa99do: simplified the logic a bit. removed the to_dict message, just print out the snake_case field name in the log

src/marqo/api/configs.py
farshidz: Makes sense

farshidz (approved):

PR #1366, opened 2026-02-11
Add validation for update documents id fields

wanliAlex:
## Checklist

### For new field types:

---

> [!NOTE]
> **Medium Risk**
> Touches the semi-structured `partial_update_documents` flow and Vespa batch-get error handling; regressions could affect update behavior and error surfacing for some IDs.
> 
> **Overview**
> **Semi-structured partial updates no longer crash on problematic `_id`s.** When map fields force a `get_batch` prefetch, missing docs are now mapped back to the original requested IDs (instead of Vespa-pruned IDs), a response-count mismatch triggers an `InternalError`, and per-document errors include a consistent generic message (plus `message` for backward compatibility).
> 
> **Vespa `get_batch` is hardened for bad IDs.** `_get_document_async` now catches `httpx.InvalidURL` (e.g., non-printable ASCII in IDs) and returns a per-document `400` response, and the API docs/comments clarify that `get_batch` preserves input order.
> 
> Adds unit + integration coverage for preserved ordering, invalid-ID handling, and the previously crashing `#`-in-ID case; bumps version to `2.24.15`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ffc9725879313679a1899a743377324c304842a4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/marqo/core/document/document.py
+++ b/src/marqo/core/document/document.py
@@ -359,5 +359,8 @@ class Document:
                                          processingTimeMs=add_docs_processing_time_ms)
 
     def extract_document_id_from_vespa_id(self, resp):
+        # TODO - This method is not reliable as Vespa might prune the document ID in the response if
+        #  the ID certain characters (e.g., #). We should update our code to either remove those illegal characters
+        #  from the document ID before feeding to Vespa, or implement a more reliable solution
         doc_id = resp.id.split('::')[-1] if resp.id else None
         return doc_id
--- a/src/marqo/core/document/document.py
+++ b/src/marqo/core/document/document.py
@@ -32,10 +32,9 @@ logger = get_logger(__name__)
 
 
 class Document:
-    GENERIC_UPDATE_DOCUMENT_ERROR_MESSAGE = "Marqo vector store couldn't update the document. " + \
-                                  "Please see: " + update_documents_response() + " for more details"
-
     """A class that handles the document API in Marqo"""
+    GENERIC_UPDATE_DOCUMENT_ERROR_MESSAGE = "Marqo vector store couldn't update the document. " + \
+                                            "Please see: " + update_documents_response() + " for more details"
 
     def __init__(self, vespa_client: VespaClient, index_management: IndexManagement, inference: Inference):
         self.vespa_client = vespa_client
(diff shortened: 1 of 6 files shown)


src/marqo/core/document/document.py
```diff
@@ -153,17 +155,42 @@ def partial_update_documents(self, partial_documents: List[Dict], marqo_index) \
...
+            get_batch_response = self.vespa_client.get_batch(ids = fetch_ids, fields = [
                 VESPA_FIELD_ID, INT_FIELDS, FLOAT_FIELDS, VESPA_DOC_FIELD_TYPES, VESPA_DOC_VERSION_UUID], schema = marqo_index.schema_name)
             responses = get_batch_response.responses
-            for resp in responses:
+
+            if len(responses) != len(fetch_ids):
+                raise InternalError(
+                    f"Vespa get batch response count {len(responses)} does not match the number of requested documents "
+                    f"{len(fetch_ids)}. This likely indicates an issue with Vespa or the Vespa client."
+                )
```
papa99do, line 167: InternalError will be mapped to a 500 error. E-comm will retry this batch for 10 times. If this error is not retryable, we need to return 400 instead. Should we just report the missing ids as individual 404?

src/marqo/vespa/vespa_client.py
```diff
@@ -1056,6 +1057,10 @@ async def _get_document_async(self,
                     resp = await async_client.get(
                         f'{self.document_url}/document/v1/{schema}/{schema}/docid/{id}', timeout=timeout
                     )
+            except httpx.InvalidURL as e:
```
papa99do, line 1060: ideally, we should not send invalid URL to vespa. `id` should be url encoded somehow.

src/marqo/core/document/document.py
wanliAlex: In the new implementation, a mismatch should not happen. I think this is retriable.

src/marqo/vespa/vespa_client.py
wanliAlex: Yes, I would say that's a bigger task to add validation to the `_id` field. Should we do it in another ticket?

PR #1391, opened 2026-03-27
Move version to components/common as single source of truth

wanliAlex:
## Summary
- Created `components/common/src/marqo_common/version.py` as the single source of truth for the Marqo version (`2.26.0`)
- Updated `marqo`, `inference_orchestrator`, and `model_management` component `version.py` files to re-export from `marqo_common.version`, preserving existing import compatibility
- Added `marqo-common` as a dependency to `model_management` (the other components already had it)

## Test plan
- [x] Verify each component can still import `__version__` and `get_version()` from their local `version.py`
- [x] Verify changing the version in `marqo_common/version.py` propagates to all components
- [ ] Verify `model_management` installs `marqo-common` dependency correctly

### DIFF
--- a/components/inference_orchestrator/src/inference_orchestrator/version.py
+++ b/components/inference_orchestrator/src/inference_orchestrator/version.py
@@ -1,3 +1,6 @@
-from marqo_common.version import __version__, get_version
+from marqo_common.version import (  # pragma: no cover - tested in components/marqo
+    __version__,
+    get_version,
+)
 
-__all__ = ["__version__", "get_version"]
+__all__ = ["__version__", "get_version"]  # pragma: no cover
--- a/components/inference_orchestrator/src/inference_orchestrator/version.py
+++ b/components/inference_orchestrator/src/inference_orchestrator/version.py
@@ -1,4 +1,3 @@
-__version__ = "2.26.0"
+from marqo_common.version import __version__, get_version
 
-def get_version() -> str:
-    return f"{__version__}"
+__all__ = ["__version__", "get_version"]
--- a/components/inference_orchestrator/src/inference_orchestrator/version.py
+++ b/components/inference_orchestrator/src/inference_orchestrator/version.py
@@ -1,6 +1,3 @@
-from marqo_common.version import (  # pragma: no cover - tested in components/marqo
-    __version__,
-    get_version,
-)
+from marqo_common.version import __version__, get_version  # pragma: no cover
 
 __all__ = ["__version__", "get_version"]  # pragma: no cover
--- a/components/inference_orchestrator/src/inference_orchestrator/version.py
+++ b/components/inference_orchestrator/src/inference_orchestrator/version.py
@@ -1,3 +1,3 @@
-from marqo_common.version import __version__, get_version  # pragma: no cover
+from marqo_common.version import __version__, get_version
 
-__all__ = ["__version__", "get_version"]  # pragma: no cover
+__all__ = ["__version__", "get_version"]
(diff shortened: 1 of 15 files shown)


components/inference_orchestrator/src/inference_orchestrator/version.py
```diff
@@ -1,4 +1,6 @@
-__version__ = "2.26.0"
+from marqo_common.version import (  # pragma: no cover - tested in components/marqo
```
vicilliar: nit: why is this multiline when the other ones are single line?

components/marqo/tests/unit_tests/marqo/test_version.py
```diff
@@ -0,0 +1,21 @@
+import re
+from unittest import TestCase
+
+from marqo_common.version import __version__ as common_version, get_version as common_get_version
+from marqo.version import __version__, get_version
+
+
+class TestVersion(TestCase):
```
vicilliar, line 7: we should add this test for model management and inference orchestrator too

components/model_management/Dockerfile
```diff
@@ -1,8 +1,17 @@
+# =============================================================================
```
vicilliar, line 1: good comment!

components/model_management/Dockerfile
```diff
@@ -1,8 +1,17 @@
...
+# IMPORTANT: Build context must be the 'components/' directory.
+# Run:  docker build -f model_management/Dockerfile -t mmc .   (from components/)
+# =============================================================================
+
+# Stage 0: Validate build context - ensures 'common/' is accessible
+FROM busybox AS context_check
```
vicilliar, line 7: what's the rationale of using busybox for model management?

compose-inference.yaml
```diff
@@ -55,7 +55,9 @@ services:
 
   mmc:
     image: 424082663841.dkr.ecr.us-east-1.amazonaws.com/marqoai/model-management:latest
-    build: ./components/model_management
+    build:
```
vicilliar, line 58: what's the rationale of why model management needs this? I may be missing some context here

vicilliar (changes requested):

components/inference_orchestrator/src/inference_orchestrator/version.py
wanliAlex: It's because inference and model management containers have ruff format check, thus the format is enforced here.

components/marqo/tests/unit_tests/marqo/test_version.py
wanliAlex: Do you think we need to duplicate these tests? They will be exactly the same.

components/model_management/Dockerfile
wanliAlex: It's a check to ensure we have the correct context to build the Docker image. BusyBox is considered a super lightweight Linux base image for these tasks.

compose-inference.yaml
wanliAlex: Model management is also using the version in common

components/inference_orchestrator/src/inference_orchestrator/version.py
wanliAlex: Change all to single line

components/marqo/tests/unit_tests/marqo/test_version.py
wanliAlex: Added tests for all components

components/inference_orchestrator/src/inference_orchestrator/version.py
vicilliar: why doesn't it fail the ruff check now though now that it's single line

components/marqo/tests/unit_tests/marqo/test_version.py
vicilliar: they will be, but they test it on different components

vicilliar (approved):
lgtm

PR #1378, opened 2026-03-10
Add CONTAINS filter for semi-structured indexes

papa99do:
## Summary
- Adds `field CONTAINS value` filter syntax that leverages Vespa's existing `contains` operator on lexical fields for token-level matching
- Supports bare values (`field CONTAINS hello`) and grouped values (`field CONTAINS (multi word)`)
- Case-insensitive keyword (`CONTAINS`, `contains`, `Contains` all work)
- Raises errors for structured and unstructured indexes (semi-structured only for now)
- Fixes AST node mutation side effect in filter string generators (`generate_contains_filter_string`, `generate_equality_filter_string`, `generate_range_filter_string`)

## Test plan
- [x] Unit tests for parser: 13 CONTAINS test cases in `test_search_filter.py` (basic, case variants, grouped value, AND/OR/NOT, combined with equality, parenthesized, escaped space)
- [x] Integration tests: 18 tests in `test_contains_filter.py` covering basic match, case insensitivity, NOT/AND/OR, equality combination, tensor/lexical/hybrid search, structured index error, grouped multi-word value, special characters, and error cases
- [x] All existing unit tests still pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Adds a new filter operator that changes parsing and Vespa query generation for semi-structured searches; incorrect escaping/field validation could impact search correctness and error handling, though coverage is strong.
> 
> **Overview**
> Adds a new `field CONTAINS value` filter operator to the filter-string AST/parser (new `ContainsTerm`, case-insensitive keyword, supports grouped multi-word values) and wires it into semi-structured Vespa query generation by translating it to a lexical-field `contains` clause.
> 
> Also tightens filter-string generation by **avoiding mutation of AST nodes** during escaping, and explicitly rejects `CONTAINS` for structured and legacy unstructured indexes with `InvalidArgumentError`. Extensive unit and integration tests cover parsing, boolean logic composition, special characters, and expected error cases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 37e795721ef3c06bfa0a1883e1a4bd853f5221de. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #1394, opened 2026-03-29
Add CONTAINS filter for semi-structured indexes (2.25.4)

papa99do:
## Summary
- Adds `field CONTAINS value` filter syntax that leverages Vespa's existing `contains` operator on lexical fields for token-level matching
- Supports bare values (`field CONTAINS hello`) and grouped values (`field CONTAINS (multi word)`)
- Case-insensitive keyword (`CONTAINS`, `contains`, `Contains` all work)
- Raises errors for structured and unstructured indexes (semi-structured only for now)
- Fixes AST node mutation side effect in filter string generators
- Bumps version to 2.25.4

Cherry-picked from #1378 onto `releases/2.25`.

## Test plan
- [x] Unit tests for parser: CONTAINS test cases including special chars, escaping, grouped values
- [x] Unit tests for filter generation in all 3 index types (semi-structured, structured error, unstructured error)
- [x] Integration tests covering basic match, case insensitivity, NOT/AND/OR, equality combination, tensor/lexical/hybrid search, structured index error, special characters in filter values
- [x] All existing unit tests still pass

---

> [!NOTE]
> **Medium Risk**
> Introduces a new filter operator and extends the filter parser plus Vespa query generation, which could impact search filtering behavior and escaping edge cases. Scope is limited by explicitly rejecting `CONTAINS` on structured and legacy unstructured indexes and adding broad unit/integ coverage.
> 
> **Overview**
> Adds a new `field CONTAINS value` filter operator that parses into a `ContainsTerm` AST node (case-insensitive keyword, supports grouped multi-word values) and converts it into a Vespa `contains` predicate over semi-structured lexical fields.
> 
> Updates semi-structured filter generation to avoid mutating parsed AST nodes while escaping values, validates `CONTAINS` fields against the index’s lexical `field_map`, and explicitly raises `InvalidArgumentError` for `CONTAINS` on structured and legacy unstructured indexes. Bumps version to `2.25.4` and adds unit + integration tests covering parsing, escaping/special chars, boolean/numeric/equality interactions, and unsupported-index errors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a81177e8a5d2f242d751658bb2d9c05a2ab2ae32. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/marqo/src/marqo/core/search/search_filter.py
+++ b/components/marqo/src/marqo/core/search/search_filter.py
@@ -374,10 +374,9 @@ class MarqoFilterStringParser:
                         read_space_until = ')'
                     elif len(self._get_current_term_value()) == 0 and c == '[' and not read_space_until:  # start of term value
                         read_space_until = ']'
-                        if self._term_type == MarqoFilterStringParser._TermType.In:
-                            self._error('Unexpected [ after IN operator.', filter_string, i)
-                        elif self._term_type == MarqoFilterStringParser._TermType.Contains:
-                            self._error('Unexpected [ after CONTAINS operator.', filter_string, i)
+                        if self._term_type in (MarqoFilterStringParser._TermType.In,
+                                               MarqoFilterStringParser._TermType.Contains):
+                            self._error('[ and ] are only usable with the RANGE operator.', filter_string, i)
                         else:
                             self._term_type = MarqoFilterStringParser._TermType.Range
                     else:
--- a/components/marqo/src/marqo/core/search/search_filter.py
+++ b/components/marqo/src/marqo/core/search/search_filter.py
@@ -374,10 +374,12 @@ class MarqoFilterStringParser:
                         read_space_until = ')'
                     elif len(self._get_current_term_value()) == 0 and c == '[' and not read_space_until:  # start of term value
                         read_space_until = ']'
-                        if self._term_type != MarqoFilterStringParser._TermType.In:
-                            self._term_type = MarqoFilterStringParser._TermType.Range
-                        else:
+                        if self._term_type == MarqoFilterStringParser._TermType.In:
                             self._error('Unexpected [ after IN operator.', filter_string, i)
+                        elif self._term_type == MarqoFilterStringParser._TermType.Contains:
+                            self._error('Unexpected [ after CONTAINS operator.', filter_string, i)
+                        else:
+                            self._term_type = MarqoFilterStringParser._TermType.Range
                     else:
                         self._append_to_term_value(c)
 
--- a/components/marqo/src/marqo/core/search/search_filter.py
+++ b/components/marqo/src/marqo/core/search/search_filter.py
@@ -159,6 +159,23 @@ class InTerm(Term):
         return f'{self.__class__.__name__}({repr(self.field)}, {repr(self.value_list)}, {repr(self.raw)})'
 
 
+class ContainsTerm(Term):
+    def __init__(self, field: str, value: str, raw: str):
+        super().__init__(field, raw)
+        self.value = value
+
+    def __eq__(self, other):
+        return (
+                type(self) == type(other) and
+                self.field == other.field and
+                self.value == other.value and
+                self.raw == other.raw
+        )
+
+    def __repr__(self):
+        return f'{self.__class__.__name__}({repr(self.field)}, {repr(self.value)}, {repr(self.raw)})'
+
+
 class And(Operator):
     def __init__(self, left: Node, right: Node, raw: str = 'AND'):
         super().__init__(left, right, raw)
@@ -228,10 +245,13 @@ class MarqoFilterStringParser:
 
     IN_TERM_DIVIDER = ' IN ('
 
+    CONTAINS_TERM_DIVIDER = ' CONTAINS '
+
     class _TermType(Enum):
         Equality = 1
         Range = 2
         In = 3
+        Contains = 4
 
     def _term_divider_is_IN(self, i: int, filter_string: str) -> bool:
         """
@@ -244,6 +264,14 @@ class MarqoFilterStringParser:
             self._error("Expected ( after ' IN '", filter_string, i)
 
         return candidate_substring == self.IN_TERM_DIVIDER
+
+    def _term_divider_is_CONTAINS(self, i: int, filter_string: str) -> bool:
+        """
+        Given 'i' and the full filter string, determine if 'i' is at the beginning of the CONTAINS term divider.
+        The keyword 'CONTAINS' is case-insensitive.
+        """
+        candidate_substring = filter_string[i:i + len(self.CONTAINS_TERM_DIVIDER)].upper()
+        return candidate_substring == self.CONTAINS_TERM_DIVIDER
 
     def _append_to_term_value(self, c: str):
         """
@@ -405,8 +433,19 @@ class MarqoFilterStringParser:
                 self._current_raw_token.append(c)
                 escape = True
             elif c == ' ':
+                # Found the ' CONTAINS ' operator.
+                if len(self._current_token) > 0 and self._term_divider_is_CONTAINS(i, filter_string):
+                    self._read_term_value = True
+                    self._term_type = MarqoFilterStringParser._TermType.Contains
+                    self._term_field = ''.join(self._current_token)
+                    self._current_token.append(self.CONTAINS_TERM_DIVIDER)
+                    self._current_raw_token.append(self.CONTAINS_TERM_DIVIDER)
+
+                    # Skip past ' CONTAINS ' (but not the last char, the loop increments i)
+                    i += len(self.CONTAINS_TERM_DIVIDER) - 1
+
                 # Found the ' IN ' operator. Look for a list starting with '(' on the next pass.
-                if self._term_divider_is_IN(i, filter_string):
+                elif self._term_divider_is_IN(i, filter_string):
                     self._read_term_value = True
                     self._term_type = MarqoFilterStringParser._TermType.In
                     self._term_value = [[]]
@@ -540,6 +579,8 @@ class MarqoFilterStringParser:
             elif term_type == self._TermType.In:
                 # For IN terms, term_value should already be a list, not a string
                 node = InTerm(term_field, term_value, raw_token)
+            elif term_type == self._TermType.Contains:
+                node = ContainsTerm(term_field, term_value, raw_token)
             else:
                 raise InternalError(f'Unexpected term type {term_type}')
 
--- a/components/marqo/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
+++ b/components/marqo/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
@@ -519,6 +519,20 @@ class SemiStructuredVespaIndex(StructuredVespaIndex, UnstructuredVespaIndex):
 
             return f'({float_field_string} OR {int_field_string})'
 
+        def generate_contains_filter_string(node: search_filter.ContainsTerm) -> str:
+            node.field = self.escape(node.field)
+            marqo_index = self.get_marqo_index()
+            if node.field not in marqo_index.field_map:
+                raise InvalidArgumentError(
+                    f"CONTAINS filter requires a lexical field, but '{node.field}' "
+                    f"is not a lexical field in index '{marqo_index.name}'. "
+                    f"Available lexical fields: {', '.join(sorted(marqo_index.lexically_searchable_fields_names))}"
+                )
+            field = marqo_index.field_map[node.field]
+            lexical_field_name = field.lexical_field_name
+            escaped_value = self.escape(node.value)
+            return f'({lexical_field_name} contains "{escaped_value}")'
+
         def tree_to_filter_string(node: search_filter.Node) -> Optional[str]:
             # Skip any terms with excluded fields first - check at node level
             if (isinstance(node, search_filter.Term) or isinstance(node, search_filter.Modifier)) and exclude_terms is not None:
@@ -565,6 +579,8 @@ class SemiStructuredVespaIndex(StructuredVespaIndex, UnstructuredVespaIndex):
                     return generate_range_filter_string(node)
                 elif isinstance(node, search_filter.InTerm):
                     raise InvalidArgumentError("The 'IN' filter keyword is not yet supported for unstructured indexes")
+                elif isinstance(node, search_filter.ContainsTerm):
+                    return generate_contains_filter_string(node)
 
             raise InternalError(f'Unknown node type {type(node)}')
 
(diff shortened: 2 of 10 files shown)


components/marqo/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
```diff
@@ -444,53 +444,53 @@ def generate_equality_filter_string(node: search_filter.EqualityTerm) -> str:
             filter_parts = []
 
             # Escape special characters in field name and value
-            node.field = self.escape(node.field)
-            node.value = self.escape(node.value)
+            escaped_field = self.escape(node.field)
+            escaped_value = self.escape(node.value)
```
papa99do, line 449: to eliminate side-effect.

vicilliar:
Just a general comment, maybe I just don't have context: Why do we need a `contains` filter if we already use vespa `contains` for our actual `OR` and `AND` terms? Can't we just make something a required term in the lexical query to achieve this effect?

Is this just so the `contains` restriction applies to the tensor results as well?

components/marqo/src/marqo/core/search/search_filter.py
```diff
@@ -346,10 +374,12 @@ def parse(self, filter_string: str) -> SearchFilter:
                         read_space_until = ')'
                     elif len(self._get_current_term_value()) == 0 and c == '[' and not read_space_until:  # start of term value
                         read_space_until = ']'
-                        if self._term_type != MarqoFilterStringParser._TermType.In:
-                            self._term_type = MarqoFilterStringParser._TermType.Range
-                        else:
+                        if self._term_type == MarqoFilterStringParser._TermType.In:
                             self._error('Unexpected [ after IN operator.', filter_string, i)
+                        elif self._term_type == MarqoFilterStringParser._TermType.Contains:
+                            self._error('Unexpected [ after CONTAINS operator.', filter_string, i)
```
vicilliar: nit: A more helpful error message for this now would be: `[ and ] are only usable with the RANGE operator`.

components/marqo/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
vicilliar: makes sense

components/marqo/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
```diff
@@ -512,13 +512,27 @@ def generate_range_filter_string(node: search_filter.RangeTerm) -> str:
...
             int_field_string = (f'({INT_FIELDS} contains '
-                                f'sameElement(key contains "{node.field}", {bound}))')
+                                f'sameElement(key contains "{escaped_field}", {bound}))')
 
             return f'({float_field_string} OR {int_field_string})'
 
+        def generate_contains_filter_string(node: search_filter.ContainsTerm) -> str:
+            escaped_field = self.escape(node.field)
+            marqo_index = self.get_marqo_index()
+            if escaped_field not in marqo_index.field_map:
```
vicilliar, line 527: this check is actually only for if the field exists. not just that it is not a lexical field.

components/marqo/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
```diff
@@ -512,13 +512,27 @@ def generate_range_filter_string(node: search_filter.RangeTerm) -> str:
...
+                raise InvalidArgumentError(
+                    f"CONTAINS filter requires a lexical field, but '{escaped_field}' "
+                    f"is not a lexical field in index '{marqo_index.name}'. "
+                    f"Available lexical fields: {', '.join(sorted(marqo_index.lexically_searchable_fields_names))}"
+                )
+            field = marqo_index.field_map[escaped_field]
+            lexical_field_name = field.lexical_field_name
+            escaped_value = self.escape(node.value)
```
vicilliar: nit: put right under the `escaped_field` assignment, for uniformity

components/marqo/src/marqo/core/unstructured_vespa_index/unstructured_vespa_index.py
```diff
@@ -228,6 +228,8 @@ def tree_to_filter_string(node: search_filter.Node) -> str:
                     return generate_range_filter_string(node)
                 elif isinstance(node, search_filter.InTerm):
                     raise InvalidArgumentError("The 'IN' filter keyword is not yet supported for unstructured indexes")
+                elif isinstance(node, search_filter.ContainsTerm):
+                    raise InvalidArgumentError("The 'CONTAINS' filter keyword is not yet supported for unstructured indexes")
```
vicilliar: maybe "legacy unstructured" would make more sense, or put "unstructured indexes created before Marqo 2.13.0"

components/marqo/tests/integ_tests/tensor_search/search/test_contains_filter.py
```diff
@@ -0,0 +1,215 @@
...
+                    {"_id": "5", "title": "Machine Learning", "description": "AI and ML concepts"},
+                    {"_id": "6", "title": "Real-time Systems", "description": "Low-latency computing"},
+                    {"_id": "7", "title": "vintage t-shirt collection", "description": "fashion items"},
+                ],
+                tensor_fields=["title", "description"]
+            )
+        )
+
+    def _search(self, filter_str, text="", search_method=SearchMethod.TENSOR):
+        return tensor_search.search(
```
vicilliar, line 68: It's worth testing that the base case (`_search` without any filter) returns all 7 docs. So we know for sure the filter is doing the work

components/marqo/tests/integ_tests/tensor_search/search/test_contains_filter.py
```diff
@@ -0,0 +1,215 @@
...
+        self.assertEqual(self._get_ids(res), ["1", "2"])
+
+    def test_no_match(self):
+        """title CONTAINS nonexistent should return 0 hits."""
+        res = self._search("title CONTAINS nonexistent")
+        self.assertEqual(len(res["hits"]), 0)
+
+    def test_not_contains(self):
+        """NOT (title CONTAINS hello) should match docs 2, 3, 5."""
+        res = self._search("NOT (title CONTAINS hello)", text="world programming learning")
```
vicilliar: can you test it also works without the parenthesis? `NOT title contains hello`

components/marqo/tests/integ_tests/tensor_search/search/test_contains_filter.py
```diff
@@ -0,0 +1,215 @@
...
+        ids = self._get_ids(res)
+        # All results should NOT have "hello" in title
+        for doc_id in ids:
+            self.assertNotIn(doc_id, ["1", "4"])
+        # Should include at least some of docs 2, 3, 5
```
vicilliar: We can have a more deterministic assertion. 
`self.assertEqual(self._get_ids(res), ["2", "3", "5", "6", "7"])`
Feels like Claude relaxed the test requirements here

components/marqo/tests/integ_tests/tensor_search/search/test_contains_filter.py
vicilliar: Also why do we need `text` param here

components/marqo/tests/integ_tests/tensor_search/search/test_contains_filter.py
```diff
@@ -0,0 +1,215 @@
...
+        self.assertTrue(len(ids) > 0)
+
+    def test_contains_with_and(self):
+        """title CONTAINS hello AND description CONTAINS greeting should match docs 1 and 4."""
+        res = self._search("title CONTAINS hello AND description CONTAINS greeting")
```
vicilliar: This doesn't actually test `AND`. Because both docs 1 and 4 have both conditions met. We should split them up, such that docs 1 and 4 have title CONTAINS hello, but maybe docs 4 and 5 have description CONTAINS greeting.

components/marqo/tests/integ_tests/tensor_search/search/test_contains_filter.py
```diff
@@ -0,0 +1,215 @@
...
+        self.assertEqual(self._get_ids(res), ["1", "4"])
+
+    def test_contains_with_or(self):
+        """title CONTAINS python OR description CONTAINS greeting should match docs 1, 3, 4."""
+        res = self._search("title CONTAINS python OR description CONTAINS greeting")
+        self.assertEqual(self._get_ids(res), ["1", "3", "4"])
+
+    def test_contains_combined_with_equality(self):
```
vicilliar, line 135: add a test combining it with `range` filter

components/marqo/tests/integ_tests/tensor_search/search/test_contains_filter.py
```diff
@@ -0,0 +1,215 @@
...
+            ids = self._get_ids(res)
+            self.assertIn("6", ids)
+
+        with self.subTest("grouped value with special chars: title CONTAINS (t-shirt collection)"):
+            res = self._search("title CONTAINS (t-shirt collection)")
+            ids = self._get_ids(res)
+            self.assertIn("7", ids)
+
+    def test_structured_index_contains_raises_error(self):
+        """Using CONTAINS on a structured index should raise InvalidArgumentError."""
```
vicilliar, line 254: test with legacy unstructured

components/marqo/tests/integ_tests/tensor_search/search/test_contains_filter.py
```diff
@@ -0,0 +1,215 @@
...
+            # Doc 7 has "vintage t-shirt collection". Vespa tokenizes "t-shirt" into
+            # tokens. The contains operator should match because the indexed text
+            # was tokenized the same way.
+            res = self._search("title CONTAINS t-shirt")
+            ids = self._get_ids(res)
+            self.assertIn("7", ids)
+
+        with self.subTest("hyphen in filter value: title CONTAINS real-time"):
+            # Doc 6 has "Real-time Systems"
+            res = self._search("title CONTAINS real-time")
```
vicilliar, line 218: test with `:` in filter value, `"` in filter value

components/marqo/tests/integ_tests/tensor_search/search/test_contains_filter.py
```diff
@@ -0,0 +1,215 @@
...
+            text=text,
+            filter=filter_str,
+            search_method=search_method,
+        )
+
+    def _get_ids(self, res):
+        return sorted([hit["_id"] for hit in res["hits"]])
+
+    def test_basic_contains_match(self):
+        """title CONTAINS hello should match docs 1 and 4 (case-insensitive after tokenization)."""
```
vicilliar, line 92: What about something like `title CONTAINS ello`, would that also work?

components/marqo/tests/unit_tests/marqo/core/semi_structured_vespa_index/test_semi_structured_vespa_index.py
```diff
@@ -87,6 +88,55 @@ def test_get_filter_string_collapse_field(self, mock_get_marqo_index):
...
+        """Test equality filter paths: _id, bool, string array, and float numeric."""
+        test_cases = [
+            # _id filter
+            ('_id:doc123', 'marqo__id contains "doc123"'),
+            # Bool filter
+            ('title:true', 'marqo__bool_fields'),
+            # String array filter
+            ('tags:foo', 'marqo__string_array_tags contains "foo"'),
+            # Float numeric filter
+            ('title:3.14', 'marqo__float_fields'),
```
vicilliar: this is missing the 2nd part of the phrase

components/marqo/tests/unit_tests/marqo/core/semi_structured_vespa_index/test_semi_structured_vespa_index.py
```diff
@@ -87,6 +88,55 @@ def test_get_filter_string_collapse_field(self, mock_get_marqo_index):
                 result_filter_string = self.vespa_index._get_filter_term(marqo_query)
                 self.assertEqual(expected_result, result_filter_string)
 
+    def test_get_filter_string_equality_paths(self):
+        """Test equality filter paths: _id, bool, string array, and float numeric."""
+        test_cases = [
+            # _id filter
+            ('_id:doc123', 'marqo__id contains "doc123"'),
+            # Bool filter
+            ('title:true', 'marqo__bool_fields'),
```
vicilliar: this is missing the 2nd part of the phrase

vicilliar (changes requested):

components/marqo/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
papa99do: yes, for semistructured_marqo_index, this field_map only contains lexical fields. I'll add a comment here.

components/marqo/tests/integ_tests/tensor_search/search/test_contains_filter.py
papa99do: No, it only does token match, not substring. 
Some context in this thread: https://marqo-ai.slack.com/archives/C09J2JWP21F/p1774306534117569
and in this original one: https://marqo-ai.slack.com/archives/C09J2JWP21F/p1772663830300769

components/marqo/tests/integ_tests/tensor_search/search/test_contains_filter.py
papa99do: no we don't need text.

components/marqo/tests/integ_tests/tensor_search/search/test_contains_filter.py
papa99do: good point, will change

components/marqo/tests/unit_tests/marqo/core/semi_structured_vespa_index/test_semi_structured_vespa_index.py
papa99do: you are correct. I'll add them. But I was wondering why claude decided to test equality path here, we didn't change equality path logic.

components/marqo/tests/unit_tests/marqo/core/semi_structured_vespa_index/test_semi_structured_vespa_index.py
papa99do: same as above.

JiwaniZakir (commented):
The case-sensitivity behavior between `_term_divider_is_CONTAINS` and `_term_divider_is_IN` is inconsistent: `_term_divider_is_CONTAINS` explicitly uppercases the candidate substring before comparing, making `CONTAINS` case-insensitive, while `_term_divider_is_IN` performs a direct string comparison against `' IN ('` with no case normalization. This should either be documented as intentional or made consistent — a user writing `field contains value` would succeed while `field in (value)` would fail.

The fix in `generate_equality_filter_string` (using `escaped_field`/`escaped_value` local variables instead of mutating `node.field` and `node.value` directly) is a meaningful correctness fix — the original code would double-escape if the same node was ever visited more than once, which could happen with filter caching or tree traversal patterns.

In `_push_token`, the `ContainsTerm` branch passes `term_value` directly without asserting it's a plain string. For `InTerm`, the list structure is enforced by the parsing logic, but for `ContainsTerm` there's no guard — if a bug elsewhere caused `term_value` to be a list, `ContainsTerm.__init__` would accept it silently and produce unexpected behavior downstream when the value is used in a Vespa filter expression.

vicilliar (approved):
lgtm

PR #1395, opened 2026-03-30
Remove NLTK in the API and upgrade NLTK in Inference

wanliAlex:
## Change Summary

- Remove NLTK in the API container
- Upgrade NLTK in the inference container

## Checklist

### For new field types:


components/inference_orchestrator/pyproject.toml
```diff
@@ -13,7 +13,7 @@ dependencies = [
     "more-itertools>=10.8.0",
     "msgpack>=1.1.1",
     "msgpack-numpy>=0.4.8",
-    "nltk>=3.9.1",
+    "nltk>=3.9.2",
```
papa99do: nltk>=3.9.2 — 🚨 Action required
Your floor of 3.9.2 is vulnerable. Two CVEs have been disclosed in the last ~6 weeks:
CVE-2025-14009 (Critical, CVSS 10.0) — published February 2026
Affects all versions up to and including 3.9.2. The _unzip_iter function in nltk/downloader.py uses zipfile.extractall() without path validation, allowing a malicious zip package to write files anywhere on the filesystem. If Python files like __init__.py are placed in an importable location, they execute automatically on the next import — leading to full RCE. Fixed in 3.9.3. GitHub
CVE-2026-33236 (High) — published March 2026
The NLTK downloader doesn't validate subdir and id attributes when processing remote XML index files, enabling an attacker to perform a path traversal (arbitrary file overwrite). Affects all versions up to 3.9.2. No fix available yet as of the advisory. GitLab
Action: Bump to nltk>=3.9.3 immediately — that closes CVE-2025-14009. CVE-2026-33236 is still unpatched upstream, so the mitigations are:

Don't use nltk.download() with untrusted or custom server URLs
If you must use the downloader, point only at the official NLTK data server over HTTPS and in a controlled environment

components/marqo/pyproject.toml
```diff
@@ -27,6 +26,10 @@ dependencies = [
     "validators>=0.35.0",
     "marqo-common",
     "pydantic-settings>=2.12.0",
+    "requests>=2.32.5",
```
papa99do: requests>=2.32.5 — ⚠️ Minor concern
A new CVE (CVE-2026-25645) was fixed in requests 2.33.0: requests.utils.extract_zipped_paths now extracts to a non-deterministic location to prevent malicious file replacement. Importantly, this does not affect default usage — only applications calling that utility function directly.
The latest non-vulnerable version is 2.33.0. If you're not using extract_zipped_paths directly, your current pin is functionally fine, but bumping to >=2.33.0 would be the clean fix.
Also from 2.32.5: CVE-2024-47081 was addressed, fixing an issue where a maliciously crafted URL in a trusted environment could retrieve credentials for the wrong hostname from a netrc file. Your floor of 2.32.5 already covers that.

papa99do (changes requested):

JiwaniZakir (commented):
The torchvision wheel URLs in `components/inference_orchestrator/uv.lock` have been silently changed from `download.pytorch.org` to `download-r2.pytorch.org` — this appears unrelated to the NLTK changes and should be called out explicitly. It's worth confirming this is an intentional mirror switch and that `download-r2.pytorch.org` is a stable, supported endpoint rather than a transient URL that could break future installs.

Additionally, `components/marqo/pyproject.toml` now adds `requests`, `jinja2`, `opentelemetry-api`, and `opentelemetry-sdk` as explicit dependencies, which have no obvious connection to removing NLTK. If these were previously pulled in as transitive dependencies of NLTK/joblib, a comment explaining why they're now being pinned explicitly would help reviewers and future maintainers understand the intent. If they were already transitive deps from other packages, it would be worth verifying they don't get dropped when those other packages are updated.

components/marqo/pyproject.toml
wanliAlex: Done

components/inference_orchestrator/pyproject.toml
wanliAlex: Done. bump to 3.9.4

papa99do (approved):

PR #1396, opened 2026-03-30
Fix compatibility test workflow missing marqo-common dependency

wanliAlex:
## Summary
- The move-version-to-common PR (#1391) made `marqo.version` import from `marqo_common.version`, but the compatibility test orchestrator workflow only installed `semver`, not `marqo-common`
- This caused `determine_to_version.py` to fail with `ModuleNotFoundError: No module named 'marqo_common'` ([failed run](https://github.com/marqo-ai/marqo/actions/runs/23727878334))
- Fix: install `../common` (the `marqo-common` package) alongside `semver` in the workflow

## Test plan
- [ ] Verify the compatibility test orchestrator workflow passes on this branch

### DIFF
--- a/.github/workflows/backwards_compatibility_marqo_execution.yml
+++ b/.github/workflows/backwards_compatibility_marqo_execution.yml
@@ -82,11 +82,11 @@ jobs:
           repository: ${{ github.repository }} #Check out the repository that contains this action since the tests exist in the same repository
           fetch-depth: 0
 
-      # Step to set up Python 3.9
-      - name: Set up Python 3.9
-        uses: actions/setup-python@v5
+      # Step to set up Python 3.11
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v6
         with:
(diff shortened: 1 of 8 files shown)


.github/workflows/backwards_compatibility_marqo_execution.yml
```diff
@@ -82,11 +82,11 @@ jobs:
           repository: ${{ github.repository }} #Check out the repository that contains this action since the tests exist in the same repository
           fetch-depth: 0
 
-      # Step to set up Python 3.9
-      - name: Set up Python 3.9
-        uses: actions/setup-python@v5
+      # Step to set up Python 3.11
+      - name: Set up Python 3.11
```
vicilliar, line 86: have we upgraded Python version for all other pipelines as well?

wanliAlex: I have upgraded the versions for other in-use workflows

PR #1398, opened 2026-03-31
Fix YQL creation for mixed required and optional terms

OwenElliottDev:
## Change Summary
This PR introduces a change to the way LEXICAL search behaves when you have a mix of quoted and unquoted terms.

Previously the YQL that was generated effectively made at least one "optional" term required. This PR changes the YQL construction for mixed optional and required terms to use Vespa's `rank()` operator, this means that it only requires the required terms while the optional terms still contribute to the BM25 scoring.

This is a change to the behaviour of lexical queries which slightly relaxes the criteria for a hit to appear in the case of mixed usage of quoted and unquoted terms. This is a bug fix though as the current behaviour is broken for queries such as `"dresses" -` which results in no hits because the `-` character is removed on ingestion but the YQL it generates `weakAnd(default contains "-")) AND (default contains "dresses")` makes it required. The new YQL for this example is `rank(default contains "dresses", weakAnd(default contains "-"))`.

## Related Jira Ticket
N/A

## Checklist

- [x] Documentation has been updated (**N/A**)
- [x] Python client changes linked or **N/A**

### For new field types:


components/marqo/tests/integ_tests/tensor_search/search/test_search_combined.py
wanliAlex, line 1: Can we add an integration test that:
1. Create an index;
2. Index some documents with proper lexical fields;
3. Assert on the returned documents for quoted queries

wanliAlex (changes requested):

vicilliar:
We need tests for the following interactions (since these features also use `rank()`):
1) bm25 custom score rerankers (global score modifiers with `marqo__score_` + query with optional & required terms
2) tensor/lexical hybrid search + query with optional & required terms
3) lexical/tensor hybrid search + query with optional & required terms

OwenElliottDev:
@vicilliar I added some integ tests that include hybrid and the custom BM25 score rerankers, can you double check it has the sort of coverage you were expecting?

components/marqo/tests/integ_tests/tensor_search/search/test_lexical_search_optional_terms.py
```diff
@@ -0,0 +1,219 @@
...
+        res = self._search('"Xylophone99"')
+        self.assertEqual(set(), self._result_ids(res))
+
+    def test_punctuation_in_doc_stripped_on_ingestion(self):
+        """A standalone '-' in document text like 'table - heavy' is stripped by Vespa's
+        tokenizer during ingestion, so it cannot be matched as a search term."""
+        res = self._search('"garden" "table"')
+        self.assertEqual({'doc5'}, self._result_ids(res))
+
+    def test_doc_with_punctuation_found_despite_optional_punctuation_in_query(self):
```
vicilliar, line 109: this test doesn't really do much, I guess it's harmless to keep

components/marqo/tests/integ_tests/tensor_search/search/test_lexical_search_optional_terms.py
```diff
@@ -0,0 +1,219 @@
...
+            ),
+            score_modifiers=ScoreModifierLists(
+                add_to_score=[
+                    {"field_name": f"{MARQO_CUSTOM_SCORE_RERANK_INPUT_PREFIX}bm25_field_title", "weight": 1.0}
+                ]
+            ),
+            result_count=10
+        )
+        result_ids = self._result_ids(res)
+        self.assertIn('doc1', result_ids)
```
vicilliar, line 181: This technically doesn't test that custom score rerankers are working, since it doesn't check scores. check patterns in `test_custom_score_reranking_feature` for how we do test this

vicilliar: For these tests, you should have a doc that has `fox` but NOT `Quick` so you know for sure that an optional term alone isn't sufficient for matching.

components/marqo/tests/integ_tests/tensor_search/search/test_lexical_search_optional_terms.py
```diff
@@ -0,0 +1,219 @@
...
+            hybrid_parameters=HybridParameters(
+                retrievalMethod=RetrievalMethod.Lexical,
+                rankingMethod=RankingMethod.Tensor,
+            ),
+            result_count=10
+        )
+        result_ids = self._result_ids(res)
+        self.assertIn('doc1', result_ids)
+        self.assertIn('doc2', result_ids)
+        self.assertNotIn('doc3', result_ids)
```
vicilliar, line 201: what's the point of this `assertNotIn`? Also same thing, this doesn't actually test that lexical retrieval with tensor ranking is working. 

A good thing would be to capture the YQL for these cases and assert on them, as a sanity check. Same way we do in `test_search_combined`

vicilliar (changes requested):

components/marqo/src/marqo/core/vespa_index/vespa_index.py
```diff
@@ -300,6 +300,20 @@ def _get_rerank_depth_and_additional_hits_from_query(query: Union[MarqoTensorQue
...
 
+    @staticmethod
+    def _combine_lexical_or_and_terms(or_terms: str, and_terms: str) -> str:
+        """Combine optional (OR) and required (AND) lexical terms using Vespa rank().
+
+        When both are present, rank() makes and_terms the recall set and or_terms
+        scoring-only, so optional terms never filter out documents.
+        """
+        if or_terms and and_terms:
+            return f'rank({and_terms}, {or_terms})'
```
vicilliar, line 311: **Interaction with custom score rerankers:**
- Creates nested rank(): `rank(rank(and_terms, or_terms), extra_bm25_term)`
- Which expands to: `rank(rank(and_terms, or_terms), rank(and_terms_for_ranking, or_terms_for_ranking))`
- This is a bit confusing, it should technically be flattened by vespa but we can flatten it ourselves into `rank(and_terms, or terms, extra_bm25_term)`
- We could change `_get_lexical_search_term` such that it doesn't use `rank()` when `is_ranking_term=True`, because it's useless (the ranking term will always be a non-first term in the outer `rank()` thus we don't need another `rank()` in it.

**Interaction with tensor/lexical and lexical/tensor hybrid search:**
- Also creates nested rank: (rank(tensor_term, rank(and_terms, or_terms)))

Maybe we could have a quick chat?

vicilliar, line 311: `weakAnd()` in or_terms might also not be necessary once it's put into the `rank()` statement. Since we can just put each term as a separate argument, and they're all used for ranking anyway. Worth checking if there are any performance implications of using/not using it. Any thoughts @wanliAlex ?

components/marqo/tests/integ_tests/tensor_search/search/test_lexical_search_optional_terms.py
```diff
@@ -0,0 +1,219 @@
...
+    def tearDown(self):
+        super().tearDown()
+        self.device_patcher.stop()
+
+    def _result_ids(self, res) -> set:
+        return {hit['_id'] for hit in res['hits']}
+
+    def test_custom_score_rerank_bm25_with_required_and_optional_terms(self):
+        """Custom score reranker (marqo__score_bm25_field_*) with required + optional terms.
+        Creates nested rank(): rank(rank(and_terms, or_terms), extra_bm25_term)."""
```
vicilliar, line 163: Note this nested rank() is flatted by `Vespa's QueryCanonicalizer.collapseLevels`. See earlier comment on possibly flattening it ourselves

vicilliar (changes requested):

vicilliar:
Covered in https://github.com/marqo-ai/marqo/pull/1415

PR #1400, opened 2026-03-31
Fix documentation URLs for restructured docs site

wanliAlex:
## Summary
- Remove versioned URL paths (`/{major}.{minor}/`) from `marqo_docs.py` — the docs site no longer uses version segments
- Remove unused `semver` and `marqo.version` imports
- Update URLs to match the new docs site structure (e.g., `reference/api/search/search/` → `reference/api/search/`)
- Point 7 functions whose pages no longer exist to `reference/` with TODO comments for when dedicated pages are available

## Test plan
- [ ] CI `test_documentation` workflow passes (all URLs return 200, no 404s or redirects)
- [ ] Spot-check a few URLs manually to confirm they resolve correctly


components/marqo/src/marqo/marqo_docs.py
```diff
@@ -1,32 +1,25 @@
...
-parsed_version = semver.VersionInfo.parse(version, optional_minor_and_patch=True)
-docs_version = f'{parsed_version.major}.{parsed_version.minor}'
-
 
 def _build_url(path):
-    return f'{base_url}/{docs_version}/{path}'
+    return f'{base_url}/{path}'
 
 
+# TODO: Update URL when a dedicated configuration page is available in the new docs
```
papa99do, line 8: we need to decide whether this is still useful going forward.

papa99do (approved):

PR #1405, opened 2026-04-01
Add opt-in lexicalOperand parameter for hybrid search

wanliAlex:
## Summary
- Adds `lexicalOperand` parameter (`or`, `and`, `weakAnd`) to `HybridParameters` for explicit control over the lexical query operator in semi-structured indexes
- Adds the same parameter to `RelevanceCutoffModel` to override the operand specifically for the relevance cutoff probe query
- Updates Java `HybridSearcher` to support `marqo__yql.lexical.probe` for probe YQL override
- All features are opt-in: default `None` preserves existing automatic operator selection behavior

## Test plan
- [ ] Run unit tests for `TestLexicalOperand` in `test_api_models.py`
- [ ] Run unit tests for `lexical_operand` in `test_relevance_cutoff_model.py`
- [ ] Run `TestLexicalOperandSemiStructured` in `test_semi_structured_vespa_index_to_vespa_query.py`
- [ ] Verify existing tests pass unchanged (no behavior change when `lexicalOperand=None`)

### DIFF
--- a/components/marqo/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
+++ b/components/marqo/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
@@ -237,6 +237,20 @@ class SemiStructuredVespaIndex(StructuredVespaIndex, UnstructuredVespaIndex):
             query["marqo__hybrid.relevanceCutoff.probeDepth"] = marqo_query.relevance_cutoff.probe_depth
             query["marqo__hybrid.relevanceCutoff.affectFacets"] = marqo_query.relevance_cutoff.affect_facets
             query["marqo__hybrid.relevanceCutoff.overrideSortCandidates"] = marqo_query.relevance_cutoff.override_sort_candidates_with_relevant_candidates
+
+            # If relevanceCutoff.lexicalOperand is set, build a separate probe YQL with the overridden operand
+            if marqo_query.relevance_cutoff.lexical_operand is not None:
+                probe_lexical_term = self._get_lexical_search_term(
+                    marqo_query, lexical_operand_override=marqo_query.relevance_cutoff.lexical_operand
+                )
+
+                select_attributes = self._get_select_attributes(marqo_query)
+                filter_term = self._get_filter_term(marqo_query)
+                filter_term = f' AND ({filter_term})' if filter_term else ''
+                query["marqo__yql.lexical.probe"] = (
+                    f'select {select_attributes} from {self._marqo_index.schema_name} '
+                    f'where ({probe_lexical_term}){filter_term}'
+                )
         # Sort by part
         if marqo_query.sort_by:
             query["marqo__hybrid.sortBy.fields"] = [field.dict() for field in marqo_query.sort_by.fields]
--- a/components/marqo/src/marqo/core/structured_vespa_index/structured_vespa_index.py
+++ b/components/marqo/src/marqo/core/structured_vespa_index/structured_vespa_index.py
@@ -1,10 +1,10 @@
-from typing import Union
+from typing import Optional, Union
 
 import marqo.core.search.search_filter as search_filter
 from marqo.core.exceptions import (InvalidDataTypeError, InvalidFieldNameError, VespaDocumentParsingError,
                                    InvalidDataRangeError, MarqoDocumentParsingError, UnsupportedFeatureError)
 from marqo.core.models import MarqoQuery
-from marqo.core.models.hybrid_parameters import RankingMethod, RetrievalMethod
+from marqo.core.models.hybrid_parameters import LexicalOperand, RankingMethod, RetrievalMethod
 from marqo.core.models.marqo_index import *
 from marqo.core.models.marqo_query import MarqoTensorQuery, MarqoLexicalQuery, MarqoHybridQuery
 from marqo.core.structured_vespa_index import common
@@ -844,8 +844,8 @@ class StructuredVespaIndex(VespaIndex):
         else:
             return '*'
 
-    def _generate_or_terms(self, marqo_query: Union[MarqoLexicalQuery, MarqoHybridQuery], is_facets_term=False) \
-            -> str:
+    def _generate_or_terms(self, marqo_query: Union[MarqoLexicalQuery, MarqoHybridQuery], is_facets_term=False,
+                           lexical_operand_override: Optional[str] = None) -> str:
         """Generate the OR/weakAnd terms for the lexical search term.
         Logic flows:
         1. If no or_phrases, return empty string
@@ -875,6 +875,13 @@ class StructuredVespaIndex(VespaIndex):
         if is_facets_term:
             return ' OR '.join(terms)
 
+        # Explicit lexicalOperand overrides automatic logic
+        lexical_operand = lexical_operand_override or (
+            marqo_query.hybrid_parameters.lexicalOperand if isinstance(marqo_query, MarqoHybridQuery) else None
+        )
+        if lexical_operand is not None:
+            return self._apply_lexical_operand(lexical_operand, terms, rerank_depth_lexical)
+
         # Has rerank depth: weakAnd with targetHits
         if rerank_depth_lexical is not None:
             if rerank_depth_lexical <= 0:  # pragma: no cover
@@ -888,15 +895,30 @@ class StructuredVespaIndex(VespaIndex):
         # Default: plain weakAnd
         return f'weakAnd({", ".join(terms)})'
 
-    def _get_lexical_search_term(self, marqo_query: Union[MarqoLexicalQuery, MarqoHybridQuery], is_facets_term=False) \
-            -> str:
+    @staticmethod
+    def _apply_lexical_operand(lexical_operand: LexicalOperand, terms: list, rerank_depth_lexical: Optional[int] = None) -> str:
+        """Apply an explicit lexical operand to combine terms."""
+        if lexical_operand == LexicalOperand.Or:
+            return ' OR '.join(terms)
+        elif lexical_operand == LexicalOperand.And:
+            return ' AND '.join(terms)
+        elif lexical_operand == LexicalOperand.WeakAnd:
+            if rerank_depth_lexical is not None:
+                return f'{{targetHits:{rerank_depth_lexical}}}weakAnd({", ".join(terms)})'
+            return f'weakAnd({", ".join(terms)})'
+        else:
+            raise InternalError(f'Unknown lexical operand: {lexical_operand}')
+
+    def _get_lexical_search_term(self, marqo_query: Union[MarqoLexicalQuery, MarqoHybridQuery], is_facets_term=False,
+                                lexical_operand_override: Optional[str] = None) -> str:
         # Empty query and wildcard
         if not marqo_query.or_phrases and not marqo_query.and_phrases:
             return 'false'
         if marqo_query.or_phrases == ["*"] and not marqo_query.and_phrases:
             return 'true'
 
-        or_terms = self._generate_or_terms(marqo_query, is_facets_term=is_facets_term)
+        or_terms = self._generate_or_terms(marqo_query, is_facets_term=is_facets_term,
+                                           lexical_operand_override=lexical_operand_override)
 
         # Required tokens
         if marqo_query.and_phrases:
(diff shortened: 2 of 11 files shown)


components/marqo/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
```diff
@@ -237,6 +237,21 @@ def _add_relevance_cutoff_and_sort_by_params(self, marqo_query, query):
             query["marqo__hybrid.relevanceCutoff.probeDepth"] = marqo_query.relevance_cutoff.probe_depth
             query["marqo__hybrid.relevanceCutoff.affectFacets"] = marqo_query.relevance_cutoff.affect_facets
             query["marqo__hybrid.relevanceCutoff.overrideSortCandidates"] = marqo_query.relevance_cutoff.override_sort_candidates_with_relevant_candidates
+
+            # If relevanceCutoff.lexicalOperand is set, build a separate probe YQL with the overridden operand
+            if marqo_query.relevance_cutoff.lexical_operand is not None:
+                original_operand = marqo_query.lexical_operand
+                marqo_query.lexical_operand = marqo_query.relevance_cutoff.lexical_operand
+                probe_lexical_term = self._get_lexical_search_term(marqo_query)
+                marqo_query.lexical_operand = original_operand
```
pandu-k: this is pretty messy - we're setting marqo_query.lexical_operand to something then doing this mutation of some sort, then getting probe_lexical_term then setting marqo_query.lexical_operand back

there's got to be a better way than this. can we try to avoid mutating objects as much as possible?

pandu-k (changes requested):

components/marqo/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
papa99do: yes, I agree. an easy way is to accept an `operand_override` param in `_get_lexical_search_term` method. by default, it's None, and we use lexical_operand set in marqoQuery, if provided, use the operand_override to construct the search term.

components/marqo/src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -875,6 +875,11 @@ def _generate_or_terms(self, marqo_query: Union[MarqoLexicalQuery, MarqoHybridQu
         if is_facets_term:
             return ' OR '.join(terms)
 
+        # Explicit lexicalOperand overrides automatic logic
+        lexical_operand = getattr(marqo_query, 'lexical_operand', None)
```
papa99do: why getattr? marqo_query (Union[MarqoLexicalQuery, MarqoHybridQuery]) should always have lexical_operand field, right?

components/marqo/src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -875,6 +875,11 @@ def _generate_or_terms(self, marqo_query: Union[MarqoLexicalQuery, MarqoHybridQu
         if is_facets_term:
             return ' OR '.join(terms)
```
papa99do, line 876: Is this still OK now? or will you change this in the next PR (handle weakAnd targetHits for lexical part of the facets query). Please note this might make the facets less accurate (because of the match set discrepancy).

components/marqo/src/marqo/core/models/marqo_query.py
```diff
@@ -59,6 +59,7 @@ class MarqoLexicalQuery(MarqoQuery):
     or_phrases: List[str]
     and_phrases: List[str]
     language: Optional[str] = None
+    lexical_operand: Optional[str] = None
```
papa99do: nit. should this be Optional[LexicalOperand] for simple validation? If you change this, you'll also need to add `use_enum_values = True` to MarqoLexicalQuery.Config to make sure it is serialised correctly.

components/marqo/src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -888,6 +893,20 @@ def _generate_or_terms(self, marqo_query: Union[MarqoLexicalQuery, MarqoHybridQu
         # Default: plain weakAnd
         return f'weakAnd({", ".join(terms)})'
 
+    @staticmethod
+    def _apply_lexical_operand(lexical_operand: str, terms: list, rerank_depth_lexical: Optional[int] = None) -> str:
+        """Apply an explicit lexical operand to combine terms."""
+        if lexical_operand == 'or':
```
papa99do: nit. also, if we store the enum value, here we can use enum comparison.

components/marqo/tests/integ_tests/tensor_search/integ_tests/test_search_relevance_cutoff_feature.py
```diff
@@ -1620,14 +1620,16 @@ def setUp(self):
...
+                collapse_fields=None, lexical_operand=None):
         if hybrid_parameters is None:
             hybrid_parameters = {
                 "retrievalMethod": "disjunction",
                 "rankingMethod": "rrf",
-                "alpha": 0.5,
+                "alpha": alpha,
             }
+        if lexical_operand is not None:
+            hybrid_parameters["lexicalOperand"] = lexical_operand
```
papa99do, line 1632: I'm a bit confused, why do we set this in hybrid_parameters?

components/marqo/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
wanliAlex: Yes, let me fix it

wanliAlex: done

components/marqo/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
pandu-k, line 1: Claude feedback:

Concern 2: Temporary mutation of marqo_query (semi_structured_vespa_index.py)
```
original_operand = marqo_query.lexical_operand
marqo_query.lexical_operand = marqo_query.relevance_cutoff.lexical_operand
probe_lexical_term = self._get_lexical_search_term(marqo_query)
marqo_query.lexical_operand = original_operand
```

This mutate-then-restore pattern works but is fragile — if _get_lexical_search_term throws, the original value is never restored. A try/finally would be safer. That said, since queries are request-scoped and single-threaded, the practical risk is low.

components/marqo/src/marqo/core/models/marqo_query.py
papa99do: wait, why do we need it here in the first place? We won't need to support this for legacy lexical. search. I think it's safe to remove it from here.

components/marqo/tests/unit_tests/marqo/core/semi_structured_vespa_index/test_semi_structured_vespa_index_to_vespa_query.py
```diff
@@ -1401,5 +1401,154 @@ def test_collapse_sort_by_not_executed(self):
...
+    def test_lexical_operand_none_uses_default_logic(self):
+        """When lexicalOperand=None, default logic applies (weakAnd for no modifiers)."""
+        q = self._make_hybrid_query()
+        self.assertIsNone(q.lexical_operand)
+        result = self.vespa_index._generate_or_terms(q)
+        self.assertEqual(result, 'weakAnd(default contains "search")')
+
+    def test_lexical_operand_or_with_rerank_depth(self):
+        """When lexicalOperand='or' with rerankDepthLexical, should use OR (no targetHits wrapping)."""
+        q = self._make_hybrid_query(lexical_operand='or', rerank_depth_lexical=100,
```
papa99do, line 1458: so lexical_operand='or' overrides rerank_depth_lexical. I think this makes sense.

components/marqo/vespa/src/main/java/ai/marqo/search/HybridSearcher.java
```diff
@@ -1266,9 +1266,25 @@ void raiseErrorIfPresent(Result resultLexical, Result resultTensor) {
      * @return A new Query object configured for probe lexical search.
      */
     Query createProbeLexialQuery(Query query, Integer probeDepth, boolean verbose) {
-        Query probeLexicalQuery =
-                createSubQuery(
-                        query, MARQO_SEARCH_METHOD_LEXICAL, MARQO_SEARCH_METHOD_LEXICAL, verbose);
+        // Use dedicated probe lexical YQL when set; else fall back to main lexical.
+        String probeLexicalYql = query.properties().getString("marqo__yql.lexical.probe", null);
```
papa99do, line 1266: is this logic new? I thought I've seen it before. Is it in another PR?

components/marqo/src/marqo/core/structured_vespa_index/structured_vespa_index.py
```diff
@@ -876,7 +876,7 @@ def _generate_or_terms(self, marqo_query: Union[MarqoLexicalQuery, MarqoHybridQu
             return ' OR '.join(terms)
 
         # Explicit lexicalOperand overrides automatic logic
-        lexical_operand = getattr(marqo_query, 'lexical_operand', None)
+        lexical_operand = lexical_operand_override or getattr(marqo_query, 'lexical_operand', None)
```
papa99do: I think my previous comments got overridden by this change. I think we can use marqo_query.lexical_operand directly here (if we decided to have it in root level)

components/marqo/src/marqo/core/semi_structured_vespa_index/semi_structured_vespa_index.py
wanliAlex: fixed

components/marqo/tests/integ_tests/tensor_search/integ_tests/test_search_relevance_cutoff_feature.py
```diff
@@ -1863,24 +1863,93 @@ def test_cutoff_and_sort_by_and_facets_collapse_field(self):
         self.assertEqual(expected_facets, result["facets"])
         self.assertEqual(expected_hits, returned_hits)
```
papa99do, line 1865: Add test case for: default main(without lexical_operand), and different operand in relevance_cutoff.

components/marqo/src/marqo/core/structured_vespa_index/structured_vespa_index.py
wanliAlex: fixed

components/marqo/tests/integ_tests/tensor_search/integ_tests/test_search_relevance_cutoff_feature.py
papa99do: How do we plan to config this for stablin? Do we need to config both "and" for main query and "or" for relevance_cutoff query or just the latter?

components/marqo/src/marqo/core/models/marqo_query.py
wanliAlex: Moved to hybridparameters and use enum

components/marqo/src/marqo/core/structured_vespa_index/structured_vespa_index.py
wanliAlex: done

components/marqo/vespa/src/main/java/ai/marqo/search/HybridSearcher.java
wanliAlex: in 2.26 joshua's PR

components/marqo/tests/integ_tests/tensor_search/integ_tests/test_search_relevance_cutoff_feature.py
wanliAlex: Added

components/marqo/src/marqo/core/structured_vespa_index/structured_vespa_index.py
wanliAlex: we use marqo_query.hybrid_parameters.lexical_operand

components/marqo/src/marqo/core/structured_vespa_index/structured_vespa_index.py
wanliAlex: I think it's fine because no matter we use OR, or weakAnd, we get enough results compared with the max() constraint in facets

papa99do (approved):

PR #1407, opened 2026-04-02
Fix update documents crash with special character IDs

wanliAlex:
## Summary
- Ports the document update bug fixes from `li/add-validation-for-update-documents-ids` to mainline (the `update_index_settings` feature was already merged)
- Fixes crash when partial update documents contain IDs with `#` or non-printable ASCII characters
- Uses index-based lookup (`fetch_ids[idx]`) instead of `extract_document_id_from_vespa_id` which was unreliable due to Vespa truncating IDs containing special characters
- Adds `httpx.InvalidURL` catch in `_get_document_async` to gracefully handle IDs with non-printable ASCII
- Adds response count validation for `get_batch` responses

## Test plan
- [ ] Unit tests pass for vespa_client (3 new tests: order preservation, InvalidURL catch, mixed valid/invalid IDs)
- [ ] Integration tests for partial update with `#` in document ID
- [ ] Integration tests for partial update with non-printable ASCII in document ID
- [ ] Integration test for get_batch order preservation

### DIFF
--- a/components/marqo/src/marqo/core/document/document.py
+++ b/components/marqo/src/marqo/core/document/document.py
@@ -33,6 +33,8 @@ logger = get_logger(__name__)
 
 class Document:
     """A class that handles the document API in Marqo"""
+    GENERIC_UPDATE_DOCUMENT_ERROR_MESSAGE = "Marqo vector store couldn't update the document. " + \
+                                            "Please see: " + update_documents_response() + " for more details"
 
     def __init__(self, vespa_client: VespaClient, index_management: IndexManagement, inference: Inference):
         self.vespa_client = vespa_client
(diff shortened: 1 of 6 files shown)


components/marqo/src/marqo/core/document/document.py
```diff
@@ -331,5 +358,8 @@ def translate_add_documents_response(self, responses: Optional[FeedBatchResponse
                                          processingTimeMs=add_docs_processing_time_ms)
 
     def extract_document_id_from_vespa_id(self, resp):
```
papa99do, line 360: nit. why don't we fix the other usage of this method as well?

components/marqo/tests/integ_tests/vespa/test_update_documents_batch.py
```diff
@@ -104,6 +104,40 @@ def test_feed_batch_documents_invalid_values(self):
...
+        # Request in shuffled order, mixing existing and non-existing IDs
+        requested_ids = [
+            "order-4",          # exists
+            "nonexistent-1",    # doesn't exist
+            "order-1",          # exists
+            "order-5",          # exists
+            "nonexistent-2",    # doesn't exist
+            "order-0",          # exists
+            "nonexistent-3",    # doesn't exist
+            "order-2",          # exists
```
papa99do, line 125: should we also add some invalid id name? 
- add a doc with id with `#` in it, and try to update with the same id. 
- update docs with non ascii chars in id

components/marqo/src/marqo/core/document/document.py
wanliAlex: Other methods won't have this issue as they can accept None.

components/marqo/tests/integ_tests/vespa/test_update_documents_batch.py
wanliAlex: They are tested here: https://github.com/marqo-ai/marqo/pull/1407/changes#diff-001e2496af7ca2c99ac28142a30a1ee90d6fdf5886a4a594d0dff7d97e84a7cbR1461

Do you need me to test them again in the Vespa client?

papa99do: no, it should be fine. OK. seems both are integ tests. All good, I'll approve

papa99do (approved):

PR #1464, opened 2026-07-26
Refuse media downloads aimed at addresses that are not publicly routable

### LINKED ISSUE #1452: marqo: unauthenticated SSRF via media URLs in add_documents
- reported via email on 9 June 2026
- reported to support on 14 June 2026 https://marqo.zendesk.com/hc/en-us/requests/330
- no response, trying here

Summary

Marqo's API is unauthenticated by default and binds 0.0.0.0. During indexing it fetches media from user-supplied URLs with no destination filtering. download_image_from_url uses pycurl with FOLLOWLOCATION=1 and no IP/host filter; fetch_content_sample uses requests.get on the raw URL. The only check is validators.url (syntax). So an unauthenticated attacker can make the backend request arbitrary internal addresses.

Proof of concept (real container)

POST /indexes/ssrf-test (no auth) -> created.
POST /indexes/ssrf-test/documents {"documents":[{"_id":"d","img":"http://INTERNAL:PORT/x.png"}],"tensorFields":["img"]}
-> my listener received the request FROM the marqo container (172.17.0.2), UA=Marqobot/1.0. A no-extension URL produced extra python-requests hits (MIME sniff). img=http://127.0.0.1:8882/ returned status 200 (internal service reachable). On cloud, http://169.254.169.254/ is reachable.

This is a blind SSRF (the body is processed as an image, not returned) - a reachability oracle and pivot primitive.

Suggested fix

Resolve and reject private/loopback/link-local/reserved IPs before fetching (re-check on redirects, since FOLLOWLOCATION is on), restrict scheme to http/https, and consider disabling redirect following for media fetches.

Vansh-Sharma27:
## Change Summary

Marqo fetches media from URLs given in `add_documents` and in search queries. Neither
fetch checks where the request ends up, and the API is unauthenticated and binds
`0.0.0.0` by default. So anyone who can reach Marqo can make it send requests to internal
addresses, including `127.0.0.1` and `169.254.169.254`.

This adds a destination check to both fetch paths. The check itself is in `marqo_common`
because both components need it and they have to agree on the answer. No new dependencies.

Fixes #1452

## What was wrong

The only check on a media URL was `validators.url`, which only looks at syntax. I tested
what it accepts:

| URL | `validators.url` | Would reach |
|---|---|---|
| `http://127.0.0.1:8882/` | accepted | Marqo's own API |
| `http://169.254.169.254/` | accepted | cloud instance metadata |
| `http://10.0.0.5/img.png` | accepted | private subnet |
| `http://[::1]/img.png` | accepted | loopback over IPv6 |
| `http://[::ffff:127.0.0.1]/` | accepted | loopback, IPv4-mapped form |
| `http://example.com@127.0.0.1/` | accepted | loopback, userinfo hides the host |
| `http://images.example.com/` pointing at `10.0.0.5` | accepted | wherever DNS says |
| `file:///etc/passwd` | rejected | — |
| `http://2130706433/` | rejected | — |

After that check, two paths connect to whatever the URL resolves to:

- `download_image_from_url` (`inference_orchestrator`, pycurl): `FOLLOWLOCATION=1`, no
  destination check, no protocol restriction.
- `fetch_content_sample` (`marqo`, requests): `requests.get` on the raw URL, redirects
  followed automatically.

`infer_modality` also runs on search queries (`tensor_search.py:664`, `672`, `1051`,
`1060`), so this is reachable from `/indexes/{index}/search` too, not only from
`add_documents`.

```mermaid
flowchart TD
    AD["POST /indexes/INDEX/documents"]
    SE["POST /indexes/INDEX/search"]

    AD --> IM
    SE --> IM

    IM["infer_modality"]
    IM -->|"URL has a known extension"| SKIP["no fetch"]
    IM -->|"no extension, sniff MIME type"| FCS

    FCS["fetch_content_sample (requests)"]
    AD --> IOC["download_image_from_url (pycurl)"]

    FCS --> GA["check host addresses,<br/>re-check each redirect hop"]
    IOC --> GB["check address libcurl resolved,<br/>before connect, each hop"]

    GA --> NET["publicly routable address"]
    GB --> NET
```

## What the check does

```mermaid
flowchart TD
    A["address"] --> B{"IPv4-mapped IPv6?"}
    B -->|yes| C["convert to IPv4"]
    B -->|no| D
    C --> D{"in MARQO_MEDIA_DOWNLOAD_ALLOWED_NETWORKS?"}
    D -->|yes| ALLOW["allow"]
    D -->|no| E{"is_global?"}
    E -->|no| DENY["refuse"]
    E -->|yes| F{"private, loopback, link-local,<br/>reserved, multicast or unspecified?"}
    F -->|yes| DENY
    F -->|no| ALLOW
```

I check all of those properties rather than just one, because neither `is_private` nor
`is_global` is enough on its own. On Python 3.11.15:

| Address | `is_private` | `is_global` | Result |
|---|---|---|---|
| `100.64.0.1` (CGNAT) | `False` | `False` | refused. `is_private` alone would allow it |
| `224.0.0.1` (multicast) | `False` | `True` | refused. `is_global` alone would allow it |
| `64:ff9b::7f00:1` (NAT64 of `127.0.0.1`) | `False` | `True` | refused. `is_global` alone would allow it |
| `::ffff:127.0.0.1` | `True` | `False` | refused, after converting to `127.0.0.1` |
| `8.8.8.8` | `False` | `True` | allowed |

## What changed on each path

**Image download (pycurl).** The address is checked from `CURLOPT_OPENSOCKETFUNCTION`.
libcurl calls this once per connection, including for each redirect it follows, and gives
the address it resolved. Returning `SOCKET_BAD` refuses the destination before the TCP
connect happens. `PROTOCOLS` and `REDIR_PROTOCOLS` are also set to HTTP and HTTPS only.

**MIME sniffing (requests).** requests has no equivalent hook, so this path resolves the
host itself and follows redirects one hop at a time, checking each hop before requesting
it. Since this takes redirect handling away from requests, it also drops credential
headers when a hop changes host, which is what requests already did.

## Why these mechanisms and not others

| Considered | Problem with it |
|---|---|
| Check the URL once before fetching | A redirect is never checked |
| `PREREQFUNCTION` | Runs after the TCP connect, so open vs closed ports are still visible by timing. It also returns `E_ABORTED_BY_CALLBACK`, which this function already maps to "exceeds the maximum allowed size" |
| Turn off `FOLLOWLOCATION` and follow hops manually | Means redoing libcurl's timeout budgeting, hop limits and relative URL handling in the indexing path |
| Read the peer address back from a streamed `requests` response | `response.raw.connection.sock` is `None` once headers are read |

For `OPENSOCKETFUNCTION` the callback has to return a real socket, so I checked what owns
the descriptor before using it: over 200 downloads the fd count does not move, and keeping
a reference to the socket object instead leaks one fd per download. libcurl takes
ownership, so the callback does not keep the socket.

## What each part is for

If you want a narrower change, this is what dropping any one piece brings back:

| Part | Dropping it allows again | Shown by |
|---|---|---|
| Checking all the address properties | CGNAT, multicast, NAT64-wrapped loopback | table above |
| Converting IPv4-mapped addresses | `http://[::ffff:127.0.0.1]/` | `test_download_from_ipv4_mapped_loopback_address_is_refused` |
| `OPENSOCKETFUNCTION` over a one-time URL check | redirect to an internal address | `test_download_refuses_redirect_to_internal_address` |
| `OPENSOCKETFUNCTION` over `PREREQFUNCTION` | port scanning by timing | 0.00s refused vs 3.00s connect timeout, `test_download_refuses_redirect_before_connecting_to_it` |
| `PROTOCOLS` / `REDIR_PROTOCOLS` | `file://` reads the container filesystem | unrestricted handle returned `/etc/hostname` |
| Per-hop check on the requests path | redirect to an internal address from search | `test_fetch_refuses_redirect_to_internal_address` |
| Requiring every resolved address to pass | a host that answers with one public and one internal address | `test_any_internal_address_in_the_answer_refuses_the_host` |
| Redirect hop cap | redirect loops | `test_fetch_refuses_a_redirect_loop` |
| Leaving the resolved address out of the error | reading internal DNS through error messages | `test_refusal_does_not_disclose_the_resolved_address` |

Two things in the diff are not new checks. Dropping credential headers across hosts keeps
what requests already did. `MARQO_MEDIA_DOWNLOAD_ALLOWED_NETWORKS` is there so private
media hosts still work.

## Configuration

```
MARQO_MEDIA_DOWNLOAD_ALLOWED_NETWORKS=10.0.0.0/8,192.168.0.0/16
```

Empty by default. I used a CIDR list rather than an on/off flag so that allowing
`10.0.0.0/8` does not also re-allow `169.254.169.254`. An invalid entry fails at startup
instead of being ignored.

## Testing

Ubuntu 22.04, Python 3.11.15, pycurl 7.45.7 / libcurl 8.16.0.

| Check | Result |
|---|---|
| `inference_orchestrator` unit tests | 402 passed, 194 subtests |
| `marqo` unit tests | 1255 passed, 3 skipped (already skipped on `mainline`), 1120 subtests |
| `ruff check` and `ruff format --check` on changed files | clean |

The tests use a loopback HTTP server rather than a mocked pycurl handle, because the check
has to hold for the address libcurl actually connects to and a mock cannot show that. They
fail on `mainline` because the download succeeds.

Covered: the PoC from the issue, the metadata endpoint, private subnets, IPv6 and
IPv4-mapped loopback, userinfo in the host, `file://`, redirect to an internal address,
redirect loops, the allow list, and a 21 case address table.

I also checked real hosts still work: `raw.githubusercontent.com`, `upload.wikimedia.org`,
`images.unsplash.com`, `storage.googleapis.com`, `s3.amazonaws.com`, `example.com` and
`www.google.com` all pass, including their IPv6 addresses, and downloading a real image
returns a valid JPEG.

## Behaviour changes and limits

- Media URLs that resolve to private, loopback, link-local, CGNAT, multicast or reserved
  addresses are now refused by default. This is the point of the change, but it is the
  part existing users can notice. The error message names the environment variable.
  If you would rather keep the upgrade quieter, blocking only loopback, link-local and
  reserved while allowing RFC1918 would still cover the PoC and the metadata endpoint. I
  did not do that because in Docker the sibling services are on RFC1918, so it would leave
  container to container requests open. Happy to change it.
- Redirects are capped at 10 hops. `FOLLOWLOCATION` had no limit before.
- The pycurl path checks the address libcurl resolved, so a host that resolves differently
  on a second lookup is still caught. The requests path resolves the host and then hands
  the URL to requests, which resolves again, so DNS rebinding is still possible there. To
  close it the resolved address would have to be pinned into the connection, and urllib3
  does not expose that. That path reads 10 KB and throws it away after checking the MIME
  type, so it is much weaker than the download path.
- A host that cannot be resolved now fails in the check rather than in the HTTP client.
  The exception type is the same.
- `RELEASE.md` is not touched, since entries there have a PR number and are written at
  release time. Happy to add one.

## Related Jira Ticket

N/A, external contribution. Tracked by #1452.

## Checklist

- [ ] Documentation has been updated. The new environment variable is in `configs.py` and
      in the error message. Happy to add it to the docs site if that lives elsewhere
- [x] Breaking changes are clearly identified, see the first bullet above
- [x] Python client changes linked or N/A. N/A


Vansh-Sharma27:
Hi @papa99do @wanliAlex , friendly bump! Would appreciate a review when you get a chance. Happy to address any feedback!

